PR #6, opened 2023-11-03
gha: adding qa workflow

andrsd:


ragusa (commented):
quick question: can we change a file that is BSD-3? Here's a note from the old repo. "The paraview contains a partial version of the ParaView source files. The directory structure does not necessarily conform to that of the original ParaView project "

andrsd:
> quick question: can we change a file that is BSD-3?

Are you asking if it is ok to reformat the file with clang-format? I would think it is ok, but we can exclude the external directory from being checked by clang-format... May be a better approach than reformatting.

andrsd:
I undid the formatting in the `external` dir and added it into ignored directories.

ragusa (approved):

PR #8, opened 2023-11-03
GHA: adding build workflow

andrsd:


ragusa:
@andrsd  I am inclined to merge this. With Petsc 3.17.5, the only tests that are failing are from the post-processor system (solver_info_01.lua and solver_info_02.lua). I am not sure why they are failing. Is the build system still the same on opensn. Can I compile the code in the old way and see, or can you look and compare the golden values and what is in the out/ folder for these runs?

ragusa (approved):

andrsd:
This takes about 2.5 hours to finish, plus the other failures that are showing on github VM, but not on `class01`. it can be mpi version, compiler, and what not, so I will close this. I don;t want to wait 2.5 hours to find out what I did wrong, when `class01` can tell me that under 10 minutes.

Issue #3, opened 2023-11-02
Code style / Naming conventions


andrsd:
This issue is here to decide on the coding style.  Currently the code is a mix of different styles, but the code should really just use one style.

```
directory_name/header_file_name.h
directory_name/source_file_name.cc
```

```
#define MacroDefinition(parameter)

namespace out_namespace {

enum OurEnumType {
  ENUM_VALUE_1,
  ENUM_VALUE_2,
  ...
}

class ThisIsAClassName {
public:
   int public_member_var;
protected:
   int my_member_variable_;
   double another_member_variable_;

   void MyCoolMemberFunction();
   void MemberFunctionWithAnArgument(int argument_name);
};

struct ThisIsAStructName {
   int public_member_var;

   void MyCoolMemberFunction();
   void MemberFunctionWithAnArgument(int argument_name);

protected:
   int my_member_variable_;
   double another_member_variable_;
};

}
```

Command line parameters:
```
--cmd-line-switch1 --cmd-line-switch2
```

andrsd:
Free free to comment on stuff that should be kept or changed.

ragusa:
I am ok with snake case and pascal case as you have shown.

wdhawkins:
I'm happy with this style. No changes from me. I would like to talk a bit about comments.  I'm not a fan of the current comment style (which typically includes a long line of "=").  Thoughts?

andrsd:
I saw 2 kinds of this style:
1. something like this `// ============` that was in between functions. Those need to go - they are useless.
2. something like `// ========= comment`. This needs to be changed into `// comment`.

wdhawkins:
> I saw 2 kinds of this style:
> 
>     1. something like this `// ============` that was in between functions. Those need to go - they are useless.
> 
>     2. something like `// ========= comment`. This needs to be changed into `// comment`.

Yes!  Perfect. I'm fine with `// ` or `/*  */` syntax, but prefer `// `.

andrsd:
I also prefer `//`. I think the `/* */` works well only for doxygen-style comments, but not in code...

zhardy-lanl:
There is also a mixture of different getter function names that should probably be standardized. Take for example
```
class MyClass
{
public:
  // Option 1
  Type Member() { return member_; }

  // Option 2
  // Type GetMember() { return member_; }
  
private:
  Type member_;
};
```

andrsd:
[...]

PR #12, opened 2023-11-07
Architectural changes

### LINKED ISSUE #1: Redo build scripts
Tasks:

- [ ] remove references to chitech
- [ ] Use find LUA via `find_package`. this is available since cmake 3.18
- [ ] PETSc should be detected using `find_package`. this will need `FindPETSc.cmake`, but we can check PETSc version, etc.
- [ ] all supporting cmake modules are typically stored in `<ROOT>/cmake`, so move stuff there
- [ ] should use `target_link_libraries` and `target_include_directories` instead of `include_directories` and `link_directories` (these are global and really the include and linking stuff is always target specific).
- [ ] binaries should not be produced in the source tree

andrsd:
Changes:
- move `Lua` wrapper into `lua/`
- compile C++ classes as `libopensn`
- compile Lua layer as `libopensnlua`
- redoing cmake scripts
   - using `find_package` to find dependencies with specified versions
   - reducing the number `CMakeLists.txt` there is only 3 now
- adding configure option `OPENSN_WITH_LUA=ON/OFF` (default is ON)
- binary for testing is called `opensn-test`
- binary to run Lua scripts is called `opensn`
- redoing #includes

Closes #1


andrsd:
Additional comments:
- `LUA_ROOT` is no longer a thing since we are using `find_package`. If libraries are installed in a non-standard location, one uses `CMAKE_PREFIX_PATH=path1:path2:..` to defined the trees where to look for header files and libraries.
- `PETSC_ROOT` still works, but `PETSC_DIR` as well
- as for VTK, no change
- no binaries are produced in a source tree
- config file is now plain `config.h` produced in the binary tree
- no extra `Makefile` is produced in the source tree, there is just the standard one produced by cmake in the binary tree
- `run_tests` should work when executed from the source tree as it did before
  - when executed from the build tree, specifying `--exe test/opensn-test` should work

ragusa:
@andrsd @wdhawkins  the tests are currently failing (quickly). The configure step is actually what is failing. Does this have to do with the class-01 machine and Daryl needs to fix something first ??

configure.sh
```diff
@@ -9,17 +9,17 @@
 #   script depends on the arguments passed:
 #   
 #   1. If called without arguments, CMake scripts for a release (optimized) build
-#      of ChiTech will be generated in the directory chi_build. Additional
+#      of ChiTech will be generated in the directory `build`. Additional
```
ragusa, line 12: ChiTech mention here

andrsd:
@wdhawkins We need to patch the gha workflow for the regression tests. It will need this:
```
- name: Setup environment
  run: |
    echo "CMAKE_PREFIX_PATH=:<PATH WHERE LUA IS>" >> $GITHUB_ENV
```
That should help `cmake` to find Lua

configure.sh
```diff
@@ -9,17 +9,17 @@
...
 #      arguments will be passed to and interpreted by the CMake generator.
 #
 #   2. If called with the single argument from the following list:
 #         Debug, Release, RelWithDebInfo, MinSizeRel,
 #      scripts for the corresponding build of ChiTech will be generated in
```
ragusa, line 17: another mention on this line

andrsd:
Or a similar thing must be done in the module file that load the environment...

framework/chi_runtime.cc
```diff
@@ -157,11 +163,13 @@ Chi::run_time::ParseArguments(int argc, char** argv)
...
+#ifdef OPENSN_WITH_LUA
   if (Chi::run_time::dump_registry_)
   {
     ChiObjectFactory::GetInstance().DumpRegister();
     Chi::console.DumpRegister();
   }
+#endif
 }
 
 // ############################################### Initialize ChiTech
```
ragusa, line 175: another mention

andrsd:
Comment on the `#include` work. After opensn is installed (which will be a thing in future) via `make install`, it will be sitting in `<INSTALL_PREFIX>/include/opensn` and `<INSTALL_PREFIX>/lib`. Typically as a client app, you only want to specify one include dir which would be `<INSTALL_PREFIX>/include/opensn` with this all the includes will just work, since that 's the common prefix I am assuming in the code.

configure.sh
andrsd: I did not change references to ChiTech in the code, yet. Just in the build script. For code, a lot more work will be required with file renames, etc.

andrsd: I guess, you see `configure.sh` as part of the build infrastructure, I can add a patch to fix it in this file. For the code changes, I would personally go with a separate PR. IMO PRs should be as small as possible and do just 1 thing.

andrsd: One more thing, what is final agreement of the capitalization of the code name? IIRC, there were 3 variants:
- OpenSn
- openSn
- opensn

If there is no agreement, yet. No point in changing that in this PR :-)

ragusa: Yes, I was thinking the .sh can be done now as I've seen you also added opensn so the configure.sh should remove the old name.

framework/chi_runtime.cc
ragusa: this is a .cc file. it can be done later

configure.sh
ragusa: A non unanimous majority likes: openSn for the code's name so we're going with that.

framework/chi_runtime.cc
andrsd: We talked about this [here](https://github.com/Open-Sn/openSn/issues/3#issuecomment-1795912565). So I assume a PR where we just fix the comment style first. And another one later with actually changing the names...

configure.sh
andrsd: Ok. I will add a patch to change the code name in `configure.sh`.

andrsd:
I am missing an INCLUDE in the `libopensnlua` target. Will patch...

ragusa (approved):

wdhawkins (approved):

PR #15, opened 2023-11-08
Add option to run_tests to allow user to specify job launch command

wdhawkins:
I think the test scripts could use a bit of refactoring, but this at least gives the user the ability to specify the launch command in situations where mpirun/mpiexec is not used.

### DIFF
--- a/test/run_tests
+++ b/test/run_tests
@@ -70,7 +70,7 @@ parser.add_argument(
 )
 
 parser.add_argument(
-   "--mpicmd", default="mpiexec -n ", type=str, required=False,
+   "--mpi-cmd", default="mpiexec -n ", type=str, required=False,
    help="MPI application launcher command"
 )
 
--- a/test/run_tests
+++ b/test/run_tests
@@ -69,6 +69,11 @@ parser.add_argument(
     help="Controls verbose failure"
 )
 
+parser.add_argument(
+   "--mpicmd", default="mpiexec -n ", type=str, required=False,
+   help="MPI application launcher command"
+)
+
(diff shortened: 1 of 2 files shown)


test/run_tests
```diff
@@ -69,6 +69,11 @@ parser.add_argument(
     help="Controls verbose failure"
 )
 
+parser.add_argument(
+   "--mpicmd", default="mpiexec -n ", type=str, required=False,
```
andrsd, line 73: Do you want to do `--mpi-cmd` instead? Usually multi-word switches have some sort of separation and most commonly used (from what I have seen) was `-`.

wdhawkins: I think that's going to fail with argparse (at least with the current implementation).  Let me check to make sure.

ragusa (approved):
I think --mpi-cmd is ok. I am going to approve this but feel free to go changing to --mpi-cmd without the need for re-approval (hopefully)

test/run_tests
wdhawkins: Aha... the trick is that "-" gets replaced by "_" once argparse parses the input.  Done!

ragusa: will argpase always behave like this int he future? why not change right away to `--mpi_cmd` then ?

andrsd: They also have `dest=` param, so in case something breaks, we just do `dest='mpi_cmd'`.

wdhawkins: > will argpase always behave like this int he future? why not change right away to `--mpi_cmd` then ?

It will... that just internally though.  For user facing input, I think I do prefer `-- mpi-cmd`.

PR #16, opened 2023-11-08
Cleanup in comments/usage, etc.

andrsd:
- Removing `// #+` comments
- Patching `// =+ text` comments
- Updating `//` comments into doxygen
- Removing commented-out code
- Moving doxygen-style comments from .cc to .h
- Removing comments inside function calls
- Default values are in .h file, no need to have that in .cc.


framework/ChiObjectFactory.cc
```diff
@@ -3,35 +3,25 @@
 #include "framework/chi_runtime.h"
 #include "framework/logging/chi_log.h"
 
-// ###################################################################
-/**Access to the singleton*/
 ChiObjectFactory&
```
ragusa, line 6: I do not understnad all of the deletion. I agree that // ##### has to go. I agree with do not like /** */ and prefer // but why is this not changed to // Access to the singleton. That is, why not keep the cmment. It looks like everything, including the comment was deleted.

andrsd: Because doxygen-comments should be *only* with the declaration and not definition, i.e. either the comment was moved (if it was not one at the declaration), or deleted if there was an identical copy there already.

wdhawkins: I'm also a big fan of not adding doxygen comments for every method.  If it's obvious what the method does, I usually don't document it.

ragusa: my bad. all good.

ragusa (approved):

wdhawkins (approved):

PR #31, opened 2023-11-09
Refactoring `LinearSolver` class

andrsd:
- Undo templating on LinearSolver class
- SourceFlags act like a bit-wise array


modules/linear_boltzmann_solvers/a_lbs_solver/lbs_structs.h
```diff
@@ -95,10 +94,102 @@ enum SourceFlags : int
   ZERO_INCOMING_DELAYED_PSI = (1 << 6)
 };
 
-inline SourceFlags
-operator|(const SourceFlags f1, const SourceFlags f2)
+/**
+ * Source representation (a combination of `SourceFlag`s)
+ */
+struct Source
```
andrsd, line 100: Feel free to suggest a better name here. I used `Source` as in combination of `SourceFlag`s...

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -17,19 +17,19 @@ void
 SourceFunction::operator()(LBSGroupset& groupset,
                            std::vector<double>& destination_q,
                            const std::vector<double>& phi_local,
-                           SourceFlags source_flags)
+                           Source source)
 {
-  if (source_flags & NO_FLAGS_SET) return;
+  if (source.Empty()) return;
```
andrsd, line 22: This was an interesting line. `NO_FLAGS_SET` was `0` so `anything & 0 == false`, so this shortcut did not work. Maybe, earlier in the code `NO_FLAGS_SET` had a non-zero value... This way it should be better...

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_structs.h
```diff
@@ -83,9 +83,8 @@ struct BoundaryPreference
   std::string source_function;
 };
 
-enum SourceFlags : int
+enum SourceFlag
```
andrsd, line 86: This could also be made into `enum class SourceFlag` to improve type safety, but then all the constants in the code would be referred to as `SourceFlag::ENUM_VALUE`. This may not be needed, since I made `Source` such that it can only be initialized using `SourceFlag`, i.e. code like this `Source src(1233);` is not gonna compile, so there should be enough safety to prevent people to initialize `Source` with random enums/values/etc.

ragusa: Source --> SourceFlags
enum SourceFlag --> SourceType

andrsd: Done

ragusa (approved):

framework/math/linear_solver/linear_solver.h
```diff
@@ -78,29 +48,54 @@ class LinearSolver
...
+  const std::string iterative_method_;
+
+  LinSolveContextPtr context_ptr_ = nullptr;
+
+  /// Matrix
+  Mat A_;
+  /// Right-hand side
+  Vec b_;
+  /// Solution vector
+  Vec x_;
```
zhardy-lanl, line 88: We could rename these to `matrix_`, `rhs_` and `sol_` (or `solution_`) to both meet syntax (`snake_case` for variables) and eliminate redundant comments.

framework/math/linear_solver/linear_solver.h
```diff
@@ -78,29 +48,54 @@ class LinearSolver
...
+  /// PETSc KSP object
+  KSP ksp_;
+  /// Number of local DOFs in the system
+  int64_t num_local_dofs_ = 0;
+  /// Number of global DOFs in the system
+  int64_t num_globl_dofs_ = 0;
```
zhardy-lanl, line 94:
```suggestion
  int64_t num_global_dofs_ = 0;
```

framework/math/linear_solver/linear_solver.h
```diff
@@ -78,29 +48,54 @@ class LinearSolver
...
+  Mat A_;
+  /// Right-hand side
+  Vec b_;
+  /// Solution vector
+  Vec x_;
+  /// PETSc KSP object
+  KSP ksp_;
+  /// Number of local DOFs in the system
+  int64_t num_local_dofs_ = 0;
+  /// Number of global DOFs in the system
```
zhardy-lanl, line 93:
```suggestion
```
Redundant

framework/math/linear_solver/linear_solver.h
```diff
@@ -78,29 +48,54 @@ class LinearSolver
...
+
+  /// Matrix
+  Mat A_;
+  /// Right-hand side
+  Vec b_;
+  /// Solution vector
+  Vec x_;
+  /// PETSc KSP object
+  KSP ksp_;
+  /// Number of local DOFs in the system
```
zhardy-lanl, line 91:
```suggestion
```
Redundant

framework/math/linear_solver/linear_solver.h
```diff
@@ -78,29 +48,54 @@ class LinearSolver
...
+  LinSolveContextPtr context_ptr_ = nullptr;
+
+  /// Matrix
+  Mat A_;
+  /// Right-hand side
+  Vec b_;
+  /// Solution vector
+  Vec x_;
+  /// PETSc KSP object
+  KSP ksp_;
```
zhardy-lanl, line 90:
```suggestion
  KSP solver_;
```
```suggestion
  /// PETSc linear solver
 KSP ksp_;
```
Redundant. Consider changing the doc or the member name.

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/adjoint_src_function.cc
```diff
@@ -15,7 +15,7 @@ void
 AdjointSourceFunction::AddVolumetricQOISources(LBSGroupset& groupset,
                                                std::vector<double>& destination_q,
                                                const std::vector<double>&,
-                                               SourceFlags source_flags)
+                                               SourceFlags source)
```
zhardy-lanl, line 18:
```suggestion
                                               SourceFlags source_flags)
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/adjoint_src_function.h
```diff
@@ -18,18 +18,18 @@ class AdjointSourceFunction : public SourceFunction
   void AddAdditionalSources(LBSGroupset& groupset,
                             std::vector<double>& destination_q,
                             const std::vector<double>& phi,
-                            SourceFlags source_flags) override
+                            SourceFlags source) override
```
zhardy-lanl, line 21:
```suggestion
                            SourceFlags source_flags) override
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/adjoint_src_function.h
```diff
@@ -18,18 +18,18 @@ class AdjointSourceFunction : public SourceFunction
...
   {
     // Inhibit -> AddPointSources
     // Add     -> AddVolumetricQOISources
-    AddVolumetricQOISources(groupset, destination_q, phi, source_flags);
+    AddVolumetricQOISources(groupset, destination_q, phi, source);
```
zhardy-lanl, line 25:
```suggestion
    AddVolumetricQOISources(groupset, destination_q, phi, source_flags);
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/adjoint_src_function.h
```diff
@@ -18,18 +18,18 @@ class AdjointSourceFunction : public SourceFunction
...
   }
 
   /**Adds Quantities of Interest to the nodal sources.*/
   void AddVolumetricQOISources(LBSGroupset& groupset,
                                std::vector<double>& destination_q,
                                const std::vector<double>& phi,
-                               SourceFlags source_flags);
+                               SourceFlags source);
```
zhardy-lanl, line 32:
```suggestion
                               SourceFlags source_flags);
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -17,19 +17,19 @@ void
 SourceFunction::operator()(LBSGroupset& groupset,
                            std::vector<double>& destination_q,
                            const std::vector<double>& phi_local,
-                           SourceFlags source_flags)
+                           SourceFlags source)
```
zhardy-lanl, line 20:
```suggestion
                           SourceFlags source_flags)
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -17,19 +17,19 @@ void
...
 {
-  if (source_flags & NO_FLAGS_SET) return;
+  if (source.Empty()) return;
```
zhardy-lanl, line 22:
```suggestion
  if (source_flags.Empty()) return;
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -17,19 +17,19 @@ void
...
-  apply_ags_scatter_src_ = (source_flags & APPLY_AGS_SCATTER_SOURCES);
-  apply_wgs_fission_src_ = (source_flags & APPLY_WGS_FISSION_SOURCES);
-  apply_ags_fission_src_ = (source_flags & APPLY_AGS_FISSION_SOURCES);
-  suppress_wg_scatter_src_ = (source_flags & SUPPRESS_WG_SCATTER);
+  apply_fixed_src_ = (source & APPLY_FIXED_SOURCES);
+  apply_wgs_scatter_src_ = (source & APPLY_WGS_SCATTER_SOURCES);
+  apply_ags_scatter_src_ = (source & APPLY_AGS_SCATTER_SOURCES);
+  apply_wgs_fission_src_ = (source & APPLY_WGS_FISSION_SOURCES);
+  apply_ags_fission_src_ = (source & APPLY_AGS_FISSION_SOURCES);
+  suppress_wg_scatter_src_ = (source & SUPPRESS_WG_SCATTER);
```
zhardy-lanl, line 32:
```suggestion
  apply_fixed_src_ = (source_flags & APPLY_FIXED_SOURCES);
  apply_wgs_scatter_src_ = (source_flags & APPLY_WGS_SCATTER_SOURCES);
  apply_ags_scatter_src_ = (source_flags & APPLY_AGS_SCATTER_SOURCES);
  apply_wgs_fission_src_ = (source_flags & APPLY_WGS_FISSION_SOURCES);
  apply_ags_fission_src_ = (source_flags & APPLY_AGS_FISSION_SOURCES);
  suppress_wg_scatter_src_ = (source_flags & SUPPRESS_WG_SCATTER);
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -144,7 +144,7 @@ SourceFunction::operator()(LBSGroupset& groupset,
     }     // for dof i
   }       // for cell
 
-  AddAdditionalSources(groupset, destination_q, phi_local, source_flags);
+  AddAdditionalSources(groupset, destination_q, phi_local, source);
```
zhardy-lanl, line 147:
```suggestion
  AddAdditionalSources(groupset, destination_q, phi_local, source_flags);
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -181,9 +181,9 @@ void
 SourceFunction::AddPointSources(LBSGroupset& groupset,
                                 std::vector<double>& destination_q,
                                 const std::vector<double>&,
-                                SourceFlags source_flags)
+                                SourceFlags source)
```
zhardy-lanl, line 184:
```suggestion
                                SourceFlags source_flags)
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -181,9 +181,9 @@ void
...
 {
-  const bool apply_fixed_src = (source_flags & APPLY_FIXED_SOURCES);
+  const bool apply_fixed_src = (source & APPLY_FIXED_SOURCES);
```
zhardy-lanl, line 186:
```suggestion
  const bool apply_fixed_src = (source_flags & APPLY_FIXED_SOURCES);
```

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/adjoint_src_function.cc
```diff
@@ -33,7 +33,7 @@ AdjointSourceFunction::AddVolumetricQOISources(LBSGroupset& groupset,
   const auto& grid = adjoint_solver.Grid();
   const auto num_groups = adjoint_solver.NumGroups();
 
-  const bool apply_fixed_src = (source_flags & APPLY_FIXED_SOURCES);
+  const bool apply_fixed_src = (source & APPLY_FIXED_SOURCES);
```
zhardy-lanl, line 36:
```suggestion
  const bool apply_fixed_src = (source_flags & APPLY_FIXED_SOURCES);
```

zhardy-lanl (changes requested):
Two small items:

1. With the change in `Source` to `SourceFlags`, I think it would be best to revert argument name back to `source_flags`.

2. There are redundant member comments in `LinearSolver`. We would be better served renaming the members in lieu of one word descriptions.

If there is consensus disagreement with these suggestions, feel free to merge anyways.

framework/math/linear_solver/linear_solver.h
ragusa: I prefer to keep the math-sounding of "Ax=b". It is in linear solver, it's going to be about some Ax=b so I prefer A_, b_ and x_ (and it is shorter). Ok ?

framework/math/linear_solver/linear_solver.h
ragusa: I agree with global being fully spelled out. It's not our fault that local has one less letter than global!

I've used n_xyz for number of xyz. I think n_ is sufficient compared to num_. It's just my opinion. No action needed.

framework/math/linear_solver/linear_solver.h
zhardy-lanl: I tend to agree despite the fact it violates the member naming convention. I would get rid of the comments still since it is standard notation.

framework/math/linear_solver/linear_solver.h
ragusa: I'm ok with removing comments but I see /// so isn't that comment used in the doxy? Generally, we need to discuss documentation soon, before the end of the year. Meanwhile, I'd say do not change it (unless I've misunderstood something!)

framework/math/linear_solver/linear_solver.h
ragusa: same as above

framework/math/linear_solver/linear_solver.h
zhardy-lanl: Yes, those comments go to doxy. The reason I say to remove them is solely based on the principal that obvious things do not need documentation. In this case, the doxy will look something like
```
num_local_dofs_
Number of local DOFs in the system.
```
which does not add anything of value.

framework/math/linear_solver/linear_solver.h
ragusa: I prefer to keep `ksp_`. In PETSc , ksp is how it's called all over the place (example: `KSPSetType(KSP ksp,KSPType method);` ). If someone goes deep enough to see `ksp_`, then they either know what it is or will figure out really quick. I know it goes a little bit against the naming convention, but, to me, a case could be name to ignore the rule a few times, especially if it is a "traditional" name.

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
ragusa: I think it is clearer to actually keep `source_flags` here. (same remark for all other comments below). I am ready to commit the suggestion (and the ones below) but I'll wait until tomorrow to see if someone disagrees with Zach's proposed changes.

zhardy-lanl (approved):

framework/math/linear_solver/linear_solver.h
zhardy-lanl: This is true. The doc saying that `KSP ksp_;` is a KSP object was what I was trying to mitigate. Keeping `ksp_` is fine. It is standard PETSc after all

ragusa (approved):

PR #33, opened 2023-11-09
Removing EventCode magic constants

andrsd:
- EventCode is an enum
- Translation from event name to event code is done inside Event


framework/event_system/event.h
```diff
@@ -10,18 +10,49 @@ namespace chi
 class Event
 {
 public:
-  Event(const std::string& name, int code);
-  Event(const std::string& name, int code, const ParameterBlock& parameter_block);
+  enum ECode
```
zhardy-lanl, line 13: Particularly since this is public, I would consider naming this `EventCode` or just `Code`.

andrsd: I started with `Code`, but then there is a `Code()` method, so I think `EventCode` will be better...

wdhawkins (approved):

ragusa (approved):

PR #35, opened 2023-11-09
Removing deprecated meshing code

wdhawkins:
Removing deprecated meshing code. Everything removed in this pr is truly dead/unused code. It has no effect on any of the existing regression tests.


ragusa (approved):

framework/mesh/mesh_modifiers/mesh_modifier.h
andrsd, line 1: Maybe, I misunderstood during our Wednesday meeting, but I thought you wanted `MeshModifier`s to stay, because they were supposed to be the way you assign the material IDs when you extrude meshes?

wdhawkins: The only mesh modifier I could find was related to mesh cutting. Let me double-check this though. Now that you mention it, I should probably run the C5G7 tests with these changes.  I'm pretty sure it's an extruded mesh.

ragusa: @andrsd is right. I overlooked that. modifiers, even though empty today, is where some functions need to be moved to. So all files related to mesh modifiers should stay. Good catch!!

wdhawkins: I'll revert the mesh modifier base.

lua/framework/mesh/mesh_modifiers/lua_mesh_modifier_apply.cc
ragusa, line 1: I think we will need the "apply" method. so we should keep this in.

lua/framework/mesh/mesh_modifiers/lua_mesh_modifiers.h
ragusa, line 1: I think we will need the "input parameter" block as this is where the user instructions will come form. so we should keep this in.

wdhawkins: I restored all of the Lua code. I'm just cutting out big swaths right now... I'll look to slim things down in smaller, more focused commits.

ragusa:
@wdhawkins I think this is ready for merging, right?

Issue #44, opened 2023-11-16
Refactor namespaces


andrsd:
### Tasks
- [ ] Introduce one global namespace called `opensn`
- [ ] remove all nested namespaces

### Lua layer

There are also namespaces in the lua layer. I propose to add a global namespace called `opensnlua` so it matches the library name. 

However, there are things like this: `<module name>::common_lua_utils::RegisterLuaEntities(L);`. This could be changed into `opensnlua::<module name>::RegisterLuaEntities(L);` There might be other function name collisions as well, this one was obvious to see.

### Notes

After code inspection there are following class name collisions in the `framework` and `modules` directory:
1. `Solver` in modules: `framework`, `cfem_diffusion`, `dfem_diffusion`, `diffusion`, `fv_diffusion`, `mg_diffusion`
2. `TransientSolver` in modules: `lbs`, `prk`
3. `SweepChunk` in modules: `framework`, 'lbs`
4. `BoundaryType` in modules: `fv_diffusion`, `framework`, `cfem_diffusion`, `dfem_diffusion`, `lbs`, `mg_diffusion`
5. `Quadrature` in modules: `framework`, `sldfe`

ragusa:
@andrsd can this be closed?

andrsd:
Yes, we can close this. #435 finished this task.

Issue #9, opened 2023-11-04
Separate LUA layer


andrsd:
There is a LUA layer in the code that is used to drive the simulation. Currently this layer is compiled into the library (i.e. the computation heart of the code).

This layer needs to be refactored away from the computation and become its own library that is linked to the driver layer, i.e. binary that drives the computation.

This will improve internal design and will be very useful for scenarios where OpenSn will be used for in-memory coupling.  This will also enable to compile the library without LUA, i.e. when users just need the raw power of the library. Finally, this will also help in future if somebody will want to use another scripting language to drive their simulation (python, etc.)

ragusa:
I agree. Note that with lua, we were able to connect a TCP port and listen in. This was useful for GUI's and later for digital twins where an action from a user or an update sensor variable can be read in during the simulation. (interaction mode you see at the very beginning of the main). This is a useful feature we should keep in mind and retain. This is only an FYI as I think we should be able to achieve this with lua, python, java, ...

andrsd:
This work should not remove any capabilities you already have. Just move them in a different spot. Just to clarify...

ragusa:
done

PR #48, opened 2023-11-28
Refactoring namespaces

andrsd:
- Introducing global namespace opensn
- Renaming namespaces in tests

Refs #44


wdhawkins (approved):
My favorite PR so far.

andrsd:
I renamed the 3 enums in `ffinterpolation.h` so they have the `FieldFunctionInterpolation` prefix. And this PR is now ready for review.

Also tagging @zhardy-lanl in case he's interested in looking at this big change.

framework/console/console.cc
```diff
@@ -96,7 +96,7 @@ Console::LuaWrapperCall(lua_State* L)
       main_arguments_block.AddParameter(arg_name, lua_tostring(L, p));
     else if (lua_istable(L, p))
     {
-      auto block = chi_lua::TableParserAsParameterBlock::ParseTable(L, p);
+      auto block = TableParserAsParameterBlock::ParseTable(L, p);
```
ragusa, line 99: @wdhawkins  @zhardy-lanl @andrsd 
Previously, someone reading this would know that TableParserAsParameterBlock was related to lua. By flattening the namespace, that connection is lost. I am fairly fine with this but I have two comments:
1. if we want to increase the clarity/hints, should such a function to named LuaTableParserAsParameterBlock  ?
2. a more general question: how will another input system (Moose system, or python API) would interfere or co-exist here ? (likely a more general discussion for another meeting ...)

andrsd: The console object will be moved into the lua layer in one of the upcoming PRs I have prepared. It will end up sitting in the `opensnlua` namespace, so we won't need to rename this object or anything else related to lua. 

The goal of issue #9 is to have all lua-related stuff in the lua layer and have the `framework` and `modules` agnostic of the outer layer using it. So things like a python wrapper and whatnot will sit in their own location using the C++ API.

andrsd: In case you want to check how the code is going to be structured to accomplish the lua/core separation, you can look at my [next branch](https://github.com/andrsd/opensn/tree/next) in my fork...

zhardy-lanl: What does everyone think of `opensn::lua` instead? It seems cleaner than `opensn<api>`. I know we agreed on a single layer, but for APIs, I think a single nesting makes sense here.

ragusa (approved):

PR #56, opened 2023-12-08
File I/O Cleanup

zhardy-lanl:
This PR cleans up and adds to existing file I/O. Changes in reference to discussion #54 will come later. This PR will eventually be followed by the `ResponseEvaluator` class which uses the new angular flux binaries for evaluating boundary responses.

### Primary Changes

- Some types changed to be more appropriate for the quantity (mainly going from `size_t`/`uint64_t` to `unsigned int` for quantities like group/moment/angle/cell node indices.
- Chi-Tech references changed to OpenSn
- Added routine to write angular fluxes per groupset into a single file

### Notes

- Binary files produced before this PR will no longer work due to the type changes. Any binaries will need to be regenerated if needed after this PR. 
- Draft status while waiting on #48 since a namespace change will be required immediately afterwards.


modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -2100,180 +2100,329 @@ LBSSolver::ReadRestartData(const std::string& folder_name, const std::string& fi
...
         {
-          if (++dof_count > num_local_dofs) goto close_file;
-
-          uint64_t dof_map = sdm->MapDOFLocal(cell, i, dof_handler, n, g);
-          double value = psi_new_local_[groupset.id_][dof_map];
+          const uint64_t dof_map = discretization_->MapDOFLocal(cell, i, uk_man, n, g);
+          const double value = src[dof_map];
 
-          file.write((char*)&cell.global_id_, sizeof(size_t));
+          file.write((char*)&cell.global_id_, sizeof(uint64_t));
```
ragusa, line 2349: what's the rule we follow for size_t versus other denominations of sizes?

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -438,7 +438,7 @@ LBSSolver::OptionsBlock()
   "The eager limit is the message size limit before which non-blocking MPI send"
   "calls will execute without waiting for a matching receive call. The limit is"
   "platform dependent but in general 64 kb. Some systems have 32 kb as a limit"
-  "and therefore we use that as a default limit in ChiTech. There is a fine"
+  "and therefore we use that as a default limit in OpenSn. There is a fine"
```
ragusa, line 441: I thought the name is openSn. Did we change that?

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -2100,180 +2100,329 @@ LBSSolver::ReadRestartData(const std::string& folder_name, const std::string& fi
...
+
+  file << header_bytes;
+
+  // Write macro info
+  auto NODES_ONLY = chi_math::UnknownManager::GetUnitaryUnknownManager();
+
+  const uint64_t num_local_nodes = discretization_->GetNumLocalDOFs(NODES_ONLY);
+  const unsigned int num_groupsets = groupsets_.size();
+
+  file.write((char*)&num_local_nodes, sizeof(uint64_t));
```
ragusa, line 2147: is the local of nodes in need of such a large size? just wondering about the different sizes of int we use.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -2100,180 +2100,329 @@ LBSSolver::ReadRestartData(const std::string& folder_name, const std::string& fi
...
+    const auto& uk_man = groupset.psi_uk_man_;
+    const auto& quadrature = groupset.quadrature_;
+
+    const unsigned int groupset_id = groupset.id_;
+    const unsigned int num_gs_angles = quadrature->omegas_.size();
+    const unsigned int num_gs_groups = groupset.groups_.size();
+
+    file.write((char*)&groupset_id, sizeof(unsigned int));
+    file.write((char*)&num_gs_angles, sizeof(unsigned int));
+    file.write((char*)&num_gs_groups, sizeof(unsigned int));
```
ragusa, line 2163: looks a bit inelegant. are there smoother ways these days? @andrsd  ?

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -2100,180 +2100,329 @@ LBSSolver::ReadRestartData(const std::string& folder_name, const std::string& fi
...
   std::ofstream file(file_name,
                      std::ofstream::binary |  // binary file
                        std::ofstream::out |   // no accidental reading
-                       std::ofstream::trunc); // clear file contents when opened
+                       std::ofstream::trunc); // clear contents first
+  ChiLogicalErrorIf(not file.is_open(), "Failed to open " + file_name + ".");
+  Chi::log.Log() << "Writing angular flux to " << file_base;
+
+  // Write the header
+  const int num_bytes = 500;
```
ragusa, line 2116: is 500 always enough? how was this chosen? I know it was there before, just wondering.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
andrsd: I recommend to use 64bit numbers consistently through out the whole code for things like this. 
1. We run on 64 bit architectures these days. 
2. It will prevent problems with narrowing conversions.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
zhardy-lanl: There is no hard and fast rule used int he code today. We should discuss adopting some rules for this. My thought here to use `uint64_t` is to be as explicit as possible.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
zhardy-lanl: Old habits die hard with the capital "O". The org is "O" while the repo is "o". I will change this to "o".

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
zhardy-lanl: The number of local nodes is the number of local spatial discretization nodes, or number of cells times nodes per cell. It can be quite big. Something like `num_cell_nodes` would be small.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
zhardy-lanl: @ragusa The binaries are written in this fashion already, and I agree it is not terribly elegant. This will get better as time goes on.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
zhardy-lanl: Yes this is just for the fixed size `header_info` string and not for the data.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -2288,282 +2437,237 @@ LBSSolver::MakeSourceMomentsFromPhi()
...
+  // Check compatibility with system macro info
+  const auto uk_man = flux_moments_uk_man_;
+  const auto NODES_ONLY = UnknownManager::GetUnitaryUnknownManager();
+
+  const uint64_t num_local_cells = grid_ptr_->local_cells.size();
+  const uint64_t num_local_nodes = discretization_->GetNumLocalDOFs(NODES_ONLY);
+  const auto num_local_dofs = discretization_->GetNumLocalDOFs(uk_man);
+
+  // clang-format off
+  ChiLogicalErrorIf(file_num_local_cells != num_local_cells,
```
ragusa, line 2579: @andrsd is there an upcoming PR with more name changes, for things like these?

ragusa (approved):

ragusa (approved):

zhardy-lanl:
Sorry for the many different changes. Rebased instead of merged, and then had to fix a formatting issue. This is ready now.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
andrsd: Yes, there will be a PR (or to be more precise PRs(!)) for things like this.

andrsd (approved):
:+1: on the clean git history

PR #63, opened 2024-01-02
Removing references to Chi

andrsd:
- Renaming ChiObject to Object
- Renaming ChiObjectFactory to ObjectFactory
- Renaming ChiLog to Logger
- MPI_Info is a global variable
- `console` is a global variable
- `log` is now a global variable
- Moving memory usage code in its own location
- Exit function is in opensn namespace
- GetVersionStr function is in opensn namespace
- Initialize and Finalize are in opensn namespace
- 'program_timer' is a global variable
- Renaming ChiMPICommunicatorSet to MPICommunicatorSet
- Replacing ChiTech references to openSn
- Removing friendship between MPI_Info and Chi
- Renaming macros for registering objects
- Renaming parameter 'chi_obj_type' to 'obj_type'


andrsd:
This is part 1 of removing references to Chi/ChiTech. There will be 2 more PRs like this one in future. Just a heads-up that not all references are removed in this one...

PR #66, opened 2024-01-06
Improved leakage calculations

zhardy-lanl:
This PR adds more generalized routines for calculating leakages from boundaries. 

### Highlights 
- Added the `ComputeLeakage` routine to `lbs::DiscreteOrdinatesSolver`. This routine takes a vector of boundary IDs and returns a map of boundary IDs to group-wise leakage values.
- Added the Lua wrapper function `lbs.ComputeLeakage`. The function takes a handle to the solver and optionally an array of boundary names (the standard ones used in OpenSn). The return value in Lua is a table of tables keyed with the boundary names and then the group indices.
- Added a simple leakage test, which gives the correct analytic solution.

### Comments
- While the test is in parallel, the test does not rigorously test the communication carried out in the new routine. This would require a test where a a single boundary is shared among multiple processors, which is more difficult for analytic solutions to compare to.

### DIFF
--- a/lua/modules/linear_bolzmann_solvers/discrete_ordinates_solver/lbs_compute_leakage.cc
+++ b/lua/modules/linear_bolzmann_solvers/discrete_ordinates_solver/lbs_compute_leakage.cc
@@ -1,12 +1,11 @@
 #include "lbs_do_lua_utils.h"
 
-#include "modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.h"
-
+#include "framework/runtime.h"
+#include "framework/logging/log.h"
 #include "framework/console/console.h"
 
-#include "framework/runtime.h"
-
-using namespace opensn;
+#include "framework/mesh/mesh_continuum/mesh_continuum.h"
+#include "modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.h"
 
 namespace opensnlua::lbs
 {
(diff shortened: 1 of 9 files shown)


ragusa:
@zhardy-lanl  formatting check failed.

zhardy-lanl:
> @zhardy-lanl formatting check failed.
One of these days I will remember to do the clang-format before submitting... Should be good now I hope.

lua/modules/linear_bolzmann_solvers/discrete_ordinates_solver/lbs_compute_leakage.cc
```diff
@@ -49,4 +48,66 @@ chiLBSComputeLeakage(lua_State* L)
...
+
+  // Get the boundaries to parse
+  std::vector<uint64_t> bndry_ids;
+  if (num_args > 1)
+  {
+    LuaCheckTableValue(fname, L, 2);
+
+    // Define mapping from boundary name to id
+    const std::map<std::string, uint64_t> bndry_map = {
+      {"xmax", 0}, {"xmin", 1}, {"ymax", 2}, {"ymin", 3}, {"zmax", 4}, {"zmin", 5}};
```
andrsd, line 73: This map is already in [`lbs_solver.cc`](https://github.com/Open-Sn/opensn/blob/9f4c021ea8564f650895129919664834434e2f6b/modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc#L655-L656), but as a local variable. If the client code needs to be able to use it, then the local variable `supported_bndry_names` from `SetBoundaryOptions` should be a public static of the `LBSSolver` and renamed to something like `boundary_id_map` (or `supported_boundary_id_name`).

lua/modules/linear_bolzmann_solvers/discrete_ordinates_solver/lbs_compute_leakage.cc
```diff
@@ -49,4 +48,66 @@ chiLBSComputeLeakage(lua_State* L)
...
+  }
+  else
+    bndry_ids = solver.Grid().GetDomainUniqueBoundaryIDs();
+
+  // Compute the leakage
+  const auto leakage = solver.ComputeLeakage(bndry_ids);
+
+  // Define mapping from boundary ids to names
+  const std::map<uint64_t, std::string> bndry_map = {
+    {0, "xmax"}, {1, "xmin"}, {2, "ymax"}, {3, "ymin"}, {4, "zmax"}, {5, "zmin"}};
```
andrsd, line 93: Similarly to the above comment, this might need to go into the `LBSSolver` as a public static, so it is consistent.

andrsd (changes requested):

andrsd:
> One of these days I will remember to do the clang-format before submitting... Should be good now I hope.

In CLion (IIRC you are using it), you can to 2 things for this:

1. In **Preferences | Tools | Action On Save**, check **Reformat code**  and you can also pick for what file types.
2. In **Preferences | Version control | Commit | Commit checks**, you can check **Reformat code** (but this is only if you use git via the CLion GUI) It is also possible to add a git pre-commit hook into your git terminal client , but it is only per repo IIRC.

I personally use option 1 (with only C/C++ files), because I developed a paranoid habit of saving files constantly :-)

lua/modules/linear_bolzmann_solvers/discrete_ordinates_solver/lbs_compute_leakage.cc
zhardy-lanl: I agree with this. I think in an ideal world, there is a public static member `std::vector<std::string> supported_boundary_names` and the mappings are internal. The natural inclination would be to make the `ComputeLeakage` routine accept a vector of strings. This doesn't jive with future plans @ragusa and I have discussed to have arbitrary boundary IDs to support more complex boundary conditions (e.g. prescribing boundary conditions to portions of boundaries). Opened issue #70 for this. I will make the maps public for the time being.

zhardy-lanl:
> > One of these days I will remember to do the clang-format before submitting... Should be good now I hope.
> 
> In CLion (IIRC you are using it), you can to 2 things for this:
> 
> 1. In **Preferences | Tools | Action On Save**, check **Reformat code**  and you can also pick for what file types.
> 2. In **Preferences | Version control | Commit | Commit checks**, you can check **Reformat code** (but this is only if you use git via the CLion GUI) It is also possible to add a git pre-commit hook into your git terminal client , but it is only per repo IIRC.
> 
> I personally use option 1 (with only C/C++ files), because I developed a paranoid habit of saving files constantly :-)

I didn't know about that. That will help. Thank you!

andrsd (approved):

ragusa (approved):
Approved but please reply to comments. Thanks!

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -50,6 +50,13 @@ namespace opensn
 {
 namespace lbs
 {
+
+std::map<std::string, uint64_t> LBSSolver::supported_boundary_names = {
+  {"xmax", 0}, {"xmin", 1}, {"ymax", 2}, {"ymin", 3}, {"zmax", 4}, {"zmin", 5}};
```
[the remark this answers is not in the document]
andrsd: Issue #70

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -727,10 +720,117 @@ DiscreteOrdinatesSolver::ComputeLeakage(const int groupset_id, const uint64_t bo
...
+  return global_leakage;
+}
+
+std::map<uint64_t, std::vector<double>>
+DiscreteOrdinatesSolver::ComputeLeakage(const std::vector<uint64_t>& boundary_ids) const
+{
+  // Perform checks
+  ChiLogicalErrorIf(not options_.save_angular_flux,
+                    "The option `save_angular_flux` must be set to `true` in order "
+                    "to compute outgoing currents.");
```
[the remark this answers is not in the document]
zhardy-lanl: Did some digging and as of right now... There's no simple way to access *face-wise* stored angular fluxes. For balance, there is outgoing data accumulated on boundary cells, but it does not discriminate on a per-face basis, which means it cannot be used for leakage at present. Opening #74 for this.

zhardy-lanl:
@ragusa I can try to rework the outgoing angular flux information here, or in a later PR. Up to you. If you are fine pushing this to a later PR, please merge.

PR #76, opened 2024-01-18
Macro tidy up

andrsd:
Things that are done:
- change macros into enums when they are used in an enum sense
- remove macros that obfuscate the code
- properly define types via `typedef` instead of macro

### DIFF
--- a/framework/math/parallel_vector/parallel_stl_vector.cc
+++ b/framework/math/parallel_vector/parallel_stl_vector.cc
@@ -13,10 +13,6 @@
 #include <cmath>
 #include <numeric>
 
-#define scint64_t static_cast<int64_t>
-#define scint static_cast<int>
-#define scshort static_cast<short>
-
 namespace opensn
 {
 
@@ -81,7 +77,7 @@ ParallelSTLVector::LocalSTLData()
 std::vector<double>
 ParallelSTLVector::MakeLocalVector()
 {
-  return std::vector<double>(values_.begin(), values_.begin() + scint(local_size_));
+  return std::vector<double>(values_.begin(), values_.begin() + static_cast<int>(local_size_));
 }
 
 double
@@ -117,7 +113,9 @@ ParallelSTLVector::Set(const std::vector<double>& local_vector)
 
   // We cannot assign values = local_vector because this might
   // destroy the internals of a ghosted vector
-  std::copy(local_vector.begin(), local_vector.begin() + scint64_t(local_size_), values_.begin());
+  std::copy(local_vector.begin(),
+            local_vector.begin() + static_cast<int64_t>(local_size_),
+            values_.begin());
 }
 
 void
@@ -326,7 +324,7 @@ ParallelSTLVector::ComputeNorm(NormType norm_type) const
     case NormType::L1_NORM:
     {
       const double local_norm_val =
-        std::accumulate(values_.begin(), values_.begin() + scint(local_size_), 0.0);
+        std::accumulate(values_.begin(), values_.begin() + static_cast<int>(local_size_), 0.0);
 
       double global_norm_val;
       MPI_Allreduce(&local_norm_val, &global_norm_val, 1, MPI_DOUBLE, MPI_SUM, comm_);
@@ -349,7 +347,7 @@ ParallelSTLVector::ComputeNorm(NormType norm_type) const
     case NormType::LINF_NORM:
     {
       const double local_norm_val =
-        *std::max_element(values_.begin(), values_.begin() + scint(local_size_));
+        *std::max_element(values_.begin(), values_.begin() + static_cast<int>(local_size_));
 
       double global_norm_val;
       MPI_Allreduce(&local_norm_val, &global_norm_val, 1, MPI_DOUBLE, MPI_MAX, comm_);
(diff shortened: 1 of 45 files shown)


andrsd:
Some of this may be controversial, so let me know if you want to retain some macro in some cases. IMO most of the use just obfuscates the code and is trying to save the amount of keystrokes one has to do which is just plain wrong usage of macros.

As for the error handling, I went for plain `throw exception` - I think the whole approach needs more thought and then make the modifications in the entire code. But that is a discussion on its own...

ragusa (approved):
it's clearer to me without those macros

framework/math/parallel_vector/parallel_stl_vector.cc
```diff
@@ -81,7 +77,7 @@ ParallelSTLVector::LocalSTLData()
 std::vector<double>
 ParallelSTLVector::MakeLocalVector()
 {
-  return std::vector<double>(values_.begin(), values_.begin() + scint(local_size_));
+  return std::vector<double>(values_.begin(), values_.begin() + static_cast<int>(local_size_));
```
zhardy-lanl, line 80: For my own edification... Are static casts necessary in scenarios like this? I see them somewhat inconsistently sprinkled throughout the code and I'm never sure whether I should use them or not. This is another standardization thing to tackle in the future.

framework/math/spatial_discretization/finite_volume/finite_volume.cc
```diff
@@ -8,15 +8,12 @@
...
-#define MappingError                                                                               \
-  "SpatialDiscretization_FV::OrderNodes: "                                                         \
-  "Error mapping neighbor cells"
-
 namespace opensn
 {
 
+static std::string mapping_error =
+  "SpatialDiscretization_FV::OrderNodes: Error mapping neighbor cells";
+
```
zhardy-lanl, line 16: It looks like this is only used in `OrderNodes`. Should we define this inside the routine instead?

framework/mesh/logical_volume/rpp_logical_volume.h
```diff
@@ -8,6 +8,16 @@ namespace opensn
 /**Rectangular Parallel Piped (RPP) logical volume*/
 class RPPLogicalVolume : public LogicalVolume
 {
+  enum ConditionDirection
```
zhardy-lanl, line 11: Not a big fan of the name. How about something like `CornerName` since this is defining the corners of a RPP.

modules/diffusion_solver/diffusion_solver.h
```diff
@@ -39,6 +33,15 @@ class Solver : public opensn::Solver
...
   bool verbose_info_ = true;
 
+  enum MaterialMode
+  {
+    DIFFUSION_MATERIALS_REGULAR = 10,
+    DIFFUSION_MATERIALS_FROM_TRANSPORTXS_TTR = 11,
+    DIFFUSION_MATERIALS_FROM_TRANSPORTXS_TTF = 12,
+    DIFFUSION_MATERIALS_FROM_TRANSPORTXS_TTF_JPART = 13,
+    DIFFUSION_MATERIALS_FROM_TRANSPORTXS_TTF_JFULL = 14
+  };
```
zhardy-lanl, line 43: While the names are old, since this is in a diffusion solver and now under the `MaterialMode` enum, how about removing the `DIFFUSION_MATERIALS` prefixes?

modules/linear_boltzmann_solvers/b_diffusion_dfem_solver/iterative_methods/mip_wgs_context2.cc
```diff
@@ -12,14 +12,13 @@
...
 
-#define sc_double static_cast<double>
-#define PCShellPtr PetscErrorCode (*)(PC, Vec, Vec)
-
 namespace opensn
 {
 namespace lbs
 {
 
+typedef PetscErrorCode (*PCShellPtr)(PC, Vec, Vec);
```
zhardy-lanl, line 20: We should discuss `using` vs `typedef` at some point

zhardy-lanl (approved):
This is a good first step. Lots to discuss about standardization in the future. Please see requests for naming changes.

modules/linear_boltzmann_solvers/b_diffusion_dfem_solver/iterative_methods/mip_wgs_context2.cc
andrsd: I think we should be using `using`. This is just an old habit of mine from pre-C++11 days.

framework/math/parallel_vector/parallel_stl_vector.cc
andrsd: `clang-tidy` complains about narrowing conversions, so I think that's where these casts are coming from. In CLion, you should be able to see those "warning" in the [Problems tool](https://www.jetbrains.com/help/clion/problems-tool-window.html)

modules/diffusion_solver/diffusion_solver.h
andrsd: :+1: Change made.

framework/mesh/logical_volume/rpp_logical_volume.h
andrsd: Change made.

wdhawkins (approved):

PR #79, opened 2024-01-19
XS Scaling Factors

zhardy-lanl:
This PR introduces scaling factors for cross sections. Existing code capabilities does not store any information about the scalings that are applied. In effect, the `MakeCombined` routine is used to generate macroscopic cross sections that remain fixed throughout a simulation. Subsequent calls to `MakeCombined` compound the scaling factors, which is not the desired behavior when trying to change something like the density applied to a microscopic cross section. To obtain the correct behavior, a number of things were implemented:

- `scaling_factor_` member added to cross sections with accessor.
- `SetScalingFactor` multiplies cross sections by the ratio of the new scaling factor to the old and sets `scaling_factor_` to the new value.
- `MakeCombined` calls `SetScalingFactor` with the specified factor and combines accordingly
    - The resulting XS set from `MakeCombined` will have a scaling factor of 1. For a single cross section set, one must use the original XS set that gets the scaling factor applied, not the one returned. 
    - For homogenizing isotopes and applying a bulk density, one must call `MakeCombined` to combine isotopes and second time to scale by the bulk density. The same principle applies to the previous bullet point.
    - This system needs improvement, and `MakeCombined` sneakily uses Lua information and should be moved in the future.

### DIFF
--- a/framework/physics/physics_material/multi_group_xs/adjoint_mgxs.cc
+++ b/framework/physics/physics_material/multi_group_xs/adjoint_mgxs.cc
@@ -5,7 +5,7 @@ namespace opensn
 
 AdjointMGXS::AdjointMGXS(const MultiGroupXS& xs) : xs_(xs)
 {
-  // transpose transfer matrices
+  // Transpose transfer matrices
   for (unsigned int ell = 0; ell <= xs_.ScatteringOrder(); ++ell)
   {
     const auto& S_ell = xs_.TransferMatrix(ell);
@@ -22,17 +22,15 @@ AdjointMGXS::AdjointMGXS(const MultiGroupXS& xs) : xs_(xs)
     transposed_transfer_matrices_.push_back(S_ell_transpose);
   } // for ell
 
-  // transpose production matrices
+  // Transpose production matrices
   if (xs_.IsFissionable())
   {
+    transposed_production_matrices_.clear();
+    transposed_production_matrices_.resize(xs_.NumGroups());
     const auto& F = xs_.ProductionMatrix();
     for (size_t g = 0; g < xs_.NumGroups(); ++g)
-    {
-      std::vector<double> F_g_transpose;
       for (size_t gp = 0; gp < xs_.NumGroups(); ++gp)
-        F_g_transpose.emplace_back(F[gp][g]);
-      transposed_production_matrices_.push_back(F_g_transpose);
-    }
+        transposed_production_matrices_[g].push_back(F[gp][g]);
   }
 }
 
--- a/framework/physics/physics_material/multi_group_xs/multi_group_xs.cc
+++ b/framework/physics/physics_material/multi_group_xs/multi_group_xs.cc
@@ -8,13 +8,12 @@ namespace opensn
 {
 
 void
-MultiGroupXS::ExportToChiXSFile(const std::string& file_name, const double fission_scaling) const
+MultiGroupXS::ExportToChiXSFile(const std::string& file_name,
+                                const double fission_scaling /* = 1.0 */) const
 {
   log.Log() << "Exporting transport cross section to file: " << file_name;
 
-  // Define utility functions
-
-  /**Lambda to print a 1D-xs*/
+  // Lambda to print a 1D cross section
   auto Print1DXS = [](std::ofstream& ofile,
                       const std::string& prefix,
                       const std::vector<double>& xs,
@@ -36,7 +35,7 @@ MultiGroupXS::ExportToChiXSFile(const std::string& file_name, const double fissi
     ofile << "\n";
     ofile << prefix << "_BEGIN\n";
     {
-      unsigned int g = 0;
+      size_t g = 0;
       for (auto val : xs)
         ofile << g++ << " " << val << "\n";
     }
@@ -47,39 +46,17 @@ MultiGroupXS::ExportToChiXSFile(const std::string& file_name, const double fissi
   std::ofstream ofile(file_name);
 
   // Write the header info
-  std::vector<double> nu, nu_prompt, nu_delayed;
-  const auto& sigma_f = SigmaFission();
-  const auto& nu_sigma_f = NuSigmaF();
-  const auto& nu_prompt_sigma_f = NuPromptSigmaF();
-  const auto& nu_delayed_sigma_f = NuDelayedSigmaF();
-  for (unsigned int g = 0; g < NumGroups(); ++g)
-  {
-    if (NumPrecursors() > 0)
-    {
-      nu_prompt.push_back(nu_prompt_sigma_f[g] / sigma_f[g]);
-      nu_delayed.push_back(nu_delayed_sigma_f[g] / sigma_f[g]);
-    }
-    nu.push_back(nu_sigma_f[g] / sigma_f[g]);
-  }
-
-  std::vector<double> decay_constants, fractional_yields;
-  for (const auto& precursor : Precursors())
-  {
-    decay_constants.push_back(precursor.decay_constant);
-    fractional_yields.push_back(precursor.fractional_yield);
-  }
-
   ofile << "# Exported cross section from ChiTech\n";
   ofile << "# Date: " << Timer::GetLocalDateTimeString() << "\n";
   ofile << "NUM_GROUPS " << NumGroups() << "\n";
   ofile << "NUM_MOMENTS " << ScatteringOrder() + 1 << "\n";
   if (NumPrecursors() > 0) ofile << "NUM_PRECURSORS " << NumPrecursors() << "\n";
 
-  // basic cross section data
+  // Basic cross section data
   Print1DXS(ofile, "SIGMA_T", SigmaTotal(), 1.0e-20);
   Print1DXS(ofile, "SIGMA_A", SigmaAbsorption(), 1.0e-20);
 
-  // fission data
+  // Fission data
   if (not SigmaFission().empty())
   {
     std::vector<double> scaled_sigma_f = SigmaFission();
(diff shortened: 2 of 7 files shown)


framework/physics/physics_material/multi_group_xs/multi_group_xs.cc
```diff
@@ -8,13 +8,14 @@ namespace opensn
...
-MultiGroupXS::ExportToChiXSFile(const std::string& file_name, const double fission_scaling) const
+MultiGroupXS::ExportToChiXSFile(const std::string& file_name,
+                                const double fission_scaling /* = 1.0 */) const
 {
   log.Log() << "Exporting transport cross section to file: " << file_name;
 
   // Define utility functions
 
-  /**Lambda to print a 1D-xs*/
+  /// Lambda to print a 1D cross section
```
andrsd, line 18: Cosmetic: I would go with `//` here

framework/physics/physics_material/multi_group_xs/multi_group_xs.cc
```diff
@@ -92,32 +71,60 @@ MultiGroupXS::ExportToChiXSFile(const std::string& file_name, const double fissi
...
+      for (const auto& precursor : Precursors())
+      {
+        lambda.emplace_back(precursor.decay_constant);
+        gamma.emplace_back(precursor.fractional_yield);
+      }
+
       Print1DXS(ofile, "NU_PROMPT", nu_prompt, 1.0e-20);
       Print1DXS(ofile, "NU_DELAYED", nu_delayed, 1.0e-20);
-      //      Print1DXS(ofile, "CHI_PROMPT", chi_prompt, 1.0e-20);
+      // Print1DXS(ofile, "CHI_PROMPT", chi_prompt, 1.0e-20);
```
andrsd, line 96: I personally recommend to delete lines like this. If is not used remove it. When it is needed have a proper change set doing the right thing. Lines like this like to rot over time.

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
```diff
@@ -83,12 +81,12 @@ SingleStateMGXS::MakeSimple1(unsigned int num_groups, double sigma_t, double c)
   //        of the time.
   //     3) The lowest energy group has the same form as 1).
 
-  for (unsigned int g = 0; g < num_groups_; ++g)
+  for (size_t g = 0; g < num_groups_; ++g)
   {
-    // downscattering
+    // Dwnscattering
```
andrsd, line 86: Typo: "Downscattering" (missing `o`)

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
```diff
@@ -786,63 +718,62 @@ SingleStateMGXS::MakeFromChiXSFile(const std::string& file_name)
...
+            "Delayed neutron precursor fractional yields must be in the range [0.0, 1.0].");
+
+          // Normalizing
+          const auto sum = std::accumulate(fractional_yields.begin(), fractional_yields.end(), 0.0);
           std::transform(fractional_yields.begin(),
                          fractional_yields.end(),
                          fractional_yields.begin(),
                          [sum](double& x) { return x / sum; });
-        }
+        } // if precursor yieldx
```
andrsd, line 771: Typo in `yieldx` (extra `x`) ?

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
```diff
@@ -887,225 +818,216 @@ SingleStateMGXS::MakeFromChiXSFile(const std::string& file_name)
...
-
-        // add delayed fission neutron yield to total
-        for (unsigned int g = 0; g < num_groups_; ++g)
+        // Check that decay data was specified
+        ChiLogicalErrorIf(decay_constants.empty(),
+                          "Precursor decay constants are required when precursors are specified.");
+
+        // Check that yield data was specified
+        ChiLogicalErrorIf(fractional_yields.empty(),
+                          "Precursor yields are required when precursors are specified..");
```
andrsd, line 871: Double `.` at the end.

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
```diff
@@ -887,225 +818,216 @@ SingleStateMGXS::MakeFromChiXSFile(const std::string& file_name)
...
+      // TODO: Develop an implementation for multi-particle delayed neutron data.
+      //       The primary challenge in this is that different precursor species exist for
+      //       neutron-induced fission than for photo-fission.
 
-      // throw error if precursors are present
-      if (num_precursors_ > 0)
-        throw std::runtime_error("This setup has not been implemented.\n"
-                                 "Currently, when a production matrix is specified, "
-                                 "no delayed neutron precursors are allowed.");
+      // Not implemented error
```
andrsd, line 944: This comment seems incorrect - it is not implemented error

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
```diff
@@ -1115,51 +1037,50 @@ SingleStateMGXS::ComputeDiffusionParameters()
...
-  // initialize diffusion data
+  // Initialize diffusion data
+  sigma_tr_.resize(num_groups_, 0.0);
   diffusion_coeff_.resize(num_groups_, 1.0);
   sigma_s_gtog_.resize(num_groups_, 0.0);
-  sigma_removal_.resize(num_groups_, 0.1);
+  sigma_r_.resize(num_groups_, 0.0);
 
-  // perfom computations group-wise
+  // Perfom computations group-wise
```
andrsd, line 1046: Typo: Perfo**r**m (missing `r`)

andrsd (changes requested):

framework/physics/physics_material/multi_group_xs/multi_group_xs.cc
zhardy-lanl: Fair point. It will likely come back at a later date, but we can always add it back, when needed.

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
zhardy-lanl: Well this is obvious, haha. I didn't want to go create a macro for not implemented error so I just shortcut it. I will use a proper error.

andrsd: My point is that you correctly handled an error state - that's is not "not implemented" error IMO. So, just remove the comment, if you plan to add something more in future. This check would be replaced, then with more code.

andrsd (approved):

framework/physics/physics_material/multi_group_xs/adjoint_mgxs.cc
```diff
@@ -22,17 +22,15 @@ AdjointMGXS::AdjointMGXS(const MultiGroupXS& xs) : xs_(xs)
...
+    transposed_production_matrices_.resize(xs_.NumGroups());
     const auto& F = xs_.ProductionMatrix();
     for (size_t g = 0; g < xs_.NumGroups(); ++g)
-    {
-      std::vector<double> F_g_transpose;
       for (size_t gp = 0; gp < xs_.NumGroups(); ++gp)
-        F_g_transpose.emplace_back(F[gp][g]);
-      transposed_production_matrices_.push_back(F_g_transpose);
-    }
+        transposed_production_matrices_[g].push_back(F[gp][g]);
```
ragusa, line 33: I suppose we have an adjoint test case that has a fissile material? I am just worried about changes to the code that do not have a test case. If ti does not have one, can you add one. I can help verify the solution if need be, I like to do that sort of things!

framework/physics/physics_material/multi_group_xs/multi_group_xs.h
```diff
@@ -32,46 +32,30 @@ class MultiGroupXS : public MaterialProperty
    */
   void ExportToChiXSFile(const std::string& file_name, const double fission_scaling = 1.0) const;
```
ragusa, line 33: @andrsd is this name scheduled to be changed in an upcoming PR?

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
```diff
@@ -110,37 +108,38 @@ SingleStateMGXS::MakeCombined(std::vector<std::pair<int, double>>& combinations)
...
   {
-    // get the cross section from the lua stack
+    // Get the cross section from the stack
     std::shared_ptr<MultiGroupXS> xs;
     xs = Chi::GetStackItemPtr(Chi::multigroup_xs_stack, combo.first, std::string(__FUNCTION__));
     xsecs.push_back(xs);
 
-    // increment densities
+    // Set the scaling factor
+    std::dynamic_pointer_cast<SingleStateMGXS>(xs)->SetScalingFactor(combo.second);
```
ragusa, line 128: curious: why do we cast combo.second now?

framework/physics/physics_material/multi_group_xs/single_state_mgxs.h
```diff
@@ -80,6 +42,14 @@ class SingleStateMGXS : public MultiGroupXS
    */
   void MakeCombined(std::vector<std::pair<int, double>>& combinations);
 
+  /**
+   * Scale the cross sections by the specified factor.
+   *
+   * @note Scaling factors do not compound. Each time this routine is called, the cross sections
+   *       are scaled by the ratio of the argument and the existing scaling factor.
```
ragusa, line 49: so scaling is always, in the end, the multiplier you want.

if the current scaling was 2 (meaning a factor 2 had already been applied) and you now give 3, then the data is multiplied by 1.5, making it effectively a factor 3 since the beginning. Right?

ragusa (approved):
just a few comments/questions for my own edification. please comment back.

framework/physics/physics_material/multi_group_xs/multi_group_xs.h
andrsd: Yes.

framework/physics/physics_material/multi_group_xs/adjoint_mgxs.cc
zachhardy: I do not believe we do. Could we table this on this PR? I have many changes coming to the adjoint capabilities in subsequent PRs. Since this has been in the code for awhile without a test, I would prefer to wait until the new adjoint capabilities are in for a test rather than making a test and changing it within a week or two.

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
zachhardy: We are not casting `combo.second` we are casting `xs` here. The `MultiGroupXS` class is abstract and does not have many of the methods, so here we cast to the concrete class `SingleStateMGXS` where `SetScalingFactor` is implemented. I am open to discussions on reworking the cross sections.

framework/physics/physics_material/multi_group_xs/single_state_mgxs.h
zachhardy: Exactly! This allows us to always have a mapping back to microscopic cross sections. The scaling factor is what gets applied to the _unscaled_ cross sections

wdhawkins (approved):
Looks good to me. We discussed my concerns in our earlier meeting and created an issue, so no need to add a bunch of extraneous comments.

framework/physics/physics_material/multi_group_xs/single_state_mgxs.cc
ragusa: ok. @andrsd what is your opinion of l.128? I am ignorant of such matters.

andrsd: If your concern is about `dynamic_cast`, then they exist for a reason. But for example, you do not want to do that inside a quadrature point loop.

So, this is ok, unless I missed something. If refactoring happens in cross-section classes, this can be changed then. No need to stress about this...

PR #80, opened 2024-01-19
Improving `configure_dependencies.py`

andrsd:
- functions are defined before any code happens
- errors are indicated by exceptions
- script stops on first error encountered rather then trying to continue from a incorrect state (i.e. do not try to compile if configure step failed)
- all packages are installed into install dir unix style, i.e. header files go to `${INSTALL}/include`, libs go to `${INSTALL}/lib`, etc.
- logs are now in `${INSTALL}/logs`
- sources are unpacked and compiled from `${INSTALL}/src`
- `PETSC_ROOT` is changed to `PETSC_DIR`
- User is informed about setting `CMAKE_PREFIX_PATH` which is needed for successful cmake build of OpenSn
- Fixed typos
- script for setting env vars is called `set_opensn_env.sh` and is located in `${INSTALL}/bin`

### DIFF
--- a/resources/configure_dependencies.py
+++ b/resources/configure_dependencies.py
@@ -1,7 +1,20 @@
 """
 This is a utility script for the download and installation of OpenSn
 dependencies.
+
+NOTES:
+- If you see errors in building fblaslapack, try to use different BLAS/LAPACK. For example, on MacOSX, the Apple
+  supplied ones are known to work (so you would just remove the option --download-fblaslapack=1)
+- If you have set CC, CXX and FC in your environment, make sure they are CC=mpicc, CXX=mpicxx and FC=mpifort.
+
+NOTES for clang compiler:
+- clang compiler does not support `-march=native`, so you will want to remove that from COPTFLAGS, CXXOPTFLAGS and
+  FOPTFLAGS.
+
+NOTES for building on MacOS:
+- you may need to `export MACOSX_DEPLOYMENT_TARGET=10.15` in your environment
 """
+
 import os
 import sys
 import argparse
@@ -33,7 +46,7 @@ class TextColors:
 error_beg = TextColors.RED
 error_end = TextColors.ENDC
 
-# ========================================== Process commandline arguments
+# Process commandline arguments
 arguments_help = textwrap.dedent('''\
 Run the dependency builder. 
 ''')
@@ -68,20 +81,7 @@ parser.add_argument(
     help="Controls verbose failure"
 )
 
-argv = parser.parse_args()  # argv = argument values
-
-# =========================================== Check stuff
-# Check the test directory exists
-if not os.path.isdir(argv.directory):
-    raise NotADirectoryError(argv.directory)
-
-install_dir = os.path.abspath(argv.directory)
-print(f"Installation directory identified: {install_dir}")
-print()
-log_file = open(f"{install_dir}/log.txt", 'w')
-log_file.write("***** ChiTech dependency configurator script *****\n\n")
-
-# ========================================== Define versions and download paths
+# Define versions and download paths
 VERSION = 0
 URL = 1
 package_info = {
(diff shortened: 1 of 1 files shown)


andrsd:
I made some significant changes in the script that builds the OpenSn dependencies. I hope I improved things rather than breaking more stuff :-)

I seems that the script was written to work with only `gcc` - I had some interesting problems with clang on Mac with arm64. I had to make some local changes to the flags that are passed into PETSc to build (these are not included in this change set). IDK if it worth the effort, but if desired I can improve this script further if this is the recommended way to build the dependencies - I personally do not use this script...

ragusa (commented):
at the end of the script, there was some useful instructions. We may want to keep some, such as: 
* you may want to export the following env vars to your .bashrc or .bash_profile (lua_root, vtk_dir, petsc_root
as a suggestion. you may want to add another suggestion to look into the installation instruction (the full versions as we will delete the simplify instructions) about adding the CMAKE_PREFIX ....

andrsd:
> at the end of the script, there was some useful instructions. 

They are still [there](https://github.com/Open-Sn/opensn/pull/80/files#diff-b680d8ca21142ce9d83fa41850e2e0cdcba6f68d8bed23d99189a94e22ec35b5R696-R716), it is just a bit lost in the diff. They are also updated and tell the user about `CMAKE_PREFIX_PATH`, etc.

ragusa (approved):

wdhawkins (approved):

resources/configure_dependencies.py
```diff
@@ -1,7 +1,14 @@
 """
 This is a utility script for the download and installation of OpenSn
 dependencies.
+
+NOTES for clang compiler:
+- CFLAGS, CXXFLAGS and FFLAGS passed into PETSc won't work. Unset them.
```
zhardy-lanl, line 6: In the PETSc build we explicitly set the CFLAGS, CXXFLAGS, and FFLAGS without a compiler check. Could you explain this a little more for me?

zhardy-lanl (approved):

resources/configure_dependencies.py
andrsd: OpenMP flag used by `gcc` is `-fopenmp`. `clang` uses `-fopenmp=libomp`. But, I think I was not super clear with this comment. When I used `clang` from `conda`, PETSc just would not configure. This is probably caused by conda exporting its own compiler flags into the environment, so the notes should be for "clang in conda environment".

As far as handling the PIC and OpenMP flags, PETSc has dedicated options for those (they are `--with-pic` and `--with-openmp`). I think this script should really use those instead. I think I want to add a patch for this change to this PR...

zhardy-lanl: Roger that. Sounds good to me

andrsd: I updated the script to use `--with-pic` (well, it was already in the flags), and added `--with-openmp`. I dropped `CFLAGS`, `CXXFLAGS` and `FCFLAGS`, so now they are set by PETSc (and work for all PETSc supported compilers). 

This opened another interesting problem: clang does not support `-march=native` which is specified in the OPTFLAGS, so I added some more notes on that.

But, it really seems that all this knowledge should be documented in the doco on pages dedicated to builds on individual platforms. I don't think we want to waste our time on maintaining this script and making sure it works everywhere...

ragusa: while we wait for the doco, I think we should go through the 4 install instructions we have: MacOS/Linux Easy/Standard.

I am in favor of supportting only the Standard versions for now. These do NOT rely on the configure-deps script at all. However, int he standard instructions, we can add a note that a canned version is possible through the scripts, without warranty. 

Is this ok for now? Can we do better? FYI, OpenMC does this: https://docs.openmc.org/en/stable/quickinstall.html

andrsd: I suggest to move this discussion into its own issue - it seems we need to make somewhat bigger changes.

ragusa:
@andrsd : have we opened a new issue for fixing/working on the install instructions?

andrsd:
I opened #87 for that

PR #89, opened 2024-01-24
removing some mesh functions, part 1

ragusa:
removing some mesh functions, part 1
* the volume extruder that is removed is not used by the [extruder_mesh_generator.cc](https://github.com/Open-Sn/opensn/blob/main/framework/mesh/mesh_generator/extruder_mesh_generator.cc). I believe this is the old extruder and it was not called during any regression test
* surface mesher was not called in any regression test
* volume mesher was not called in any regression test


andrsd (approved):
💣

zhardy-lanl (approved):

wdhawkins (approved):

PR #94, opened 2024-01-29
Removing alias typedef ending with `Ptr`

andrsd:
This removes custom defined types like `typedef std::shared_ptr<SomeType> SomeTypePtr`.

### DIFF
--- a/framework/mesh/unpartitioned_mesh/unpartitioned_mesh.cc
+++ b/framework/mesh/unpartitioned_mesh/unpartitioned_mesh.cc
@@ -604,7 +604,6 @@ UnpartitionedMesh::CopyUGridCellsAndPoints(vtkUnstructuredGrid& ugrid,
   const std::string fname = "UnpartitionedMesh::CopyUGridCellsAndPoints";
   typedef Vector3 Vec3;
   typedef Vec3* Vec3Ptr;
-  typedef LightWeightCell* CellPtr;
 
   const vtkIdType total_cell_count = ugrid.GetNumberOfCells();
   const vtkIdType total_point_count = ugrid.GetNumberOfPoints();
@@ -625,7 +624,7 @@ UnpartitionedMesh::CopyUGridCellsAndPoints(vtkUnstructuredGrid& ugrid,
 
   if (has_global_ids)
   {
-    std::vector<CellPtr> cells(total_cell_count, nullptr);
+    std::vector<LightWeightCell*> cells(total_cell_count, nullptr);
     std::vector<Vec3Ptr> vertices(total_point_count, nullptr);
 
     auto cell_gids_ptr = ugrid.GetCellData()->GetGlobalIds();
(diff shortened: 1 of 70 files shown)


framework/mesh/unpartitioned_mesh/unpartitioned_mesh.cc
```diff
@@ -625,8 +622,8 @@ UnpartitionedMesh::CopyUGridCellsAndPoints(vtkUnstructuredGrid& ugrid,
 
   if (has_global_ids)
   {
-    std::vector<CellPtr> cells(total_cell_count, nullptr);
-    std::vector<Vec3Ptr> vertices(total_point_count, nullptr);
+    std::vector<LightWeightCell*> cells(total_cell_count, nullptr);
+    std::vector<Vector3*> vertices(total_point_count, nullptr);
```
ragusa, line 626: @andrsd  are those raw pointers? what do you think of seeing a few of these int he code?

ragusa (approved):

framework/mesh/unpartitioned_mesh/unpartitioned_mesh.cc
andrsd: Yes, they are. I personally have nothing against  raw pointers. I learned coding before smart pointers were a thing in C++, so I am conditioned on memory allocation/deallocation. But, in more complicated codes it can easily happen that memory is not deallocated and memory leaks can occur. Then, you can have things like double free, etc.  

So, if you are thinking of solely using smart pointers (for consistency) and prevent memory related issues, then it needs to be a buy-in from all (or most) developers. And mixing raw and smart pointers should not be a thing, I think - developers will be always thinking about "was this a raw pointer or a smart pointer"...

ragusa: so, there should be no time penalty between raw and smart pointers? if so, should we remove all raw pointers?

zhardy-lanl: I would vote to remove raw pointers for objects

wdhawkins: I think we should eliminate raw pointers wherever we can. We're going to get more and more students working in the code that probably won't be very familiar with handling raw pointers. Eliminating them wherever possible will save us a lot of headaches in the future.

ragusa: so there is a consensus to remove raw pointers (or most of them). The only question that remains is whether this happens in this PR or if an issue gets created for a subsequent PR. Thanks!

zhardy-lanl (approved):

framework/mesh/unpartitioned_mesh/unpartitioned_mesh.cc
zhardy-lanl: As a practice, I think the scope of PRs should not be expanded unless absolutely necessary or the change is very small. If the only raw pointer in the code is the one here, I would change this PR, if there are more, I would create an issue and address in a new PR.

andrsd: I opened #95 for raw pointers elimination.

PR #96, opened 2024-01-30
Using `mpicpp-lite` for all MPI needs

### LINKED ISSUE #40: Thin wrapper for MPI
Per our discussion from that day: https://tamu.slack.com/archives/C0648L2TPT3/p1699972728830889 

Introduce a thin wrapper around all MPI calls. This is a quality-of-life improvement, so people do not have to do this, for example:
```
  auto BroadcastSingleInteger = [](int* int_being_bcast)
  { MPI_Bcast(int_being_bcast, 1, MPI_INT, 0, mpi.comm); };
  auto HomeBroadcastStringAsRaw = [](std::string string_to_bcast, int length)
  {
    char* raw_string_to_bcast = string_to_bcast.data();
    MPI_Bcast(raw_string_to_bcast, length, MPI_CHAR, 0, mpi.comm);
  };
  auto NonHomeBroadcastStringAsRaw = [](std::string& string_to_bcast, int length)
  {
    std::vector<char> raw_chars(length + 1, '\0');
    MPI_Bcast(raw_chars.data(), length, MPI_CHAR, 0, mpi.comm);

    string_to_bcast = std::string(raw_chars.data());
  };

    BroadcastSingleInteger(&console_input_len); // Non-Home locs wait here
    if (HOME)
      HomeBroadcastStringAsRaw(console_input, console_input_len);
    else
      NonHomeBroadcastStringAsRaw(console_input, console_input_len);

```
when they can do this:
```
std::string str;
comm.broadcast(str, 0);
```

Header-only library that can do this is [mpicpp-lite](https://github.com/andrsd/mpicpp-lite/).

andrsd:
This is adding `mpicpp-lite` - a quality of life layer for working with MPI.

Closes #40


framework/math/math_mpi.cc
ragusa, line 1: if this file used anywhere in the code?

framework/math/vector_ghost_communicator/vector_ghost_communicator.cc
```diff
@@ -71,7 +71,7 @@ VectorGhostCommunicator::MakeCachedParallelData()
   // MPI utility MapAllToAll in Chi-Tech accomplishes this task,
   // returning a mapping of processes to the global ids that this
   // process needs to send.
-  std::map<int, std::vector<int64_t>> send_map = MapAllToAll(recv_map, MPI_INT64_T, comm_);
```
ragusa, line 74: how is MPI_INT64_T incorporated in the new mpi wrappers?

framework/mesh/sweep_utilities/communicators/aah_asyn_comm.cc
```diff
@@ -280,34 +274,9 @@ AAH_ASynchronousCommunicator::ReceiveDelayedData(int angle_set_num)
...
-                     << "\n";
-          char error_string[BUFSIZ];
-          int length_of_error_string, error_class;
-          MPI_Error_class(error_code, &error_class);
-          MPI_Error_string(error_class, error_string, &length_of_error_string);
-          err_stream << error_string << "\n";
-          MPI_Error_string(error_code, error_string, &length_of_error_string);
-          err_stream << error_string << "\n";
-          log.LogAllWarning() << err_stream.str();
-        }
```
ragusa, line 310: how is the error handled now?

modules/linear_boltzmann_solvers/a_lbs_solver/acceleration/diffusion.cc
```diff
@@ -115,21 +115,21 @@ DiffusionSolver::Initialize()
 
   if (options.verbose) log.Log() << text_name_ << ": Global number of DOFs=" << num_global_dofs_;
 
-  opensn::mpi.Barrier();
+  opensn::mpi_comm.barrier();
   log.Log() << "Sparsity pattern";
-  opensn::mpi.Barrier();
+  opensn::mpi_comm.barrier();
```
ragusa, line 120: for later. @wdhawkins has mentioned several times that he believes we have too many barriers. This comes from a time when the old code had "training wheels". I believe we should make an issue to investigate whether we have excessive barriers. It's a to-do for after the cleanup phase.

ragusa (approved):
Great. Just a few comments from me.

framework/math/vector_ghost_communicator/vector_ghost_communicator.cc
andrsd: There is no direct mapping for `MPI_INT64_T`. But `long long int` should be that on a 64bit architecture. I can add the `intN_t` types, since MPI has mapping for them.

framework/mesh/sweep_utilities/communicators/aah_asyn_comm.cc
andrsd: This was an interesting observation. The default MPI behavior is that if an MPI call fails, MPI exits with an error and prints the error message, so you never see an error code other then MPI_SUCCESS unless you explicitly change the MPI behavior via [MPI_Comm_set_errhandler](https://www.mpich.org/static/docs/v1.5.x/www3/MPI_Comm_set_errhandler.html).

However, what the MPI wrapper does is that every MPI call is wrapped in a `check_mpi_error` call which check for an error and terminates the run, so application code does not have to worry about error handling any more. You just use the wrapper and you know that every MPI error is handled for you. This is a general concept, so if you use an MPI implementation that returns error codes, the wrapper will handle that for you.

Also, the wrapper has API for changing this behavior, i.e. returning error codes or terminating.

modules/linear_boltzmann_solvers/a_lbs_solver/acceleration/diffusion.cc
andrsd: Opened #97 for this.

framework/math/math_mpi.cc
andrsd: The recent code coverage shows no usage. But, IMO removal of dead code should happen in separate PRs.

framework/math/vector_ghost_communicator/vector_ghost_communicator.cc
andrsd: Never mind this. If I add those new types, compiler sees them as equivalent and reports a redefinition error. So this is captured correctly...

zhardy-lanl (changes requested):
A few quick questions before approval:

1. Is this something you wrote or a commonly used package that was copied in? Just curious.
2. Is there a rationale for the `impl` directory?
3. Should the file naming convention match the rest of the code?
4. Should the code be formatted like the rest of the code base?

wdhawkins:
Very happy to see this! It makes all of the mpi code that much better. IMHO, we should treat this as a third-party library like petsc or anything else and should not enforce our coding standards upon it.

wdhawkins (approved):

zhardy-lanl (approved):

PR #100, opened 2024-01-31
Adjoint Option

zhardy-lanl:
This PR introduces an adjoint option into the `LBSSolver`. This is primarily done as a first step towards separating response evaluations from the actual solver. The adjoint option was added to the `Options` struct, and to `LBSSolver::InputParameters` as well as routines for reorienting angles.

### Highlights

- Within `InitializeMaterials` a check is performed to determine if adjoint mode is on or off. This dictates whether `matid_to_xs_map` gets set with `SingleStateMGXS` or `AdjointMGXS`.
- When the `adjoint` option is specified via `SetOptions`, the routine explicitly looks it before going through other parameters because it is order sensitive. 
    - If the new flag is the same as the already set flag, nothing is done
    - If the flags are different and a discretization is not present, the solver has not been initialized and will be initialized in the appropriate form for an adjoint solve upon a call to the main `Initialize` routine.
    - If the flags are different and a discretization is present, the materials must be reinitialized, sources and boundary conditions cleared, and solutions zeroed out. This occurs within the `SetOptions` routine. The user is expected to pass the necessary source and boundary conditions with the adjoint option after initialization (as they are before). There is no good way to remove material sources, so one must be careful there.
- Tests updated to use the adjoint option.

### Notes

The previous adjoint solver pre-dated the executioner system. Its routine to reorient angles lived within its own execute routine. Since the executioner system, however, the `LBSSolver` have not implemented such a routine, which implies that the reorientation did not occur in the pre-existing adjoint solver. The tests passed because they were all scalar flux based, which does not get modified as a result of angle reorientation. The adjoint option implementation adds the reorientation routine to the base `LBSSolver`, where it does nothing, and then overrides it in the `DiscreteOrdinatesSolver` class. This allows for it to be called within the executioner system.

### DIFF
--- a/modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
+++ b/modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
@@ -18,6 +18,7 @@
 #include "framework/math/time_integrations/time_integration.h"
 #include "framework/math/spatial_discretization/finite_element/piecewise_linear/piecewise_linear_discontinuous.h"
 #include "framework/physics/physics_material/physics_material.h"
+#include "framework/physics/physics_material/multi_group_xs/adjoint_mgxs.h"
 #include "framework/physics/field_function/field_function_grid_based.h"
 #include <algorithm>
 #include <iomanip>
@@ -480,6 +481,8 @@ LBSSolver::OptionsBlock()
   "obtained elsewhere.");
   params.AddOptionalParameter("save_angular_flux", false,
   "Flag indicating whether angular fluxes are to be stored or not.");
+  params.AddOptionalParameter("adjoint", false,
+  "Flag for toggling whether the solver is in adjoint mode.");
   params.AddOptionalParameter("verbose_inner_iterations", true,
   "Flag to control verbosity of inner iterations.");
   params.AddOptionalParameter("verbose_outer_iterations", true,
@@ -578,6 +581,38 @@ LBSSolver::SetOptions(const InputParameters& params)
     if (user_params.GetParamValue<bool>("clear_point_sources")) point_sources_.clear();
   if (user_params.Has("clear_distributed_sources"))
     if (user_params.GetParamValue<bool>("clear_distributed_sources")) distributed_sources_.clear();
+  if (user_params.Has("adjoint"))
+  {
+    const bool adjoint = user_params.GetParamValue<bool>("adjoint");
+    if (adjoint != options_.adjoint)
+    {
+      options_.adjoint = adjoint;
+
+      // If a discretization exists, the solver has already been initialized.
+      // Reinitialize the materials to obtain the appropriate xs and clear the
+      // sources to prepare for defining the adjoint problem
+      if (discretization_)
+      {
+        // The materials are reinitialized here to ensure that the proper cross sections
+        // are available to the solver. Because an adjoint solve requires volumetric or
+        // point sources, the material-based sources are not set within the initialize routine.
+        InitializeMaterials();
+
+        // Forward and adjoint sources are fundamentally different, so any existing sources
+        // should be cleared and reset through options upon changing modes.
+        point_sources_.clear();
+        distributed_sources_.clear();
+        boundary_preferences_.clear();
+
+        // Set all solutions to zero.
+        phi_old_local_.assign(phi_old_local_.size(), 0.0);
+        phi_new_local_.assign(phi_new_local_.size(), 0.0);
+        for (auto& psi : psi_new_local_)
+          psi.assign(psi.size(), 0.0);
+        precursor_new_local_.assign(precursor_new_local_.size(), 0.0);
+      }
+    }
+  }
 
   // Handle order insensitive options
   for (size_t p = 0; p < user_params.NumParameters(); ++p)
@@ -748,7 +783,7 @@ LBSSolver::Initialize()
 
   mpi_comm.barrier();
 
-  InitMaterials();                   // c
+  InitializeMaterials();             // c
   InitializeSpatialDiscretization(); // d
   InitializeGroupsets();             // e
   ComputeNumberOfMoments();          // f
@@ -856,14 +891,15 @@ LBSSolver::PrintSimHeader()
 }
 
 void
-LBSSolver::InitMaterials()
+LBSSolver::InitializeMaterials()
 {
-  const std::string fname = "lbs::SteadyStateSolver::InitMaterials";
   log.Log0Verbose1() << "Initializing Materials";
 
   // Create set of material ids locally relevant
-  std::set<int> unique_material_ids;
+  const size_t num_physics_mats = material_stack.size();
+
   int invalid_mat_cell_count = 0;
+  std::set<int> unique_material_ids;
   for (auto& cell : grid_ptr_->local_cells)
   {
     unique_material_ids.insert(cell.material_id_);
@@ -877,10 +913,10 @@ LBSSolver::InitMaterials()
     if (cell.material_id_ < 0) ++invalid_mat_cell_count;
   }
 
-  if (invalid_mat_cell_count > 0)
-  {
-    log.LogAllWarning() << "Number of invalid material cells: " << invalid_mat_cell_count;
-  }
+  log.Log() << "Invalid cell materials " << invalid_mat_cell_count;
+  ChiLogicalErrorIf(invalid_mat_cell_count > 0,
+                    std::to_string(invalid_mat_cell_count) +
+                      " cells encountered with an invalid material id.");
 
   // Get ready for processing
   std::stringstream materials_list;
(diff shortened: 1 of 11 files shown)


modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -480,7 +478,110 @@ DiscreteOrdinatesSolver::SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::v
...
+      {
+        // Reorient flux moments
+        //
+        // Because flux moments are integrated angular fluxes, the
+        // angular flux and spherical harmonics must be evaluated at
+        // opposite angles in the quadrature integration. Taking advantage
+        // of the even/odd nature of the spherical harmonics, i.e.
+        // Y_{\ell,m}(-\Omega) = (-1)^\ell Y_{\ell,m}(\Omega), the flux
+        // moments must be multiplied by (-1)^\ell.
+        for (int m = 0; m < num_moments_; ++m)
```
ragusa, line 556: In angular_quadrature_base.h, the HarmonicIndices struct has two entries: ell and m, following the standard math notation for spherical harmonics. 
Here, m is used has the running index for the mapping between the 2 indices {ell,m}(math) and the one index m (programming). Maybe m(programming) should be imom or n or something.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.cc
```diff
@@ -578,6 +581,38 @@ LBSSolver::SetOptions(const InputParameters& params)
...
+      options_.adjoint = adjoint;
+
+      // If a discretization exists, the solver has already been initialized.
+      // Reinitialize the materials to obtain the appropriate xs and clear the
+      // sources to prepare for defining the adjoint problem
+      if (discretization_)
+      {
+        // The materials are reinitialized here to ensure that the proper cross sections
+        // are available to the solver. Because an adjoint solve requires volumetric or
+        // point sources, the material-based sources are not set within the initialize routine.
```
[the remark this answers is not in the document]
zhardy-lanl: Sources attached to materials. The only one that exists is `IsotropicMultiGrpSource`. What this comment is saying is that we do not go into `Material` objects, look for properties, and remove them for you.

modules/linear_boltzmann_solvers/a_lbs_solver/lbs_solver.h
```diff
@@ -457,23 +457,30 @@ class LBSSolver : public opensn::Solver
...
+  virtual void SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::vector<int>& groupset_ids,
+                                                          Vec x,
                                                           PhiSTLOption which_phi);
 
+  /**
+   * A method for post-processing an adjoint solution.
+   *
+   * @note This does nothing for diffusion-based solvers.
+   */
+  virtual void ReorientAdjointSolution(){};
```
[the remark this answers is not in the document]
zhardy-lanl: There is a doc-string for the overridden method which details what happens. In documentation, if you look at `DiscreteOrdinatesdSolver::ReorientAdjointSolution` what you are saying would be there. I did not put it here since this routine in `LBSSolver` does not do that. I can still add it here if you'd like.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -480,7 +478,110 @@ DiscreteOrdinatesSolver::SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::v
...
+
+      // Go through angles until all are paired
+      std::set<size_t> visited;
+      for (int m = 0; m < num_gs_angles; ++m)
+      {
+        // Skip if already encountered
+        if (visited.count(m) > 0) continue;
+
+        bool found = true;
+        for (int n = 0; n < num_gs_angles; ++n)
```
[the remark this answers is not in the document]
zhardy-lanl: This is a good idea. I will make the change.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
zhardy-lanl: Fair enough. I will change this.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -480,7 +478,110 @@ DiscreteOrdinatesSolver::SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::v
...
+    const auto& uk_man = groupset.psi_uk_man_;
+
+    // Build reversed angle mapping
+    std::map<int, int> reversed_angle_map;
+    if (options_.save_angular_flux)
+    {
+      const auto& omegas = groupset.quadrature_->omegas_;
+      const auto num_gs_angles = omegas.size();
+
+      // Go through angles until all are paired
```
ragusa, line 505: angles --> directions

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -480,7 +478,110 @@ DiscreteOrdinatesSolver::SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::v
...
+    // Angular flux info
+    auto& psi = psi_new_local_[gs];
+    const auto& uk_man = groupset.psi_uk_man_;
+
+    // Build reversed angle mapping
+    std::map<int, int> reversed_angle_map;
+    if (options_.save_angular_flux)
+    {
+      const auto& omegas = groupset.quadrature_->omegas_;
+      const auto num_gs_angles = omegas.size();
```
ragusa, line 503: num_gs_angles --> num_gs_dirs or num_gs_directions

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -480,7 +478,110 @@ DiscreteOrdinatesSolver::SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::v
...
+      std::set<size_t> visited;
+      for (int idir = 0; idir < num_gs_angles; ++idir)
+      {
+        // Skip if already encountered
+        if (visited.count(idir) > 0) continue;
+
+        bool found = true;
+        for (int jdir = 0; jdir < num_gs_angles; ++jdir)
+        {
+          // Angles are opposite if their sum is zero
```
ragusa, line 515: Angles --> Directions

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -480,7 +478,110 @@ DiscreteOrdinatesSolver::SetPrimarySTLvectorFromMultiGSPETScVecFrom(const std::v
...
+
+    // Moment map for flux moments
+    const auto& moment_map = groupset.quadrature_->GetMomentToHarmonicsIndexMap();
+
+    // Angular flux info
+    auto& psi = psi_new_local_[gs];
+    const auto& uk_man = groupset.psi_uk_man_;
+
+    // Build reversed angle mapping
+    std::map<int, int> reversed_angle_map;
```
ragusa, line 499: reversed_angle_map --> reversed_direction_map

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
zhardy-lanl: This is a less trivial change. Most of the code base calls directions angles, not the other way around... See `modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.cc` for example. Thoughts?

ragusa: This comment was specific to the commented line:

// Go through angles until all are paired
-->
// Go through directions until all are paired

ragusa: if you are referring to angle_sets, I am willing to discuss this elsewhere. 
Here, we are really talking about psi^*(omega) = psi(-omega) and I think we would more likely say [gerund verb] the opposite direction rather than [gerund] the opposite angle.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
ragusa: num_gs_angles is a local var.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
ragusa: I refer to the comment

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
ragusa: local var

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
zhardy-lanl: I am referring to the 24 instances of "angle" throughout (the sweeper loops over angles, not directions) versus the 7 instances of "direction" within as many lines of code. Using "direction" is the exception, not the rule in the code base, which is why I am hesitant to break from the existing convention. And yes, I am aware that these are all local variables. 

I tend to agree with you that maybe direction is better than angle, but I don't think we should start deciding to break from existing convention in small parts of the code via PRs separate topics. Let's create an issue and discuss further.

ragusa (approved):

andrsd (approved):

wdhawkins (approved):

PR #109, opened 2024-02-06
Renaming ChiXS to OpenSn XS

andrsd:
This change set does:
- Renaming ChiXS -> OpenSnXS
- Changing file extension used for cross section files from `.cxs` to `.xs`


framework/physics/physics_material/multi_group_xs/single_state_mgxs.h
```diff
@@ -55,9 +55,9 @@ class SingleStateMGXS : public MultiGroupXS
 
 public:
   /**
-   * This method populates a transport cross section from a Chi cross section file.
+   * This method populates a transport cross section from a OpenSn cross section file.
    */
-  void MakeFromChiXSFile(const std::string& file_name);
+  void MakeFromOpenSnXSFile(const std::string& file_name);
```
ragusa, line 60: my only comment would be: do we want to keep this name (MakeFromOpenSnXSFile)? @wdhawkins @zachhardy @zhardy-lanl. Specifically, I am just not sure about the action verb "Make". 
Otherwise (and it's not that important), it's good to go.

ragusa (approved):

framework/physics/physics_material/multi_group_xs/single_state_mgxs.h
zhardy-lanl: I have a few thoughts on this... But I like `Create` over `Make` for all of the `Make...` routines. This is a little out of scope, so opened #111 to further discuss XS naming/structure.

zhardy-lanl (approved):

wdhawkins (approved):

PR #110, opened 2024-02-06
Removing references to ChiTech

andrsd:
Removing some left-over references to ChiTech


framework/mesh/surface_mesh/surface_mesh.cc
```diff
@@ -1076,7 +1076,7 @@ SurfaceMesh::ExportToOBJFile(const char* fileName)
   }
 
   fprintf(outputFile, "# Exported mesh file from tringulation script\n");
```
ragusa, line 1078: --> triangulation

ragusa (approved):
1 approval is enough. one typo to fix and then good to go.

zhardy-lanl (approved):
Ditto on the typo fix.

Issue #30, opened 2023-11-09
Physics directory reorganization (or removal)


zhardy-lanl:
Currently, the `physics` directory houses field functions, materials, solvers, time steppers, and an event publisher. I think these should be reorganized as follows:

- `physics/physics_event_publisher` -> `event_system` directory
- `physics/field_function` -> `field_function` (see #29)
- `physics/physics_material` -> `material`
- `physics/solver_base` -> `solvers/solver.h(cc)`
- `physics/time_steppers` -> `solvers/time_steppers`

andrsd:
There is something called `math_time_stepping.{cc|h}`, should time steppers go into `math` or those 2 files should go into `solver/time_steppers` (I assume `solvers` is a new directory at the same level as `physics`)?

zhardy-lanl:
Oh wow, a deeper look and this is more of a can of worms. There is `physics/SolverBase` which is meant to be a base class for modules, then there is `math/LinearSolvers` and `math/NonlinearSolvers` which do not have a connection to the aforementioned solver. There is also a `physics/TimeSteppers` and a `math/TimeIntegrations`. There is probably going to need to be a deeper discussion on this. 

My preliminary thoughts were to have `framework/solvers` be a place to define the platforms from which modules can be derived. You would have the generic `Solver`, which is generic, then a generic `LinearSolver` and `NonlinearSolver (which could have a PETSc, Trillions, etc. derived class) that inherits from `Solver`. In this way, a module inheriting from `Solver` has ultimate flexibility, and a module inheriting from say `LinearPETScSolver` has the basic PETSc solver structure available. This is obviously future work, so for the meantime, I would suggest:

- `framework/(math/)solvers`
    - `solver.{h|cc}`
    - `linear(_solvers)` (from `math/LinearSolver`)
    - `nonlinear(_solvers)` (from `math/NonLinearSolver`)

The `_solvers` I am on the fence about for redundancy, and there is an argument to be made for `solvers` staying in `math`.

As for the time steppers, those can go under `math` in a single directory. If `spatial_discretization` is under `math`, time discretization should probably be there too. I don't think we have any time-dependent tests, so there is a lot of flexibility in how we ultimately handle that part of the code. I think a lot of that was put in very recently.

PR #112, opened 2024-02-06
Physics directory reorganization, pt1

andrsd:
This PR works towards #30. It:

1. Moves `physics_event_publisher` into `framework/event_system`.
2. Renames `MaterialProperty` to `PhysicsMaterialProperty`. This is just an interim step to avoid class name collision with another class named `MaterialProperty`. More work in unifying the material system is expected. I added the prefix `Physics` to denote that this came from the physics directory.

Refs #30


ragusa (approved):
Just a reminder that there is a newer Material class in framework/materials that uses the GetInputParameters() but the transition from physics material to material was never done. When we finally settle in how to work with the input parameters, we should revisit materials one more time.

zhardy-lanl (commented):
Forgive me if I am not remembering a discussion, but are we sure we want to go with `PhysicsMaterial(Property)`? To me all materials are physics materials, since this is  physics code. The way I was hoping to move things, as I stated in #30 was to have a `material` directory and to move everything from `physics/material` into it, including cross sections. What is everyone's thoughts on this?

andrsd:
From the PR description:

> Renames MaterialProperty to PhysicsMaterialProperty. This is just an interim step to avoid class name collision with another class named MaterialProperty. More work in unifying the material system is expected. I added the prefix Physics to denote that this came from the physics directory.

ragusa:
As David said, this is a partial step, not a full step. Is this doubly-defined material (the old physics material and the new material that is still empty) the situation that will force us to address the GetInputParameter issue sooner rather than later @zhardy-lanl @andrsd  ?

zhardy-lanl:
> From the PR description:
> 
> > Renames MaterialProperty to PhysicsMaterialProperty. This is just an interim step to avoid class name collision with another class named MaterialProperty. More work in unifying the material system is expected. I added the prefix Physics to denote that this came from the physics directory.

Well, I certainly should have taken an extra second to read that before diving in... Sorry about that. As long as this is not the final step, fine for now.

andrsd:
> Is this doubly-defined material (the old physics material and the new material that is still empty) the situation that will force us to address the GetInputParameter issue sooner rather than later?

Issue #30 requests directory reorg. This accomplishes that requirement. Re-designing/refactoring the material system would be THE step to fix that part of the code. But, that will need a design meeting IMO. After that we can open an issue to do the work on the material system which is have all the details from the design step...

I am not sure that I want to tackle the full task of fixing the input parameter issue(s). Maybe working on these smaller systems will have the side effect of moving the input params where they need to be...

andrsd:
> Well, I certainly should have taken an extra second to read that before diving in... Sorry about that.

No problem, that happens...

Issue #17, opened 2023-11-08
mesh refactoring


ragusa:
Let's address mesh refactoring (for detailled discussions, please go to the appropriate slack channel).
- [x] `VolumeMesher` has to go but some lua functions rely on methods from `VolumeMesher`: `chiVolumeMesherSetProperty`, `chiVolumeMesherSetupOrthogonalBoundaries`, `chiVolumeMesherSetMatIDToAll`. The methods we need to keep should go to `MeshModifiers`
- [x] `UnpartitionedMesh`: keep but some lua functions are unknown to me and are not called: `chiUnpartitionedMeshUploadVertex`, `chiUnpartitionedMeshUploadCell`, `chiUnpartitionedMeshFinalizeEmpty` so I think these can go
- [x] `SurfaceMesher`: I couldn't see any test that used any of the 3 lua functions from this so it can go.
- [x] `mesh_cutting `should go (we determined that cutting for KBA is no longer something we want to do)
- [x] `domain_decomposition `should go (looks old and unused)
- [x] I do not think `SurfaceMesh `is of any use. IT is, of course, referenced in many places. Unless it is useful for setting boundaryID using LV (if boundaryID cannot be set with LV), I'd say remove. I'd like for @zachhardy to confirm

zhardy-lanl:
@ragusa I have never used `SurfaceMesh`. No issue removing it on my end.

andrsd:
I identified what `UnpartitionedMeshUploadVertex`, `UnpartitionedMeshUploadCell`, and `UnpartitionedMeshFinalizeEmpty` do. They allow users to create meshes directly from the lua file, i.e. `UnpartitionedMeshUploadVertex` is used to create vertices and `UnpartitionedMeshUploadCell` is used to build cells.

So question. Do you want to keep this functionality or just remove it? They are not being tested by anything.

ragusa:
We want to limit users from creating meshes from opensn directly. I think they need to either be satisfied with ortho grids and our extrusion options. Everything else should be done in real mesh generators. So, I am fine killing the above functionality.

ragusa:
`SurfaceMesh` is used when create a logical volume with a skin mesh. so we are keeping `SurfaceMesh`. See PR #135

ragusa:
SurfaceMesher was deleted in PR #89

andrsd:
[...]

PR #118, opened 2024-02-12
Remove VolumeMesher

andrsd:
- Remove `VolumeMesher`
- No more `MeshHandler`, `MeshContinuum` instances are stored in `std::vector<std::shared_ptr<MeshContinuum>> mesh_stack`
- VolumeMesherExportXYZ is now `mesh.ExportToXYZ`
- `GetCurrentHandler().GetGrid()` is now `GetCurrentMesh()`
- `VolumeMesher::Set{Boundary|Material}ID` moved into `MeshContinuum`
- Removed lots of unused/obsolete code.

Refs #17


framework/mesh/mesh.cc
```diff
@@ -1,40 +1,22 @@
...
-
-size_t
-PushNewHandlerAndGetIndex()
-{
-  meshhandler_stack.push_back(std::make_shared<MeshHandler>());
-
-  int index = (int)meshhandler_stack.size() - 1;
-  current_mesh_handler = index;
-
-  return index;
```
ragusa, line 27: what are the implications of this? only one mesh can be "live" at a time during execution? @zhardy-lanl for opinion. This looks simpler but I just want to be sure.

ragusa (commented):
I agree with your comment "they are more like SetMatIDToAll call. So should those become mesh.SetMaterialIDFromLogicalVolume, etc". Yes for me. Adding @zhardy-lanl to confirm.

zhardy-lanl:
> I agree with your comment "they are more like SetMatIDToAll call. So should those become mesh.SetMaterialIDFromLogicalVolume, etc". Yes for me. Adding @zhardy-lanl to confirm.

Agreed on this. One thing I will say is that `SetMatIDToAll` is a bit strange of a name. `SetUniformMaterialID` would be more appropriate in my view and go along with `SetMaterialIDFrom...` routines better.

framework/mesh/mesh.cc
andrsd: There is this now: `std::vector<std::shared_ptr<MeshContinuum>> mesh_stack`. This serves as a stack, so the lua layer should be able to grab any mesh from the stack and use it. But, right now, as @zhardy-lanl said, there is only one element in the vector.

andrsd:
I don't like `SetMatIDToAll`. I like `SetUniformMaterialID` so I will use that. One more thing. Right now, these calls assume to grab the top most mesh from the mesh_stack. Should they also take the mesh object somehow? Something like: `mesh.SetUniformMaterialID(m, material_id)`, where `m` would be the `MeshContinuum` instance that the user would grab via some API?

ragusa:
that would make sense to me

zhardy-lanl:
> I don't like `SetMatIDToAll`. I like `SetUniformMaterialID` so I will use that.
> 
> One more thing. Right now, these calls assume to grab the top most mesh from the mesh_stack. Should they also take the mesh object somehow? Something like: `mesh.SetUniformMaterialID(m, material_id)`, where `m` would be the `MeshContinuum` instance that the user would grab via some API?

What you propose here would make more sense to me than just grabbing the mesh at the top of the stack, however, it kind of doesn't jive with the change to `GetCurrentMesh` proposed above. I would make sure both are consistent (i.e. grabs the top, or uses some sort of handle).

andrsd:
`GetCurrentMesh` is a temporary band aid until I can pass the `MeshContinuum` into the `Solver` c-tor. Then, it would go away, and the lua layer would be grabbing objects from the stacks as needed.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -892,24 +890,26 @@ DiscreteOrdinatesSolver::InitializeSweepDataStructures()
...
+  //
+  //      bool is_1D_geometry = options_.geometry_type == GeometryType::ONED_SLAB;
+  //
+  //      if (no_cycles_parmetis_partitioning and not is_1D_geometry and (opensn::mpi_comm.size() >
+  //      1))
+  //        throw std::logic_error("When using PARMETIS type partitioning then groupset iterative "
+  //                               "method must be NPT_CLASSICRICHARDSON_CYCLES or
+  //                               NPT_GMRES_CYCLES");
+  //    } // for groupset
+  //  }
```
andrsd, line 912: I wanted to ask about this. Is this still relevant? I commented this out and it does not break anything, but wanted to double check with you...

andrsd:
Rebased on the latest `main`, so this should be easier to read now...

doc/PAGES/MeshTutorials/MeshTutorial_07_3DExtrudedMesh.h
```diff
@@ -56,7 +56,7 @@ or
  SurfaceMesherExecute();
  VolumeMesherExecute();
 
- MeshHandlerExportMeshToVTK("export_mesh_without_IDs")
+ mesh.ExportToVTK("export_mesh_without_IDs")
```
ragusa, line 59: to help the user, we should start doing this (module.action)

doc/PAGES/MeshTutorials/MeshTutorial_07_3DExtrudedMesh.h
```diff
@@ -78,7 +78,7 @@ cylinder),
 -- Logical Volumes
 my_LV = LogicalVolumeCreate(RCC, 0, 0, 0.1, 0, 0, 0.2, 0.4)
```
ragusa, line 79: and here's an exmaple where we do not have module.function/action

doc/PAGES/ProgrammersManual/CodingTutorials/CodeTut1.h
```diff
@@ -49,7 +49,7 @@ MeshCreateUnpartitioned2DOrthoMesh(mesh,mesh)
 VolumeMesherExecute();
 
 --############################################### Set Material IDs
-VolumeMesherSetMatIDToAll(0)
+mesh.SetMatIDToAll(0)
```
ragusa, line 52: didn't Zach have another name for this one? with uniform or something like that?

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -1132,4 +1131,76 @@ MeshContinuum::GetGlobalNumberOfCells() const
   return num_cells;
 }
 
+void
+MeshContinuum::SetMatIDToAll(int mat_id)
```
ragusa, line 1135: check with @zhardy-lanl about the final name for this one

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
ragusa: @zhardy-lanl @wdhawkins , remind me what the NPT_ prefix is all about. Is this all/deprecated? and also, reply to David's query.

ragusa (commented):
Will approve after @zhardy-lanl chimes in for the name and after @zhardy-lanl + @wdhawkins chime in for the commented out test about cycles.

doc/PAGES/ProgrammersManual/CodingTutorials/CodeTut1.h
andrsd: I did not make that change, yet.

framework/mesh/mesh_continuum/mesh_continuum.cc
andrsd: I like the `SetUniformMaterialID` better that this and I will add a patch for that.

doc/PAGES/MeshTutorials/MeshTutorial_07_3DExtrudedMesh.h
andrsd: Personally, I would not focus too much on the header files in `doc`. I expect those to go away. I usually modify those for 2 reasons: (1) in case of salvaging the content, it will be closer to the current state, (2) when I do code-wide search, it does not show up any more, so less filtering on my side...

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
zhardy-lanl: Ah, that is VERY old! Stands for **N**eutral **P**article **T**ransport. The prefixes can be removed (or the options eliminated) if unused.

andrsd: So, just to confirm, this is safe to delete. Plus, I can add a patch to remove the options, etc.

doc/PAGES/MeshTutorials/MeshTutorial_07_3DExtrudedMesh.h
andrsd: True. I am assuming we will be removing the old syntax piece by piece as we move towards the new one?

andrsd:
I started to work on refactoring `mesh.SetProperty(...` and it is going to be a bigger thing so I will open an issue for it and do that in an individual PR, so this does not get big and doing multiple things.

ragusa (approved):

zhardy-lanl (approved):

wdhawkins (approved):

PR #130, opened 2024-02-15
Refactoring AAH and CBC sweep chunks

### LINKED ISSUE #120: Re-factor sweep chunks
The sweep chunks are in need of a massive re-factoring. It looks to me like there is quite a bit of simplification that can be done.

wdhawkins:
This is a first pass at refactoring the CBC and AAH sweep chunks. There's more to do, but this is a good start. No effort was made to improve performance. This was mostly about code quality, readability, and maintenance.

This is a hefty PR. Sorry about that.

Closes #120


modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.cc
```diff
@@ -39,277 +37,242 @@ AAH_SweepChunk::Sweep(AngleSet& angle_set)
...
+    const auto& sigma_t = xs_.at(cell.material_id_)->SigmaTotal();
 
-    // Get Cell matrices
-    const auto& fe_intgrl_values = unit_cell_matrices_[cell_local_id_];
-    G_ = &fe_intgrl_values.intV_shapeI_gradshapeJ;
-    M_ = &fe_intgrl_values.intV_shapeI_shapeJ;
-    M_surf_ = &fe_intgrl_values.intS_shapeI_shapeJ;
-    IntS_shapeI_ = &fe_intgrl_values.intS_shapeI;
+    // Get cell matrices
+    const auto& G = unit_cell_matrices_[cell_local_id].intV_shapeI_gradshapeJ;
```
ragusa, line 78: that must be the PDT way, with 2 integration by parts, where only 1 applies upwinding. otherwise, I would have expected the name to be intV_shapeJ_gradshapeI. So, I suppose the surface mass matrix M_surf ONLY applied to the incoming faces.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.cc
```diff
@@ -39,277 +37,242 @@ AAH_SweepChunk::Sweep(AngleSet& angle_set)
...
       {
-        g_ = gs_gi_ + gsg;
-        gsg_ = gsg;
-        sigma_tg_ = sigma_t[g_];
-
-        const auto& M = *M_;
-        const auto& m2d_op = groupset_.quadrature_->GetMomentToDiscreteOperator();
+        double sigma_tg = sigma_t[gs_gi + gsg];
 
         // Contribute source moments q = M_n^T * q_moms
```
ragusa, line 165: what is that transpose? I bet m2d and d2m are built with the same meaning for rows and cols while, in reality they should not be that way. If I am right, this can be really confusing for someone reading the code after reading a paper. Let's remember this when you tackle the angular qaudrature.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.cc
```diff
@@ -39,277 +37,242 @@ AAH_SweepChunk::Sweep(AngleSet& angle_set)
...
-          } // for m
-          source_[i] = temp_src;
-        } // for i
-
-        // Mass Matrix and Source
-        // Atemp  = Amat + sigma_tgr * M
-        // b     += M * q
-        for (int i = 0; i < cell_num_nodes_; ++i)
+            const size_t ir = cell_transport_view.MapDOF(i, m, static_cast<int>(gs_gi + gsg));
+            temp_src += m2d_op[m][direction_num] * source_moments_[ir];
```
ragusa, line 172: yep. in m2d, the col index should be moments and the row index should be dir.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.cc
```diff
@@ -39,277 +37,242 @@ AAH_SweepChunk::Sweep(AngleSet& angle_set)
...
-      // Flux updates
-      const auto& d2m_op = groupset_.quadrature_->GetDiscreteToMomentOperator();
+      // Update phi
       auto& output_phi = GetDestinationPhi();
-
       for (int m = 0; m < num_moments_; ++m)
       {
-        const double wn_d2m = d2m_op[m][direction_num_];
-        for (int i = 0; i < cell_num_nodes_; ++i)
+        const double wn_d2m = d2m_op[m][direction_num];
```
ragusa, line 200: here, the meaning of the indices is correct. as you can see, it's identical to m2d while the meaning should be swapped for m2d, unless there is an efficiency reason.

ragusa (approved):
we need more commits like this one!

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.h
```diff
@@ -10,31 +9,13 @@ namespace opensn
...
-  double* GetDownwindPsi(int face_node_local_idx) const override;
-};
-
-/**The new sweep chunk class.*/
 class AAH_SweepChunk : public SweepChunk
 {
 public:
   AAH_SweepChunk(const MeshContinuum& grid,
-                 const SpatialDiscretization& discretization,
+                 const SpatialDiscretization& grid_fe_view,
```
andrsd, line 16: Is there a reason to call `SpatialDiscretization` `grid_fe_view`?

andrsd (approved):

andrsd:
A general comment on chunk naming. There are things like `CBC_FLUDS`. I think, the underscore is there to split the 2 acronyms, but that does not conform to our naming conventions. May be, it should be called `CbcFluds`. There is also, `CBC_SweepChunk`, that may become `CbcSweepChunk`. Similarly for the other sweepers. But, I think the naming change can be done later. Just bringing this up, since you (@wdhawkins) are touching the code...

wdhawkins:
> we need more commits like this one!

I was very happy to have this all cleaned up. I have a few more cleaning commits coming along shortly, plus quadrature work.

wdhawkins:
> A general comment on chunk naming. There are things like `CBC_FLUDS`. I think, the underscore is there to split the 2 acronyms, but that does not conform to our naming conventions. May be, it should be called `CbcFluds`. There is also, `CBC_SweepChunk`, that may become `CbcSweepChunk`. Similarly for the other sweepers. But, I think the naming change can be done later. Just bringing this up, since you (@wdhawkins) are touching the code...

Let me make a first pass and bringing the names in line with our convention. It should be a relatively trivial change.

wdhawkins:
> A general comment on chunk naming. There are things like `CBC_FLUDS`. I think, the underscore is there to split the 2 acronyms, but that does not conform to our naming conventions. May be, it should be called `CbcFluds`. There is also, `CBC_SweepChunk`, that may become `CbcSweepChunk`. Similarly for the other sweepers. But, I think the naming change can be done later. Just bringing this up, since you (@wdhawkins) are touching the code...

I've fixed the sweep chunk class names to comply with the coding style and propagated them upstream. For things like `CBC_FLUDS`, I'll fix them as I refactor those classes. I will eventually rewrite, refactor, and extend all of the sweep code.

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/aah_sweep_chunk.cc
ragusa: @wdhawkins , quick question. you marked this as resolved. does this mean that in an upcoming update (on quadratures), the m2d array will be numbered differently and at that momenet, you wil fix also the comment that has ^T ?

wdhawkins: Yes, that's correct.  I'll address the issue with m2d after I pull it in to the quadrature class and change how it's built. I'll create an issue as a reminder.

zhardy-lanl:
Just for posterity since I was too late getting to this... On naming conventions, I personally feel like acronyms should stay capitalized in the spirit of CamelCase. To me `Aah` looks much more like ah than `AAH`. We discussed wanting to change the name, so I am assuming further changes will be coming. But we should discuss naming conventions with acronyms further. If I missed a discussion on this somewhere I apologize.

PR #131, opened 2024-02-15
Fixing lv surface

ragusa:
making sure a logical volume can be created with a skin mesh (a skin mesh is a 3D surface mesh). 
- due to confusion between old and new way of interacting at the lua level, the surface mesh stack (which didn't work) was replaced by the object stack in the old way. in a few weeks, this will be completely removed so I am fine leaving it as is
- added two text cases in the testing folder and update to the json gold file.


andrsd (changes requested):
Code is fine. 

For commit history:
1) remove the merge commit. please stop doing that. that does not belong in the work flow that is developing new stuff in feature branches like we do.
2) squash patch "removing extra cout" into "fixing lv volume using surface mesh with David help". 
   - use `git rebase --interactive HEAD~5`. 
   - in the editor move the line that says "removing extra cout" right after the line that says "fixing lv volume using surface mesh with David help". 
   - change `pickup` to `fixup`
   - save and exit the file in the editor (if using `vim`: `ESC` `x` `Enter`)
   - then force push the branch
   - NOTE: you this same procedure to fix the formatting

andrsd (approved):
Good :+1:

PR #119, opened 2024-02-12
Removing kernel/callback system from LBS sweep chunks

wdhawkins:
Fixes #116

PR #133, opened 2024-02-16
Export Fix

rodriguez-jay:
Added mpi barriers to the ends of mesh and physics exports to resolve issue with visualizing extruded geometries in visit and paraview. Issue #132.

PR #134, opened 2024-02-17
Bug fixes for PWLRZ sweepchunk. Minor renaming across sweep chunks.

wdhawkins:
This PR fixes PWLRZ bugs introduced by PR #119. The PWLRZ sweepchunk was based on the AAH sweepchunk and used custom kernels to introduce RZ-specific code. Unfortunately, the regression tests for RZ produce the same answers for both RZ and XY, so the bugs weren't discovered until after #119 was committed. I have re-factored the PWLRZ sweepchunk to be self-contained like the AAH and CBC sweepchunks. This results in some code duplication, but the end result is easier to read, understand, and maintain.

In addition, there is minor variable renaming across sweepchunks to be consistent with the rest of the code.


ragusa (approved):
it was clear from your comment whether this actually fixes the rz solve, i.e., is the rz solve giving now different answers then the xy? if so, the json file was not updated. or is this more comestic/standardization, and the issue whether the rz solve works is still a question for later (in addition to the triangular grid issue mentioned by Zach)

wdhawkins:
> it was clear from your comment whether this actually fixes the rz solve, i.e., is the rz solve giving now different answers then the xy? if so, the json file was not updated. or is this more comestic/standardization, and the issue whether the rz solve works is still a question for later (in addition to the triangular grid issue mentioned by Zach)

Yes, I believe this does fix the rz solve. I tested the various pieces as I was implementing this, and all of the rz pieces need to be in place for the regression tests to pass. I believe the current  regression tests were designed to give the same answer in rz and xy for comparison purposes. We absolutely need additional rz regression tests. It’s unknown whether this rewrite resolves the issue with triangular rz meshes. I need an input to test with.

modules/linear_boltzmann_solvers/c_discrete_ordinates_curvilinear_solver/lbs_curvilinear_solver.cc
```diff
@@ -384,7 +384,7 @@ DiscreteOrdinatesCurvilinearSolver::ComputeSecondaryUnitIntegrals()
 std::shared_ptr<SweepChunk>
 DiscreteOrdinatesCurvilinearSolver::SetSweepChunk(lbs::LBSGroupset& groupset)
 {
-  auto sweep_chunk = std::make_shared<SweepChunkPWLRZ>(*grid_ptr_,
+  auto sweep_chunk = std::make_shared<SweepChunkPwlrz>(*grid_ptr_,
```
zhardy-lanl, line 387: What is the rationale for lower case `Pwlrz`? These are essentially distinct words, so in my view they should stay capitalized. Personally, I think `SweepChunkRZ` would be more appropriate. Since the discretization is not in the name for Cartesian sweep chunks, it doesn't feel like it should be for cylindrical coordinates either.

zhardy-lanl:
> > it was clear from your comment whether this actually fixes the rz solve, i.e., is the rz solve giving now different answers then the xy? if so, the json file was not updated. or is this more comestic/standardization, and the issue whether the rz solve works is still a question for later (in addition to the triangular grid issue mentioned by Zach)
> 
> Yes, I believe this does fix the rz solver. I tested the various pieces as I was implementing this, and all of the rz pieces need to be in place for the regression tests to pass. I believe the current regression tests were designed to give the same answer in rz and xy for comparison purposes. We absolutely need additional rz regression tests. It’s unknown whether this rewrite resolves the issue with triangular rz meshes. I need an input to test with.

Unfortunately, I haven't been able to find my cylindrical mesh input yet. I'm still looking though and will let you know if I track it down.

ragusa:
Why isn't something like this, based on an exisitng triangular mesh, not good enough as a first try?
```
meshgen1 = mesh.MeshGenerator.Create
({
  inputs =
  {
    mesh.FromFileMeshGenerator.Create
    ({
      filename="../../../../resources/TestMeshes/TriangleMesh2x2Cuts.obj"
    }),
  }})
mesh.MeshGenerator.Execute(meshgen1)
```

modules/linear_boltzmann_solvers/c_discrete_ordinates_curvilinear_solver/lbs_curvilinear_solver.cc
ragusa: I think it is clearer to keep things like XY, XYZ, RZ, ..., in upper case. Is this really a deviation from the standards we established?

zhardy-lanl: Yes, I agree. To me acronyms should follow the spirit of CamelCase with each distinct word getting a capital letter. I think we should revisit the change of AAH and CBC to Aah and Cbc from #130 as well.

zhardy-lanl:
> Why isn't something like this, based on an exisitng triangular mesh, not good enough as a first try?
> 
> ```
> meshgen1 = mesh.MeshGenerator.Create
> ({
>   inputs =
>   {
>     mesh.FromFileMeshGenerator.Create
>     ({
>       filename="../../../../resources/TestMeshes/TriangleMesh2x2Cuts.obj"
>     }),
>   }})
> mesh.MeshGenerator.Execute(meshgen1)
> ```

It should be good enough. The input file I am searching for used a very simple triangular mesh I generated in gmsh. It maybe had a few dozen cells max.

modules/linear_boltzmann_solvers/c_discrete_ordinates_curvilinear_solver/lbs_curvilinear_solver.cc
ragusa: technically, Pwl is also an acronym so should it remain PWL ?

zhardy-lanl: In my opinion, and as the original comment suggests, yes. I think the sweep chunk naming could use some discussion. Namely, I think we should address including `PWL` in cylindrical but not Cartesian, including `RZ` but not `XYZ`, and having `AAH` and `CBC` for Cartesian, but not for cylindrical.

wdhawkins: 100% agree. I do not like lowercase for acronyms. I prefer PWL(D), RZ, XY, AAH, etc. However, as David pointed out, this is not consistent with our coding standards. We should probably change that, I would like to broadly rename the spatial discretizations with some consistency.  Maybe something like METHOD_AGGREGATION_GEOMETRY. So, Pwlrz would become PWLD_Cellset_XYZ or PWLD_Cell_XYZ... etc. Maybe we can include this in the broader sweep organization discussion.

wdhawkins:
> > Why isn't something like this, based on an exisitng triangular mesh, not good enough as a first try?
> > ```
> > meshgen1 = mesh.MeshGenerator.Create
> > ({
> >   inputs =
> >   {
> >     mesh.FromFileMeshGenerator.Create
> >     ({
> >       filename="../../../../resources/TestMeshes/TriangleMesh2x2Cuts.obj"
> >     }),
> >   }})
> > mesh.MeshGenerator.Execute(meshgen1)
> > ```
> 
> It should be good enough. The input file I am searching for used a very simple triangular mesh I generated in gmsh. It maybe had a few dozen cells max.

If I can get an input and mesh, I'm happy to try to track this down.

ragusa:
just replace the ortho mesh for a cylindrical example (to be found in test/modules/...) with the mesh above and that should make a test file for you to try

modules/linear_boltzmann_solvers/c_discrete_ordinates_curvilinear_solver/lbs_curvilinear_solver.cc
andrsd: I would object to the use of underscores in the class names. `RZ`, `XYZ` make sense and we can add a note into naming conventions that these are ok to use. Plus, add other exceptions you know of...

modules/linear_boltzmann_solvers/c_discrete_ordinates_curvilinear_solver/sweep_chunks/lbs_curvilinear_sweep_chunk_pwl.cc
```diff
@@ -65,5 +66,308 @@ SweepChunkPWLRZ::SweepChunkPWLRZ(
...
+        for (size_t i = 0; i < cell_num_nodes; ++i)
+        {
+          const size_t imap =
+            i * groupset_angle_group_stride_ + direction_num * groupset_group_stride_ + gs_ss_begin;
+          for (int gsg = 0; gsg < gs_ss_size; ++gsg)
+            cell_psi_data[imap + gsg] = b[gsg][i];
+        }
+      }
+
+      // For outoing, non-boundary faces, copy angular flux to fluds and
```
andrsd, line 308: "outoing" -> "outgoing" (typo)

andrsd (changes requested):

wdhawkins:
> just replace the ortho mesh for a cylindrical example (to be found in test/modules/...) with the mesh above and that should make a test file for you to try

Using any of the triangle meshes in `resources` generates the following error: 
`**!**ERROR**!** D_DO_RZ_SteadyState::SteadyStateSolver::PerformInputChecks : mesh contains boundary faces with outward-oriented unit normal vector [-1 -0 -0] with vertices characterised by v(0) != 0`

ragusa:
not a single 2d mesh with origin at (0, 0, 0) ?? Is that the check that is failing?

wdhawkins:
> not a single 2d mesh with origin at (0, 0, 0) ?? Is that the check that is failing?

I created simple 2D triangular mesh in gmesh with origin at (0, 0) and extents (1,1). This ran fine with the 2g rz regression input.

ragusa:
I've sent you a file as well. did it work with mine as well? it looks like all triangular grids in the resource folder contain negative x values, which is not compatible as an rz mesh as r must be >= 0 !!

andrsd:
Please, do not merge `main` into your feature branch!

image

wdhawkins:
> Please, do not merge `main` into your feature branch!

Yeah, that was my bad. 😦

zhardy-lanl:
Interesting that it is working with triangles now... I'll have to dive back into this

zhardy-lanl (commented):
Two quick things:

1. What was the bug that was introduced and fixed?
2. I'll approve after the git history is fixed.

wdhawkins:
> Please, do not merge `main` into your feature branch!

My inexperience with git is showing here. What is the best way to fix the commit history?

wdhawkins:
> Two quick things:
> 
>     1. What was the bug that was introduced and fixed?

I introduced the bug when I removed the RZ sweepchunk kernels. I failed to recognize that the RZ sweep chunk was derived from the XY sweepchunk with the kernels providing the RZ functionality. The regression tests give the same answer in both RZ and XY, so they continued to pass. After I committed the kernel removal PR, I realized what happened, so I rewrote the RZ sweepchunk as a stand-alone sweepchunk like AAH and CBC.

andrsd:
> What is the best way to fix the commit history?

Since the merge commit is the last one, you can just remove it via `git reset --hard HEAD~1`. Then, force-push your local branch into your fork and it will fix the problem. If there are merge conflicts after the push, then you would need to fix those by rebasing you branch on top of the lates `main`.

andrsd:
Great! No merge conflicts.

andrsd (approved):

andrsd:
I swear I saw 3 commits on this branch after you pushed and I approved this. Why is there a merge commit again?

andrsd:
I see. That explains the merge commits. If I need changes from `main`, I rebase the feature branch on the updated local `main`. Typically, that is only needed when there are conflicts created by other changes that got into `main`.

zhardy-lanl:
> > I swear I saw 3 commits on this branch after you pushed and I approved this. Why is there a merge commit again?
> 
> There should only be one additional commit which was updating my fork to be current with upstream and merging those changes into my feature branch. The git workflow I'm used to is:
> 
> 1. create a feature branch in my fork
> 2. develop in my feature branch
> 3. keep my fork up to date with upstream and merge new changes from fork::main into my feature branch
> 4. When my feature branch is complete and up to date with main (for my fork) which should be up to date with main (upstream), create a PR
> 5. Update my fork and feature branch as necessary until the PR is closed... making sure that my changes continue to work with other changes across the codebase

The only difference is that instead of merging we prefer to do a rebase. When you do a merge, it will create many different merge commits in the history of the main branch. When you do a rebase, you are essentially just moving you commits to the front of the history so that when you merge, only the particular commits relevant to your feature branch show up in the history. This comes at the cost of having to force push somewhat often, but it makes for a much cleaner history.

zhardy-lanl (approved):
Remove the merge commit if you can, but good to go.

andrsd:
Also, the merge commit hoses up github PR view. So, if `main` changed, you see all those changes as part of your PR, which makes is difficult to review (in case you changed something).

zhardy-lanl:
> > > > I swear I saw 3 commits on this branch after you pushed and I approved this. Why is there a merge commit again?
> > > 
> > > 
> > > There should only be one additional commit which was updating my fork to be current with upstream and merging those changes into my feature branch. The git workflow I'm used to is:
> > > 
> > > 1. create a feature branch in my fork
> > > 2. develop in my feature branch
> > > 3. keep my fork up to date with upstream and merge new changes from fork::main into my feature branch
> > > 4. When my feature branch is complete and up to date with main (for my fork) which should be up to date with main (upstream), create a PR
> > > 5. Update my fork and feature branch as necessary until the PR is closed... making sure that my changes continue to work with other changes across the codebase
> > 
> > 
> > The only difference is that instead of merging we prefer to do a rebase. When you do a merge, it will create many different merge commits in the history of the main branch. When you do a rebase, you are essentially just moving you commits to the front of the history so that when you merge, only the particular commits relevant to your feature branch show up in the history. This comes at the cost of having to force push somewhat often, but it makes for a much cleaner history.
> 
> I've always used `merge` since it seems to lead to less conflicts when updating, but I can see the benefits in a clean history.

There's no doubt about that, haha. Rebasing can be a huge pain if significant conflicts arise.

andrsd:
Look what the merge approach did on #133 . The actual code fixing the bug were literally 3 lines. After the update by merge, the PR has 20 commits, 30 changed files, 1600+ added lines, 1900+ removed lines. That's not good.

PR #135, opened 2024-02-17
fixing logical volume to work with skin mesh

ragusa:
making sure a logical volume can be created with a skin mesh (a skin mesh is a 3D surface mesh).

due to confusion between old and new way of interacting at the lua level, the surface mesh stack (which didn't work) was replaced by the object stack in the old way. in a few weeks, this will be completely removed so I am fine leaving it as is
added two text cases in the testing folder and update to the json gold file.


test/framework/mesh/logical_volume/lv_lua_func.lua
```diff
@@ -0,0 +1,37 @@
+--[[# test for lus function used as a logical volume
+--]]
```
zhardy-lanl, line 2:
```suggestion
-- test for lua function used as a logical volume
```
If this is a block comment, why use it for one line of text?

test/framework/mesh/logical_volume/lv_boolean_test1.lua
```diff
@@ -1,3 +1,10 @@
+--[[# test for boolean operations on logical volumes
+lv1 = sphere
+lv2 = right circular cylinder (rcc)
+lv3 = in rcc but not in sphere
+--]]
```
zhardy-lanl, line 5:
```suggestion
--[[# 
test for boolean operations on logical volumes
lv1 = sphere
lv2 = right circular cylinder (rcc)
lv3 = in rcc but not in sphere
--]]
```
Is this a Lua block comment syntax?? Never seen it before that is interesting. I think it looks nicer to move the first line to its own line, if that doesn't break anything.

test/framework/mesh/logical_volume/lv_skinmesh.lua
```diff
@@ -0,0 +1,42 @@
+--[[# test for skin (surface) mesh used as a delimiter for a logical volume
+--]]
```
zhardy-lanl, line 2:
```suggestion
-- test for skin (surface) mesh used as a delimiter for a logical volume
```
Block comment for one line?

zhardy-lanl (approved):
Please address the unimportant test file comments as you see fit. I'm not sure if there is anything special about the `--[[#` and `--]]` syntax, but I assumed it is like a block comment in my suggestions.

andrsd (approved):

PR #144, opened 2024-02-19
Adding input path to runtime. Changing default split mesh file prefix.

wdhawkins:
This PR adds the input path to the openSn runtime as a `std::filesystem::path` object. This allows us to use the input path and filename stem throughout the code. In this case, I'm using the input filename stem as the default prefix for split mesh files. This prevents a whole host of issues when using the same filename prefix across multiple runs. IMHO, we should be using the input file stem for all files generated by the code instead of generic names that lack provenance.

This should fix the race condition encountered with the split file regression tests, and I have re-enabled the offending test.

There may already be a Lua method for grabbing the input file path from the command-line.  I'm not very familiar with Lua or the Lua interface yet. If such a method exists, I'm happy to use it in place of the changes I've made here.

### DIFF
--- a/framework/mesh/mesh_generator/split_file_mesh_generator.cc
+++ b/framework/mesh/mesh_generator/split_file_mesh_generator.cc
@@ -40,9 +40,8 @@ SplitFileMeshGenerator::GetInputParameters()
     "split_mesh",
     "Path of the directory to be created for containing the split meshes.");
 
-  params.AddOptionalParameter("split_file_prefix",
-                              opensn::input_path.stem().string(),
-                              "Prefix to use for all split mesh files");
+  params.AddOptionalParameter(
+    "file_prefix", opensn::input_path.stem().string(), "Prefix to use for all split mesh files");
 
   params.AddOptionalParameter(
     "read_only", false, "Controls whether the split mesh is recreated or just read.");
@@ -60,7 +59,7 @@ SplitFileMeshGenerator::SplitFileMeshGenerator(const InputParameters& params)
   : MeshGenerator(params),
     num_parts_(params.GetParamValue<int>("num_partitions")),
     split_mesh_dir_path_(params.GetParamValue<std::string>("split_mesh_dir_path")),
-    split_file_prefix_(params.GetParamValue<std::string>("split_file_prefix")),
+    file_prefix_(params.GetParamValue<std::string>("file_prefix")),
     read_only_(params.GetParamValue<bool>("read_only")),
     verbosity_level_(params.GetParamValue<int>("verbosity_level"))
 {
@@ -149,7 +148,7 @@ SplitFileMeshGenerator::WriteSplitMesh(const std::vector<int64_t>& cell_pids,
   {
     t_write.TimeSectionBegin();
     const std::filesystem::path file_path =
-      dir_path.string() + "/" + split_file_prefix_ + "_" + std::to_string(pid) + ".cmesh";
+      dir_path.string() + "/" + file_prefix_ + "_" + std::to_string(pid) + ".cmesh";
 
     std::ofstream ofile(file_path.string(), std::ios_base::binary | std::ios_base::out);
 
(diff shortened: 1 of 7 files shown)


ragusa (commented):
Good for me. But I must admit it looks a tiny bit obfuscated. Before, the meaning was crystal clear with the variable input_file_name but if this is what you have to do ... I looked a bit into lua for this, I found this:
https://stackoverflow.com/questions/6380820/get-containing-path-of-lua-file
which uses standard modules in lua. What do you guys think? @andrsd @zhardy-lanl

wdhawkins:
> #141 was also patching `framework/mesh/mesh_generator/split_file_mesh_generator.cc` which changed the `split_file_prefix` parameter in `SplitFileMeshGenerator` and did some other changes related to parallel execution. This bit seems to be missing here. Is this PR complete? Just checking...

Yes, this PR is complete. I stripped out the performance piece because I discovered a failure mode at high core counts. I'll create a separate PR for it at a later date. It wasn't really part of the bug fix anyway and should have a separate PR.

wdhawkins:
> Good for me. But I must admit it looks a tiny bit obfuscated. Before, the meaning was crystal clear with the variable input_file_name but if this is what you have to do ... I looked a bit into lua for this, I found this: https://stackoverflow.com/questions/6380820/get-containing-path-of-lua-file which uses standard modules in lua. What do you guys think? @andrsd @zhardy-lanl

I can go back to input_file_name. To me, that's not as descriptive, but I'm ok with it... especially if you think it's clearer and easier to understand.

zhardy-lanl:
Am I understanding this correctly when I make the comparison that using `std::filesystem` essentially just objectifies a path like `Pathlib` in Python does? This doesn't seem to change anything on the input side and seems to offer more benefits than just using a string for the input path.  If this is true, I like this change

zhardy-lanl:
> Good for me. But I must admit it looks a tiny bit obfuscated. Before, the meaning was crystal clear with the variable input_file_name but if this is what you have to do ... I looked a bit into lua for this, I found this: https://stackoverflow.com/questions/6380820/get-containing-path-of-lua-file which uses standard modules in lua. What do you guys think? @andrsd @zhardy-lanl

It doesn't look like this changes anything in Lua. See my last comment. It seems like we are just replacing the C++ `input_file_name` with a path-like C++ object in `std::filesystem` that has convenient routines available. I don't think this looks obfuscated, personally. We are just representing a path differently.

wdhawkins:
> Am I understanding this correctly when I make the comparison that using `std::filesystem` essentially just objectifies a path like `Pathlib` (or kind of `os`) in Python does? This doesn't seem to change anything on the input side and seems to offer more benefits than just using a string for the input path. If this is true, I like this change

Yes, that is correct. Treating input path as a std::filesystem path object provides us with a number of benefits: dissecting the path into individual pieces, checking if the path exists, etc. Much more useful than just a string.

zhardy-lanl (approved):

andrsd (dismissed):
This is good. Just the title is misleading. This PR is not "Changing default split mesh file prefix".

andrsd:
I approved, but please change the title before merging. It is good for future us searching through PRs.

wdhawkins:
> I approved, but please change the title before merging. It is good for future us searching through PRs.

That was my fault. When I removed the changes to the split mesh file generation algorithm, I also removed the change that update the file prefix. I've add that change back into the patch. It's also necessary for re-enabling the regression test.

andrsd (changes requested):
Merge the formatting patch with `ab531c3`. The usual business:

```
git rebase --interactive main
- mark the formatting patch as fixup
- save and exit
git push -f
```

wdhawkins:
Just one additional note. I know that most of the methods that create output files have the option to take a file prefix (just like the split mesh routine). IMHO, we should default to using the input file stem as the prefix. The user can certainly change the prefix if necessary, but ,this way, there is a clear association between output files and input files without any manual intervention by the user.

andrsd (approved):
:+1:

ragusa (approved):

framework/mesh/mesh_generator/split_file_mesh_generator.cc
```diff
@@ -40,8 +40,9 @@ SplitFileMeshGenerator::GetInputParameters()
     "split_mesh",
     "Path of the directory to be created for containing the split meshes.");
 
-  params.AddOptionalParameter(
-    "split_file_prefix", "split_mesh", "Prefix to use for all split mesh files");
+  params.AddOptionalParameter("split_file_prefix",
```
zhardy-lanl, line 43: I would recommend changing this to just `file_prefix` since this is always within the block for creating s split file mesh.

zhardy-lanl (approved):
Please see my recommended input parameter name change. I won't hold up the PR for the change, but please consider it before merging.

PR #142, opened 2024-02-19
Export Fix

rodriguez-jay:
Closes #132

PR #145, opened 2024-02-20
Fixing file export in parallel

### LINKED ISSUE #132: export to vtk incomplete in unbalanced parallel partitions. needed to add MPI barrier in lua for sucessful completion
This could be connected to the general pass-through we wanted to do regarding the pertinence of MPI Barriers dispersed throughout the code.

However, this is an issue that is pretty annoying that I amsingling it out. When export to VTK for a parallel simulations that has severely unbalanced partition, the code may exit before all ranks are done writing their .vtu portion. This can happen when exporting just the mesh or when exporting field functions (including multiple field functions).

I was able to fix this from lua by adding in lua an MPIBarrier() call., example below. We should not rely ont he user to know or do this so I suggest that we add barriers in the various functions that perform write to disk (export to vtk, write field functions, write multiple field functions, maybe the split mesh business as well, and possibly others.

```
MeshHandlerExportMeshToVTK("Stage_3")
MPIBarrier()
```

andrsd:
Closes #132

### DIFF
--- a/framework/field_functions/field_function_grid_based.cc
+++ b/framework/field_functions/field_function_grid_based.cc
@@ -300,6 +300,7 @@ FieldFunctionGridBased::ExportMultipleToVTK(
   WritePVTUFiles(ugrid, file_base_name);
 
   log.Log() << "Done exporting field functions to VTK.";
+  opensn::mpi_comm.barrier();
 }
 
 std::vector<double>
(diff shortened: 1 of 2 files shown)


andrsd:
This is a salvage of #142

ragusa (approved):

wdhawkins (approved):

Issue #102, opened 2024-01-31
undo KSP convergence testing


ragusa:
undo KSP convergence testing that is done in-house while it could be done directly in PETSc.

- [x] https://github.com/Open-Sn/opensn/blob/e7b713b62009db8ff3b7ca0eb480af1f8cc6cda7/framework/math/petsc_utils/petsc_utils.cc#L118 

- [ ] https://github.com/Open-Sn/opensn/blob/e7b713b62009db8ff3b7ca0eb480af1f8cc6cda7/modules/linear_boltzmann_solvers/a_lbs_solver/iterative_methods/wgs_linear_solver.cc#L48

ragusa:
@wdhawkins @andrsd : can this be closed or is there work left to do?

wdhawkins:
I'm going to close this since item #2 is addressed in #103.

PR #151, opened 2024-02-21
Removing custom relative convergence test

andrsd:
- Removing RelativeResidualConvergenceTest
- Adjusting the relative tolerance in tests so that the residual drops below the expected "gold" value

Refs #102


ragusa (approved):
if there are more items in petsc_utils.cc to remove, please go ahead. 

I was not sure about KSPMonitorStraight. What do you think? But, PR is good to go for me.

andrsd:
> I was not sure about KSPMonitorStraight.

That's dead code, we can safely delete that. Let me add a patch here for that.

wdhawkins (approved):

PR #153, opened 2024-02-27
Cell-wise densities

zhardy-lanl:
Implementation of cell-wise densities into the LBS framework. This change has the effect of the code "assuming" microscopic cross sections instead of macroscopic. This allows for more control over cross sections based on material states and flexibility in how the problem is defined. Of course, using a default density of one in all cells has the same effect as assuming a macroscopic cross sections.

### What this PR does

- Added a density attribute that gets initialized to one in `LBSSolver` and associated getters.
- Hooked up densities to source functions for scaling scattering and fission sources.
- Hooked up densities to sweepers for scaling `sigma_t`.

### What this PR doesn't do

- Read in and set cell-wise densities from files
- Include routines to set densities outside of the non-const getter.

### DIFF
--- a/modules/linear_boltzmann_solvers/a_lbs_solver/iterative_methods/power_iteration_keigen2.cc
+++ b/modules/linear_boltzmann_solvers/a_lbs_solver/iterative_methods/power_iteration_keigen2.cc
@@ -43,6 +43,7 @@ PowerIterationKEigen2(LBSSolver& lbs_solver, double tolerance, int max_iteration
   auto& q_moments_local = lbs_solver.QMomentsLocal();
   auto& phi_old_local = lbs_solver.PhiOldLocal();
   auto& phi_new_local = lbs_solver.PhiNewLocal();
+  const auto& densities_local = lbs_solver.DensitiesLocal();
   auto primary_ags_solver = lbs_solver.GetPrimaryAGSSolver();
   auto& groupsets = lbs_solver.Groupsets();
   auto active_set_source_function = lbs_solver.GetActiveSetSourceFunction();
@@ -87,11 +88,14 @@ PowerIterationKEigen2(LBSSolver& lbs_solver, double tolerance, int max_iteration
 
   /**Lambda for the creation of fission sources.*/
   auto SetLBSFissionSource =
-    [&active_set_source_function, &front_gs](const VecDbl& input, VecDbl& output)
+    [&active_set_source_function, &front_gs, &densities_local](const VecDbl& input, VecDbl& output)
   {
     Set(output, 0.0);
-    active_set_source_function(
-      front_gs, output, input, APPLY_AGS_FISSION_SOURCES | APPLY_WGS_FISSION_SOURCES);
+    active_set_source_function(front_gs,
+                               output,
+                               input,
+                               densities_local,
+                               APPLY_AGS_FISSION_SOURCES | APPLY_WGS_FISSION_SOURCES);
   };
 
   // Start power iterations
@@ -121,6 +125,7 @@ PowerIterationKEigen2(LBSSolver& lbs_solver, double tolerance, int max_iteration
     active_set_source_function(front_gs,
                                q_moments_local,
                                phi_new_local,
+                               densities_local,
                                APPLY_AGS_SCATTER_SOURCES | APPLY_WGS_SCATTER_SOURCES);
 
     frons_wgs_context->ApplyInverseTransportOperator(SourceFlags());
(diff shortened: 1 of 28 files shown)


andrsd (approved):
:+1:

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
```diff
@@ -112,7 +111,7 @@ SourceFunction::operator()(const LBSGroupset& groupset,
             if (apply_ags_scatter_src_)
               for (const auto& [_, gp, sigma_sm] : S_ell.Row(g))
                 if (gp < gs_i_ or gp > gs_f_)
-                  rhs += sigma_sm * phi[gp];
+                  rhs += rho * sigma_sm * phi_im[gp];
```
ragusa, line 114: if this really rho or the ratio of the new to the old rho? if it is rho, does it mean that sigma_sm is now a microXS ? also, when I see rho, I wrongly think of it as g/cc first, but it's atom/cc, right?

ragusa (commented):
can we have an input file in the test? something like, homogeneous medium, fully reflected, solution saturates to q/sigma_a. next, double the density and the solution should now saturate at half the previous value. simple enough but would test the change in density

modules/linear_boltzmann_solvers/a_lbs_solver/iterative_methods/power_iteration_keigen2.cc
```diff
@@ -87,11 +88,14 @@ PowerIterationKEigen2(LBSSolver& lbs_solver, double tolerance, int max_iteration
...
+    [&active_set_source_function, &front_gs, &densities_local](const VecDbl& input, VecDbl& output)
   {
     Set(output, 0.0);
-    active_set_source_function(
-      front_gs, output, input, APPLY_AGS_FISSION_SOURCES | APPLY_WGS_FISSION_SOURCES);
+    active_set_source_function(front_gs,
+                               output,
+                               input,
+                               densities_local,
+                               APPLY_AGS_FISSION_SOURCES | APPLY_WGS_FISSION_SOURCES);
```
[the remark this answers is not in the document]
zhardy-lanl: Because it goes over the 100 character limit.

modules/linear_boltzmann_solvers/a_lbs_solver/source_functions/source_function.cc
zhardy-lanl: The short answer is that sigma_sm is microscopic. As the PR description says, the internals of the code now treat cross sections as microscopic. If the user does nothing, the densities are all one, it is the same as using macroscopic cross sections. Right now, there is no precise form for rho, but it must be consistent with the microscopic cross sections. The main idea of this is that you could have a combination of microscopic cross sections based on relative isotopic composition and a number density derived from something like a material density .

zhardy-lanl:
> can we have an input file in the test? something like, homogeneous medium, fully reflected, solution saturates to q/sigma_a. next, double the density and the solution should now saturate at half the previous value. simple enough but would test the change in density

We can but it would require a bit of work since there is currently no way to set cell-wise densities. I was planning on adding a test when I eventually add a way to set the densities. This is just the infrastructure for using cell-wise densities.

ragusa:
@zhardy-lanl can you create an issue to add the test I described earlier: homogeneous medium, fully reflected, solution saturates to q/sigma_a. next, double the density and the solution should now saturate at half the previous value. simple enough but would test the change in density.

zhardy-lanl:
> @zhardy-lanl : can we agree that our units must be:
> 
> * cross sections in 1/cm when density is **not** supplied (obvious)
> * cross sections in barns and densities in at/(barn-cm) otherwise? (this is quite common in codes that deal with microXS. it avoids carrying large numbers because of Avogadro's ....).

On point 1 I agree. My only point of contention would be that there is nothing in the code that I am aware of that is explicitly tied to units (constants, etc.). If there aren't any constants tied to particular units in the code, then the code should be agnostic to units. In other words, the units of the problem are set by the units of the cross section data and densities. 

Point 2 I feel a little differently. I don't think we need to have strict enforcement of at/(barn-cm) vs. at/cc. The user supplies both the cross sections and the densities, it should be their responsibility to ensure consistency of units in that regard. When we have time-dependent solvers this becomes a bit more important as that time scales can vary dramatically based on the application making the time units of choice far less standard.

More generally, there isn't anything (constants) in the code that I know of that is tied to specific units.

ragusa:
except that the mesh and the macroXS need to be compatible. So, yes, the user should put the densities and microXS in whatever they feel like, as long as the resulting macroXS is compatible of the length units used in the mesh.

ragusa (approved):

wdhawkins (approved):

PR #154, opened 2024-02-28
First pass at reorganizing and refactoring the sweep infrastructure

wdhawkins:
NOTE: This PR implements an input-breaking change! 

This commit is mostly reorganizing files and folders. Moving framework/mesh/sweep_utilities to modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep and moving the contents of b_discrete_ordinates_solver/sweepers into b_discrete_ordinates_solver/sweep. In addition, the sweep infrastructure has been moved into the lbs namespace. As part of the namespace move, the sweep boundary enum has been deprecated in favor of the lbs boundary enum, and some of the boundary enum labels have been changed. This label change has also been propagated through to the input language to maintain consistency. This is a breaking change for existing inputs. The boundary input language has changed as follows:

`incident_isotropic -> isotropic`
`incident_anisotropic_heterogeneous -> arbitrary`

The SweepChunk class has been re-factored and simplified. The sweep infrastructure is solely contained within modules/linear_boltzmann_solvers and any sweep headers have been removed from other pieces of the code.


framework/mesh/cell/cell.cc
```diff
@@ -6,6 +6,8 @@
 #include "framework/runtime.h"
 #include "framework/logging/log.h"
 
+#include <set>
```
ragusa, line 9: so there should be an empty line to distinguish between our include files and cpp's, right?

ragusa (approved):

ragusa:
clang format is failing, though

framework/mesh/cell/cell.cc
wdhawkins: I honestly don't have strong feelings either way so long as we are consistent throughout the code. I have a large set of change coming into this commit, so I can make either change.

wdhawkins: I think single block is probably the best option. I'll update my incoming changes.

ragusa: maybe we can always list the cpp includes first, and then the ones from the code. ok?

wdhawkins: Sounds good. Easy enough.

zhardy-lanl: @ragusa My only issue with this is in class implementation files. Normally we include the class header first. This would require it to be included elsewhere. My preference is most specific to least in the order:

- modules
- framework
- external packages
- standard library

lua/modules/linear_bolzmann_solvers/lbs_solver/tools/lbs_bndry_func_lua.h
zhardy-lanl, line 1: Macroscopic comment... Can this be moved out of tools since it is the only thing in tools? Or should other files be moved in to make it useful?

framework/mesh/cell/cell.cc
wdhawkins: We should probably add this to our coding guide. I think the following order is best:

1. Header for this compilation unit (.h file that corresponds to .cc/.cpp file)
2. Other OpenSn headers (as Zach mentioned above, most specific to least)
3. Non-standard, non-system libraries
4. C++ headers
5. C headers

zhardy-lanl (approved):
This looks great, especially the bc name changes! I won't hold this up, but see my one comment on the placement of a Lua c++ file.

lua/modules/linear_bolzmann_solvers/lbs_solver/tools/lbs_bndry_func_lua.h
ragusa: a side note to Daryl. I looked into the file and saw quadrature_phi_theta_angles. I think we should move away from this data. we can talk more about this if you want (I think that everything should be in terms of mu's and phi's as the solid angle is dmu dphi, not dtheta dphi and spherical harmonics (should) take mu and phi as input, not theta and phi, as they are products of functions of **mu** and sine/cosine functions of phi)

framework/mesh/cell/cell.cc
andrsd: This is what I prefer and do as well.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -15,7 +15,9 @@
 #include <vtkExodusIIWriter.h>
 #include <vtkPointData.h>
 #include <vtkCellData.h>
+
```
andrsd, line 18: remove this line, `#include`s should be in a single block.

framework/mesh/raytrace/raytracer.cc
```diff
@@ -2,7 +2,9 @@
 #include "framework/mesh/cell/cell.h"
 #include "framework/mesh/mesh_continuum/mesh_continuum.h"
 #include "framework/logging/log.h"
+
```
andrsd, line 5: Remove this line

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/sweep_chunk.h
```diff
@@ -1,42 +1,127 @@
...
+class SweepChunk
 {
 public:
-  SweepChunk(std::vector<double>& destination_phi,
-             std::vector<double>& destination_psi,
+  /**
+   * Convenient typdef for the moment call back function. See moment_callbacks.
+   *  Arguments are:
+   *  SweepChunk *
+   *  AngleSet *
```
andrsd, line 21: I get it that `SweepChunk` and `AngelSet` are descriptive enough, then do you want to keep these with no comments? Or do you think people would miss they are arguments (i.e. they would not be familiar with the C++ syntax) and this comment would help them?

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/sweep_chunk.h
```diff
@@ -1,42 +1,127 @@
...
+  }
+
+  /**Sets all elements of the output angular flux vector to zero.*/
+  void ZeroDestinationPsi() { (*destination_psi).assign((*destination_psi).size(), 0.0); }
+
+  /**Returns a reference to the output angular flux vector.*/
+  std::vector<double>& GetDestinationPsi() { return *destination_psi; }
+
+  /**Activates or deactives the surface src flag.*/
+  void SetBoundarySourceActiveFlag(bool flag_value) // Done
```
andrsd, line 100: Can you remove `// Done`? IDK what is that for, may be copy-paste?

andrsd (changes requested):
Just a few small cosmetic problems.

I also, like the new BC names :+1:

modules/linear_boltzmann_solvers/b_discrete_ordinates_solver/sweep_chunks/sweep_chunk.h
wdhawkins: I left the description and removed the argument comment.

framework/mesh/raytrace/raytracer.cc
wdhawkins: Done.

framework/mesh/mesh_continuum/mesh_continuum.cc
wdhawkins: Done. My latest commit reformatted and reordered all of the include blocks.

lua/modules/linear_bolzmann_solvers/lbs_solver/tools/lbs_bndry_func_lua.h
wdhawkins: I'm going to defer this to a future commit, but I'll take a look. It sounds like we may end up refactoring this functionality.

andrsd (approved):

ragusa (approved):

PR #160, opened 2024-03-01
Updating regression workflow to use PETSc-3.17.5.

wdhawkins:


andrsd (approved):

ragusa:
But we are still moving to 3.19, right?

zhardy-lanl (approved):

PR #162, opened 2024-03-01
Minor fixes to support updated versions of PETSc.

wdhawkins:
This fix allows the code to support PETSc versions 3.17.0 - 3.20.5. With versions of PETSc > 3.17.5, the following regression tests fail:

1. `tutorials/fv_test1.lua`
2. `tutorials/fv_test2.lua`
3. `tutorials/pwlc_test1.lua`
4. `tutorials/pwlc_test2.lua`

It looks to me like these tests are checking the residual value at a specific iteration number to determine pass/fail. That's probably not what we want to do.

### DIFF
--- a/framework/physics/physics_utils.cc
+++ b/framework/physics/physics_utils.cc
@@ -26,12 +26,20 @@ GetPETScConvergedReasonstring(KSPConvergedReason reason)
     case KSP_CONVERGED_ITS:
       ostr << "KSP_CONVERGED_ITS";
       break;
+#if PETSC_VERSION_LT(3, 19, 0)
     case KSP_CONVERGED_CG_NEG_CURVE:
       ostr << "KSP_CONVERGED_CG_NEG_CURVE";
       break;
+#else
+    case KSP_CONVERGED_NEG_CURVE:
+      ostr << "KSP_CONVERGED_NEG_CURVE";
+      break;
+#endif
+#if PETSC_VERSION_LT(3, 19, 0)
     case KSP_CONVERGED_CG_CONSTRAINED:
       ostr << "KSP_CONVERGED_CG_CONSTRAINED";
       break;
+#endif
     case KSP_CONVERGED_STEP_LENGTH:
       ostr << "KSP_CONVERGED_STEP_LENGTH";
       break;


ragusa:
test based on iteration values are frequent. see https://github.com/Open-Sn/opensn/blob/main/test/modules/linear_boltzmann_solvers/transport_steady/tests.json 

i can see this as being a real pain when a little thing is changed in petsc (even maybe a bug fix or improvement), and now our tests can be failing. also, how will we maintain a set of golden results when there could be tiny changes due to the petsc version?

however, there could be a value in understanding why something that has passed the regression suite for so long starts to fail.

should be have golden results per supported petsc version, @wdhawkins @andrsd  ?

wdhawkins:
I'll have to give this some thought. I could quickly see this devolving into a mess depending on how many petsc versions we want to support. If we want to fix the number of versions at, say, 3 or so, it might be manageable. It's also likely to require some significant changes to the testing infrastructure.

ragusa:
@wdhawkins and @andrsd , let's talk about what to do about slight variations in golden values between petsc versions.

ragusa (approved):

andrsd:
So, few notes on the subject of testing:

1. the code in `test` does not support execution of tests based on conditions, i.e. run this test only if using PETSc version X.Y.Z / Run this test if you have VTK / etc.

2. Testing that PETSc converged the residual in exactly `N` iterations is too stringent, IMO. We are prescribing the drop in tolerance and we let PETSc do its thing. What matters is that the residual drops to meet the convergence criteria. Does not matter if it is 10 or 11 iterations. Also, do not forget that different architectures/compilers may have different floating point operations, i.e. will lead to a different final residual (this is handled by the `tol` parameter in the checks).

3. The 4 tests that were failing are IMO ad-hoc tests and very poorly designed.  For verification purposes, you do not really need to check the exact residual value and the iteration. What matters is that the **solution** meets the criteria you are testing.

This being said, IMO there is no need to introduce testing against different versions of PETSc at the moment. However, there is a huge need to double check what tests are doing and how and potentially change them. Notice, that the rest of the test suite had no problem with a new PETSc version. What I observed was that with the new PETSc version, it took longer to finish the tests (something like 480 sec vs 540 secs on my machine, IIRC).

andrsd (approved):

wdhawkins:
> So, few notes on the subject of testing:
> 
>     1. the code in `test` does not support execution of tests based on conditions, i.e. run this test only if using PETSc version X.Y.Z / Run this test if you have VTK / etc.
> 
>     2. Testing that PETSc converged the residual in exactly `N` iterations is too stringent, IMO. We are prescribing the drop in tolerance and we let PETSc do its thing. What matters is that the residual drops to meet the convergence criteria. Does not matter if it is 10 or 11 iterations. Also, do not forget that different architectures/compilers may have different floating point operations, i.e. will lead to a different final residual (this is handled by the `tol` parameter in the checks).
> 
>     3. The 4 tests that were failing are IMO ad-hoc tests and very poorly designed.  For verification purposes, you do not really need to check the exact residual value and the iteration. What matters is that the **solution** meets the criteria you are testing.
> 
> 
> This being said, IMO there is no need to introduce testing against different versions of PETSc at the moment. However, there is a huge need to double check what tests are doing and how and potentially change them. Notice, that the rest of the test suite had no problem with a new PETSc version. What I observed was that with the new PETSc version, it took longer to finish the tests (something like 480 sec vs 540 secs on my machine, IIRC).

100% agree with all of this. IMHO, if we're smart about the iterations limits in our regression inputs, we won't need to check them during testing. For example, if we are only concerned about a new version of PETSc taking 2x or 1.5x as many iterations to converge a problem, we shouldn't set the iteration limit to 1000 if we only expect it to take 10 iterations.  Appropriate iteration limits can go a long way towards improving our testing.

zhardy-lanl:
> So, few notes on the subject of testing:
> 
> 1. the code in `test` does not support execution of tests based on conditions, i.e. run this test only if using PETSc version X.Y.Z / Run this test if you have VTK / etc.
> 2. Testing that PETSc converged the residual in exactly `N` iterations is too stringent, IMO. We are prescribing the drop in tolerance and we let PETSc do its thing. What matters is that the residual drops to meet the convergence criteria. Does not matter if it is 10 or 11 iterations. Also, do not forget that different architectures/compilers may have different floating point operations, i.e. will lead to a different final residual (this is handled by the `tol` parameter in the checks).
> 3. The 4 tests that were failing are IMO ad-hoc tests and very poorly designed.  For verification purposes, you do not really need to check the exact residual value and the iteration. What matters is that the **solution** meets the criteria you are testing.
> 
> This being said, IMO there is no need to introduce testing against different versions of PETSc at the moment. However, there is a huge need to double check what tests are doing and how and potentially change them. Notice, that the rest of the test suite had no problem with a new PETSc version. What I observed was that with the new PETSc version, it took longer to finish the tests (something like 480 sec vs 540 secs on my machine, IIRC).

This seems to deserve an issue if there is not one already.  I agree that we should make our tests more resilient to changes in our dependency versions

zhardy-lanl (approved):

PR #166, opened 2024-03-06
Removing garbage from `tutorials`

andrsd:


ragusa (approved):
does this also remove the subfolders `meshing/` `first/` `angular_quad/` or just the files in them?

andrsd:
> does this also remove the subfolders `meshing/` `first/` `angular_quad/` or just the files in them?

`git` does not track directories, only files. So after we delete the files you need to remove the empty dirs yourself.

wdhawkins (approved):

PR #171, opened 2024-03-07
CreateCommonKrylovSolverSetup can specify `abs_tol` and `rel_tol`

### LINKED ISSUE #168: `CreateCommonKrylovSolverSetup` lies about the tolerance parameter
The API has this signature:

https://github.com/Open-Sn/opensn/blob/2bfe5cfd281a33089abff793218c4fa68b1fdcf4/framework/math/petsc_utils/petsc_utils.cc#L96-L103
And then `in_relative_residual_tolerance` argument is used like this:
https://github.com/Open-Sn/opensn/blob/2bfe5cfd281a33089abff793218c4fa68b1fdcf4/framework/math/petsc_utils/petsc_utils.cc#L115-L116

The `KSPSetTolerances` signature is:
```
PetscErrorCode KSPSetTolerances(KSP ksp, PetscReal rtol, PetscReal abstol, PetscReal dtol, PetscInt maxits)
```
See [doco](https://petsc.org/release/manualpages/KSP/KSPSetTolerances/)

Notice that `in_relative_residual_tolerance` is being passed into PETSc as **absolute** one 🤦‍♂️

andrsd:
Closes #168

### DIFF
--- a/framework/math/petsc_utils/petsc_utils.cc
+++ b/framework/math/petsc_utils/petsc_utils.cc
@@ -98,7 +98,8 @@ CreateCommonKrylovSolverSetup(Mat ref_matrix,
                               const std::string& in_solver_name,
                               const std::string& in_solver_type,
                               const std::string& in_preconditioner_type,
-                              double in_relative_residual_tolerance,
+                              double in_rel_tol,
+                              double in_abs_tol,
                               int64_t in_maximum_iterations)
 {
   PETScSolverSetup setup;
@@ -112,8 +113,7 @@ CreateCommonKrylovSolverSetup(Mat ref_matrix,
   KSPGetPC(setup.ksp, &setup.pc);
   PCSetType(setup.pc, in_preconditioner_type.c_str());
 
-  KSPSetTolerances(
-    setup.ksp, 1.e-50, in_relative_residual_tolerance, 1.0e50, in_maximum_iterations);
+  KSPSetTolerances(setup.ksp, in_rel_tol, in_abs_tol, 1.0e50, in_maximum_iterations);
   KSPSetInitialGuessNonzero(setup.ksp, PETSC_TRUE);
 
   KSPSetFromOptions(setup.ksp);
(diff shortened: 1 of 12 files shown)


wdhawkins (approved):

framework/math/petsc_utils/petsc_utils.cc
```diff
@@ -98,7 +98,8 @@ CreateCommonKrylovSolverSetup(Mat ref_matrix,
                               const std::string& in_solver_name,
                               const std::string& in_solver_type,
                               const std::string& in_preconditioner_type,
-                              double in_relative_residual_tolerance,
+                              double in_rel_tol,
+                              double in_abs_tol,
```
zhardy-lanl, line 102: Just making note of this for future reference: We should get rid of the `in_<var>` argument names.

modules/cfem_diffusion/cfem_diffusion_solver.cc
```diff
@@ -336,6 +336,7 @@ Solver::Execute()
                                   TextName(),
                                   KSPCG,
                                   PCGAMG,
+                                  0.,
```
zhardy-lanl, line 339: Is there a preference in the codebase for `0.` vs `0.0` in our standards? I believe the codebase primarily uses the trailing zero (which is also my preference). Not requesting a change here, but `0.0` would be welcomed where `0.` was introduced in various spots this PR.

zhardy-lanl (approved):
Please see my comment on `0.0` vs. `0.` and change if you would feel necessary. Good to go other than that.

modules/cfem_diffusion/cfem_diffusion_solver.cc
andrsd: > Is there a preference in the codebase for 0. vs 0.0 in our standards?

I do not think there is anything in the code standards regarding this.

andrsd: There is a small amount of `0.` or `<number>.` in the code. I will update this PR and send another one to unify this. Plus a change to code standards.

framework/math/petsc_utils/petsc_utils.cc
andrsd: Rough search shows about 300 occurrences. Opened issue #173 for this task.

PR #175, opened 2024-03-11
Improving tests

### LINKED ISSUE #165: test harness update needed
- [x] checks should have `abs_tol` and `rel_tol` and we should check `|value-gold| < abs_tol + rel_tol*gold`
- [ ] no longer use residual at a certain iteration for convergence
- [ ] limit the max number of iterations

andrsd:
- Reducing the number of maximum allowed iterations
- Checking that the solve converged via a prescribed tolerance. If so,
  we print "Converged" into the log and we check that via a StrCompare check

Closes #165

### DIFF
--- a/test/framework/tutorials/fv_test1.cc
+++ b/test/framework/tutorials/fv_test1.cc
@@ -112,12 +112,14 @@ SimTest01_FV(const InputParameters&)
   // Create Krylov Solver
   opensn::log.Log() << "Solving: ";
   auto petsc_solver =
-    CreateCommonKrylovSolverSetup(A, "FVDiffSolver", KSPCG, PCGAMG, 0.0, 1.0e-9, 1000);
+    CreateCommonKrylovSolverSetup(A, "FVDiffSolver", KSPCG, PCGAMG, 0.0, 1.0e-15, 25);
 
   // Solve
   KSPSolve(petsc_solver.ksp, b, x);
-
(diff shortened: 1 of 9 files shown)


ragusa (approved):

wdhawkins (approved):

test/framework/tutorials/fv_test1.cc
```diff
@@ -112,12 +112,14 @@ SimTest01_FV(const InputParameters&)
   // Create Krylov Solver
   opensn::log.Log() << "Solving: ";
   auto petsc_solver =
-    CreateCommonKrylovSolverSetup(A, "FVDiffSolver", KSPCG, PCGAMG, 0.0, 1.0e-9, 1000);
+    CreateCommonKrylovSolverSetup(A, "FVDiffSolver", KSPCG, PCGAMG, 0.0, 1.0e-15, 25);
```
zhardy-lanl, line 115: For my own edification, why bump the tolerance to 1e-15 here?

zhardy-lanl (approved):

test/framework/tutorials/fv_test1.cc
andrsd: That's most likely a left-over from when I was looking at the PETSc 3.17 vs 3.19 and tried to converged some failing tests more to see if they would pass.

PR #176, opened 2024-03-11
Use OpenSn elapsed time for regression test reporting

wdhawkins:
This MR modifies the Lua app to always report the OpenSn version header, program start time, and program elapsed time. In addition, it modifies the regression test script to pull the elapsed time from the OpenSn output file for reporting purposes. This eliminates timing overhead introduced by Python, batch job systems, etc. I've also added a new parameter to the test script: `check_numlines`. This parameter defines the number of lines after `skiplines_top` to diff for regression purposes. This should make our regression testing a bit more robust... especially for console output.

### DIFF
--- a/lua/framework/lua_app.cc
+++ b/lua/framework/lua_app.cc
@@ -58,8 +58,8 @@ LuaApp::Run(int argc, char** argv)
   console.PostMPIInfo(opensn::mpi_comm.rank(), opensn::mpi_comm.size());
 
   opensn::log.Log() << opensn::name << " version " << GetVersionStr();
-  opensn::log.Log() << Timer::GetLocalDateTimeString() << " Running " << opensn::name
-                    << " in interactive-mode with " << opensn::mpi_comm.size() << " processes.";
+  opensn::log.Log() << Timer::GetLocalDateTimeString() << " Running " << opensn::name << " with "
+                    << opensn::mpi_comm.size() << " processes.";
   opensn::log.Log() << opensn::name << " number of arguments supplied: " << argc - 1;
   opensn::log.LogAll();
   console.FlushConsole();
--- a/lua/framework/lua_app.cc
+++ b/lua/framework/lua_app.cc
@@ -24,8 +24,6 @@ const std::string command_line_help_string_ =
   "     a=b                         Executes argument as a lua string. "
   "i.e. x=2 or y=[[\"string\"]]\n"
   "     --allow-petsc-error-handler Allows petsc error handler.\n"
-  "     --suppress-beg-end-timelog  Suppresses time logs at the \n"
-  "                                 beginning and end of execution.\n"
   "     --suppress-color            Suppresses the printing of color.\n"
   "                                 useful for unit tests requiring a diff.\n"
   "     --dump-object-registry      Dumps the object registry.\n"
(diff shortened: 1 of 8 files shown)


wdhawkins:
The `solver_info_0N` tests are not good regression tests as designed. I'm going to look into refactoring them as part of this commit.

wdhawkins:
> The `solver_info_0N` tests are not good regression tests as designed. I'm going to look into refactoring them as part of this commit.

I punted on trying to design new regression tests. Although I still don't think these are necessarily the best, I can see the utility in diffing against console output. Along those lines, I've added a new parameter to the testing script: `check_numlines`. This parameter defines the number of lines (after `skiplines_top`) to diff for regression purposes. With these two parameters, we should be able to select any contiguous section of output data to diff for our regression tests.

lua/framework/lua_app.cc
```diff
@@ -59,6 +57,13 @@ LuaApp::Run(int argc, char** argv)
   opensn::Initialize();
   console.PostMPIInfo(opensn::mpi_comm.rank(), opensn::mpi_comm.size());
 
+  opensn::log.Log() << opensn::name << " version " << GetVersionStr();
+  opensn::log.Log() << Timer::GetLocalDateTimeString() << " Running " << opensn::name
+                    << " in interactive-mode with " << opensn::mpi_comm.size() << " processes.";
```
andrsd, line 62: Originally, OpenSn was printing if it runs in interactive or batch mode. This changed here and reports only interactive (even for the batch mode). However, is that really needed? May be we just print:
```
<time> Running OpenSn with <n> processors.
```

If you wanted to keep batch/interactive, you need to look at `sim_option_interactive_` and print either `interactive` or `batch` here.

test/src/test_slot.py
```diff
@@ -134,8 +128,18 @@ def PerformChecks(self):
...
 
-        time_taken_message = " {:.2f}s".format(self.time_end - self.time_start)
+        opensn_elapsed_time_sec = 0.0
+        if os.path.exists(output_filename):
+          for line in open(output_filename, 'r'):
+            found = re.search("Elapsed execution time:", line)
+            if found:
+              values_slice = re.split(r'[,:]',line.strip())
+              opensn_elapsed_time_sec = float(values_slice[1])*3600 + float(values_slice[2])*60 + float(values_slice[3])
+              break;
```
andrsd, line 138: remove the trailing `;`

test/src/test_slot.py
```diff
@@ -134,8 +128,18 @@ def PerformChecks(self):
         width = 120 - len(prefix + test_file_name) + pad
         message = message.rjust(width, ".")
 
-        time_taken_message = " {:.2f}s".format(self.time_end - self.time_start)
+        opensn_elapsed_time_sec = 0.0
+        if os.path.exists(output_filename):
+          for line in open(output_filename, 'r'):
+            found = re.search("Elapsed execution time:", line)
+            if found:
+              values_slice = re.split(r'[,:]',line.strip())
```
andrsd, line 136: Missing space after a `,`.

andrsd (changes requested):

lua/framework/lua_app.cc
wdhawkins: Nice catch. I removed the interactive piece as you suggested.

test/src/test_slot.py
wdhawkins: Fixed.

test/src/test_slot.py
wdhawkins: Fixed. C++ habit showing up.

andrsd (approved):

ragusa (approved):

zhardy-lanl (approved):

Issue #86, opened 2024-01-22
Code audit: usage of typedefs


andrsd:
Since we are using C++17, we can use the new way of defining types, which is using the [`using` statement](https://en.cppreference.com/w/cpp/language/type_alias).

Audit the code for `typedef` occurrences and change them to `using`.

PR #179, opened 2024-03-14
Refactoring sweep communicators.

wdhawkins:
Initial refactoring of `sweep/communicators`. No major code changes. Slight performance improvements here and there. Changing `EAGER_LIMIT` (which was very poorly named, imho) to `max_mpi_message_size`. It might be worth it to change the name to `sweep_max_mpi_message_size` or something similar. Thoughts?


modules/linear_boltzmann_solvers/discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -1192,15 +1189,15 @@ DiscreteOrdinatesSolver::InitFluxDataStructures(LBSGroupset& groupset)
...
-          auto angleSet = std::make_shared<TAAH_AngleSet>(angle_set_id++,
-                                                          gs_ss_size,
-                                                          gs_ss,
-                                                          *sweep_ordering,
-                                                          fluds,
-                                                          angle_indices,
-                                                          sweep_boundaries_,
-                                                          options_.sweep_eager_limit,
-                                                          *grid_local_comm_set_);
+          auto angleSet = std::make_shared<AAH_AngleSet>(angle_set_id++,
```
andrsd, line 1192: `angleSet` -> `angle_set`

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/communicators/aah_async_comm.cc
```diff
@@ -44,228 +44,201 @@ AAH_ASynchronousCommunicator::ClearLocalAndReceiveBuffers()
...
     if (num_unknowns > 0)
     {
-      deplocI_message_blockpos[deplocI].push_back(dep_block_pos);
-      deplocI_message_size[deplocI].push_back(num_unknowns);
+      deplocI_message_blockpos_[deplocI].push_back(dep_block_pos);
+      deplocI_message_size_[deplocI].push_back(num_unknowns);
     }
 
-    deplocI_message_request.emplace_back(message_count);
+    deplocI_message_request_.emplace_back(message_count, MPI_Request());
```
andrsd, line 194: Technically, `mpi::Request()` to match the underlying type.

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/communicators/aah_async_comm.h
```diff
@@ -24,42 +22,44 @@ class FLUDS;
...
+                               int max_mpi_message_size,
                                const MPICommunicatorSet& in_comm_set);
+
+  int GetMaxNumMessages() const { return max_num_messages_; }
+
+  void SetMaxNumMessages(int count) { max_num_messages_ = count; }
+
   /**
-   * Returns the private flag done_sending.
+   * Returns the private flag done_sending_.
```
andrsd, line 62: This seems like a product of search-and-replace, but it would deserve better wording

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/communicators/cbc_async_comm.h
```diff
@@ -51,6 +53,9 @@ class CBC_ASynchronousCommunicator : public AsynchronousCommunicator
 protected:
   const size_t angle_set_id_;
   CBC_FLUDS& cbc_fluds_;
+
+  // location_id, cell_global_id, face_id
+  typedef std::tuple<int, uint64_t, unsigned int> MessageKey;
```
andrsd, line 58: This is not in code standard yet, but the preference is to use `using` over `typedef` (see #86), so:
```
using MessageKey = std::tuple<int, uint64_t, unsigned int>;
```

andrsd (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
wdhawkins: Fixed. I'll have a future PR with a refactoring of the DiscreteOrdinatesSolver class as well.

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/communicators/aah_async_comm.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/communicators/cbc_async_comm.h
wdhawkins: Fixed.

zhardy-lanl (approved):
Looks like there are conflicts, but otherwise good to go from me. The renaming of `EAGER_LIMIT` is very welcome

ragusa (commented):
Eager is one protocol of mpi, along with rendezvous. I am not opposed to the change of the name, however, please consider hinting to the developers that will read this later that we are in eager mode. Others please chime in if I am too rusty on mpi protocols.

wdhawkins:
> Eager is one protocol of mpi, along with rendezvous. I am not opposed to the change of the name, however, please consider hinting to the developers that will read this later that we are in eager mode. Others please chime in if I am too rusty on mpi protocols.

This is one of the reasons that I removed that language. We are not necessarily using the eager protocol, and we shouldn't be thinking like that.  Eager is an implementation artifact and is very dependent on the interconnect. Just setting a specific message size does not imply that we are using eager - it may also require tuning MPI to adjust the eager threshold. In my experience, most eager thresholds are very small (think 2k - 4k), and I doubt we are using eager on most systems with a message size of 32,000 (which is altogether a weird number... this should be 32k, not 32,000). I'm not saying we shouldn't be tuning our message sizes for the best performance on a particular platform, but I am saying we shouldn't just sweep everything under the "eager" rug and say that's the best thing to do and it's going to give the best performance - that's just not accurate.

PR #183, opened 2024-03-18
Print only file name in test results

andrsd:
Fixing regression introduced by #176

### DIFF
--- a/test/src/test_slot.py
+++ b/test/src/test_slot.py
@@ -139,7 +139,7 @@ class TestSlot:
 
         time_taken_message = " {:.1f}s".format(opensn_elapsed_time_sec)
 
-        print(test_path + message + time_taken_message)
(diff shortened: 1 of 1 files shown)


wdhawkins:
> Also, the number of ranks used for the test is gone. Was that intentional @wdhawkins ?
> 
> We see now:
> 
> ```
> framework/mesh/read_wavefront_obj1.lua........................................................................Passed 0.0s
> ```
> 
> instead of
> 
> ```
> [1] framework/mesh/read_wavefront_obj1.lua........................................................................Passed 0.0s
> ```
> 
> I can add that back too if in favor...

That was not intentional. I modified the output line at one point for testing, and I must have munged it when I reconstructed it.

andrsd:
Ok. I will add that back.

ragusa (approved):

wdhawkins (approved):

PR #186, opened 2024-03-20
Adding CMake option to optimize for build architecture

wdhawkins:


andrsd:
`cmake` should be able to support custom build configuration via [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html). So, I am thinking if we should support this via something like `-DCMAKE_BUILD_TYPE=Native` or (Optimized). I did not play with this much, but I would like to give that approach a try. What do you think about this idea? However, it would not capture scenarios like `Debug+native` which your approach allows...

wdhawkins:
> `cmake` should be able to support custom build configuration via [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html). So, I am thinking if we should support this via something like `-DCMAKE_BUILD_TYPE=Native` or (Optimized). I did not play with this much, but I would like to give that approach a try. What do you think about this idea? However, it would not capture scenarios like `Debug+native` which your approach allows...

I do think I prefer the custom build configuration. Not supporting Debug+Native isn't a big deal... that would probably be a very rarely used combination.

wdhawkins:
It looks like there are more elegant ways to set flags for custom build types using `SET(CMAKE_CXX_FLAGS_MYBUILDTYPE...)`, but I didn't see a way to call something like `CheckCXXCompilerFlag` without using an `if` block.

andrsd:
I am looking into that too. But I don't see my custom flags set at all. I think we will need the check, clang before 15 on Apple M1 does not support `-march=native` so we have to check that.

andrsd:
Can you try [this](https://github.com/andrsd/opensn/commit/598d3ff3733b08d93420d4fa610247079489189e)? I had to add `CACHE STRING "" FORCE` to the `set` commands.

andrsd:
One side-effect in that patch is that if users override the NATIVE flags, we override it back :-) So to prevent that, the block for setting the flags would have to be enclosed in:
```
if (NOT DEFINED CMAKE_CXX_FLAGS_NATIVE)
   set(...)
   ...
endif()
```

andrsd (approved):

ragusa (approved):

zachhardy (approved):

Issue #191, opened 2024-03-27
Improve API for interaction with Lua interpreter


andrsd:
Create new API to process function arguments:

- To get a parameter from a Lua stack use:
   ```
   auto var = LuaArg<T>(L, arg_num)
   ```
   This would store the `arg_num` argument into `var` variable (plus would do all error handing like checking that the types match, etc.)
   ```
   auto var = LuaArgOptional<T>(L, arg_num, default_value);
   ```
   This would retrieve `arg_num` argument if it was specified. If not, then `var` would be initialized with `default_value`.

- To check that arguments on the Lua stack have the expected type and there is enough of them,
   ```
   LuaCheckArgs<T1, T2, ..., TN>(L, fname);
   ```
   This would check that there is `N` arguments on the stack and their types are `T1`, `T2`, ... `TN`.

- To put results on the Lua stack for further processing in the script,
   ```
   LuaReturn(L, arg1, arg2, ..)
   ```
   This will put `argX`s on the stack in the selected order and return the number of arguments which need to be passed back to the interpreter. The usage is designed for doing `return LuaReturn(L, ...)`. If no arguments are stored on the stack, use `LuaReturn(L)`.

andrsd:
I have this implemented but need to wait on #189 to go in, since it is touching the lua layer. Just saying for coordination reasons...

andrsd:
Code is in my fork on the [lua-arg](https://github.com/andrsd/opensn/commits/lua-args/) branch

PR #192, opened 2024-03-27
Pass correct handle into mat.{Add|Set}Property

andrsd:
mat.{Add|Set}Property takes a handle not a string value of an index.


andrsd:
This does not trigger any errors, since `lua_tonumber` called on a string parameter that contains a number converts the number into the integer value.  This is a preparation for #191 .

ragusa (approved):

wdhawkins (approved):

zhardy-lanl (approved):

PR #182, opened 2024-03-15
Introducing LUA modules

andrsd:
- Moving lua API related to logging into a log module
- Moving lua API related to materials into a `mat` module
- Moving lua API related to solver into a `solver` module
- Moving lua API related to logical volumes into a `logvol` module
- Cleanup in the lua mesh module
- Moving lua API related to mesh into a `mesh` module
- Moving surface mesh related API into a single .cc file
- Moving cross-section related API into a `xs` module
- Moving lua CFEM diffusion solver API into a `diffusion` module
- Moving lua DFEM diffusion solver API into a `diffusion` module
- Moving lua FV diffusion solver API into a `diffusion` module
- Moving lua CFEM MG diffusion solver API into a `diffusion` module
- Deleting obsolete LBS lua API
- Moving lua LBS solver related API into the `lbs` module
- Deleting PrintToPythonSLDFESQAngularQuadrature
- Moving lua angular quadrature related API into the `aquad` module
- Moving lua field function related API into the `fieldfunc` module
- Updating tutorials to the new syntax
- Moving lua prk module API into the `prk` module

Closes #122

PR #193, opened 2024-03-27
Pass correct handle into `aquad.OptimizeForPolarSymmetry`

andrsd:
A nil value was really passed into `aquad.OptimizeForPolarSymmetry` since there was a typo in the lua variable name.


andrsd:
This is also a preparation for #191. This fixes the problem reported by @ragusa [here](https://github.com/Open-Sn/opensn/pull/182#discussion_r1528554254) when a `nil` value is silently converted into `0`

ragusa (approved):

wdhawkins (approved):

zhardy-lanl (approved):

PR #194, opened 2024-03-27
Removing obsolete API `logvol.Create`

andrsd:
This API has been replaced by `logvol.LVObject.Create`. The only files using this API are the old tutorials. However, they no longer run. There is some `AddRegion` API that does not exist anymore. Also there are active tests that do the same thing. 
So, I am proposing to remove these outdated tutorials. New tutorials can be created to replace those later, if desired.


ragusa:
I created the 4th tutorial and it was very useful to find a bug in interior-penalty diffusion. (https://github.com/chi-tech/chi-tech/commit/f58caf7603dc03d568cd49360e301b024c5fb37f#diff-b5e3d00b4cd9c54df4360ac81cb71958ecd97b2f403fe9207dcaba340c7c0f44)
I would like to keep this and update it for a tutorial. What part exactly is broken with the new logvol create?

andrsd:
There is nothing broken in `logvol.Create` as far as I can tell. There is almost direct mapping from `logvol.Create` to the new  API style `logvol.LVObject.Create({})`, so any old files can be updated very easily.

The problem is that the old tutorials are using `SolverAddRegion`, see [here](https://github.com/Open-Sn/opensn/blob/main/tutorials/Tutorial02.lua#L31) for example and that API does not exist anywhere in the current code.

This is why I think this API can go away. Any new input files should be using `logvol.LVObject.Create({})` rather then `logvol.Create(...)`.

andrsd:
After reading my description message I see where I might have confused you. It should have said: 

> The only files using **the old API** are the old tutorials.

wdhawkins (approved):

zhardy-lanl (approved):

PR #195, opened 2024-03-28
Correcting the CXX_FLAGS condition for Native build

andrsd:
The variable is always present, thus we never enter the if statement. Changing
this to test if the variable is empty. If it contains something, we still honor that.


andrsd:
My insufficient testing mislead @wdhawkins to this bug. To redeem myself, here is a fix for it :-)

ragusa (approved):

wdhawkins (approved):

PR #196, opened 2024-04-01
Remove event-based timing logger and add Caliper support

wdhawkins:
This PR removes the event-based timing logger (that was only used for sweep timings) and adds Caliper support (https://github.com/LLNL/Caliper). This results in a net removal of ~1000 lines of code and significantly improves our timing and performance profiling abilities. The additional code required to support Caliper is minimal. I have more or less fully instrumented steady-state LBS solves. Instrumenting the code is relatively quick and easy, typically requiring only the addition of a `CALI_CXX_MARK_FUNCTION` or `CALI_CXX_MARK_SCOPE` macro at the top of a function. Other more complicated annotations are also supported.

By default, Caliper timings are disabled. A default timing/memory report can be generated using the --caliper command line flag:

`srun -n 4 lua/opensn --caliper transport_3d_5_cycles_2.lua`

Following the completion of the run, Caliper will print a timing report to `stdout` similar to the following:
```
Path                                             Min time/rank Max time/rank Avg time/rank Time %
Simulation                                           57.849906     57.928184     57.887312 99.999600
  SteadyStateSolver::Initialize                      27.172302     27.172427     27.172395 46.939969
    DiscreteOrdinatesSolver::Initialize              27.172274     27.172376     27.172350 46.939891
      LBSSolver::Initialize                          19.486737     19.486840     19.486814 33.663224
        LBSSolver::InitializeMaterials                0.000133      0.000134      0.000133  0.000231
        LBSSolver::InitializeSpatialDiscretization   19.387839     19.387851     19.387842 33.492252
          LBSSolver::ComputeUnitIntegrals            19.184495     19.184509     19.184499 33.140979
        LBSSolver::InitializeGroupsets                0.000072      0.000080      0.000078  0.000134
```

The Caliper configuration can be customized using the `--caliper=` option. For example, the following command line:

`srun -n 4 lua/opensn --caliper=runtime-report(output=cali.txt),mpi.message.count transport_3d_5_cycles_2.lua`

will generate a profiling report similar to the above, but also include mpi functions/message counts and write the output to the file `cali.txt`. Caliper includes support for CUDA, ROCM, MPI, varoius third-party profilers, and numerous other options. Caliper support is embedded in the runtime so that it is also available if OpenSn is used as a third-party library.

Replacing the grind time calculation that was previously using the event-based timers required only a dozen or so lines of code. I have promoted it to be always "on" and removed the chunk overhead and src calculation times. The src calculation time is available from the Caliper report, and the chunk overhead  was not a useful stat, IMHO.


ragusa:
Reg test failing for now.

wdhawkins:
> Reg test failing for now.

I forgot to install Caliper on the class01 runner. Should be running now.

framework/runtime.cc
```diff
@@ -24,34 +17,40 @@ namespace opensn
...
 int
 Initialize()
 {
-  auto& t_main = log.CreateTimingBlock(opensn::name);
-  t_main.TimeSectionBegin();
+  if (use_caliper)
+  {
+    cali_mgr.add(cali_config.c_str());
+    cali_set_global_string_byname("opensn.version", GetVersionStr().c_str());
+    cali_set_global_string_byname("opesn.input", input_path.c_str());
```
andrsd, line 48: Typo: `opesn.input` -> `opensn.input`

framework/runtime.cc
```diff
@@ -24,34 +17,40 @@ namespace opensn
...
+    cali_mgr.start();
+  }
+
+  CALI_MARK_BEGIN("Simulation");
```
andrsd, line 52: Would it make sense to mark this section "OpenSn simulation"? If so, then use `opensn::name + " Simulation"`. 

In case OpenSn runs coupled with another code using caliper and the code also used "Simulation", would those 2 sections be merged into single chunk? Just asking...

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/angle_aggregation/angle_aggregation.cc
```diff
@@ -233,6 +240,9 @@ AngleAggregation::InitializeReflectingBCs()
 std::pair<size_t, size_t>
 AngleAggregation::GetNumDelayedAngularDOFs()
 {
+  CALI_CXX_MARK_SCOPE("AngleAggregation::GetNumDelayedAngularDOFs");
+  ;
```
andrsd, line 244: Extra `;`

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/angle_aggregation/angle_aggregation.cc
```diff
@@ -280,6 +290,9 @@ AngleAggregation::GetNumDelayedAngularDOFs()
 void
 AngleAggregation::AppendNewDelayedAngularDOFsToArray(int64_t& index, double* x_ref)
 {
+  CALI_CXX_MARK_SCOPE("AngleAggregation::AppendNewDelayedAngularDOFsToArray");
+  ;
```
andrsd, line 294: Extra `;`

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/scheduler/sweep_scheduler.cc
```diff
@@ -246,20 +232,14 @@ SweepScheduler::ScheduleAlgoDOG(SweepChunk& sweep_chunk)
...
 {
-  typedef AngleSetStatus Status;
-
-  log.LogEvent(sweep_event_tag_, Logger::EventType::EVENT_BEGIN);
-
-  auto ev_info_i = std::make_shared<Logger::EventInfo>(std::string("Sweep initiated"));
+  CALI_CXX_MARK_SCOPE("SweepScheduler::ScheduleAlgoFIFO");
 
-  log.LogEvent(sweep_event_tag_, Logger::EventType::SINGLE_OCCURRENCE, ev_info_i);
+  typedef AngleSetStatus Status;
```
andrsd, line 242: Can you get rid of this? I see some code below already using `AngleSetStatus`

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/scheduler/sweep_scheduler.cc
```diff
@@ -398,6 +352,9 @@ SweepScheduler::ZeroOutgoingDelayedPsi()
 void
 SweepScheduler::ZeroOutputFluxDataStructures()
 {
+  CALI_CXX_MARK_SCOPE("SweepScheduler::ZeroOutputFluxDataStructures");
+  ;
```
andrsd, line 356: Extra `;`

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/wgs_linear_solver.cc
```diff
@@ -141,6 +153,9 @@ WGSLinearSolver::SetInitialGuess()
 void
 WGSLinearSolver::SetRHS()
 {
+  CALI_CXX_MARK_SCOPE("WGSLinearSolver::SetRHS");
+  ;
```
andrsd, line 157: Extra `;`

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -794,6 +788,9 @@ LBSSolver::SetBoundaryOptions(const InputParameters& params)
 void
 LBSSolver::Initialize()
 {
+  CALI_CXX_MARK_SCOPE("LBSSolver::Initialize");
+  ;
```
andrsd, line 792: Extra `;`

andrsd (changes requested):

andrsd:
Another request I have. At least the install instructions need to be updated on how to compiler Caliper. We cannot rely on people knowing that. I, for example, compiled with `WITH_MPI=ON` (which was not on by default) and `WITH_KOKKOS=OFF` (which was ON by default). I see some 20+ options there, so just to be on the same page.

andrsd:
Also, this is much nicer that the Event/Timing code. Glad to see the code getting leaner and reducing the maintenance burden :+1:

andrsd:
One more comment, should we be using `__FUNCTION__` in the `CALI_XYZ` macros. That should expand just to the function name, but may not be supported by **all** compilers. Seems like gcc and clang do the desired expansion.

andrsd:
I needed to include this patch to be able to build on MacOS X:
```
diff --git a/lua/CMakeLists.txt b/lua/CMakeLists.txt
index 39bd62fb..3fd8640e 100644
--- a/lua/CMakeLists.txt
+++ b/lua/CMakeLists.txt
@@ -18,6 +18,7 @@ target_link_libraries(libopensnlua
     ${PETSC_LIBRARY}
     MPI::MPI_CXX
     libopensn
+    caliper
 )
 
 target_compile_definitions(libopensnlua PRIVATE OPENSN_WITH_LUA)
```

I think, it is because caliper is privately linked in `libopensn`, but then it is used in the lua layer, so symbols are not resolved. This was the linker error message:
```
Undefined symbols for architecture arm64:
  "cali::ConfigManager::flush()", referenced from:
      opensnlua::LuaApp::Run(int, char**) in lua_app.cc.o
  "cali::ConfigManager::check(char const*, bool) const", referenced from:
      opensnlua::LuaApp::ParseArguments(int, char**) in lua_app.cc.o
ld: symbol(s) not found for architecture arm64
```

framework/runtime.cc
wdhawkins: Nice catch!

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/angle_aggregation/angle_aggregation.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/scheduler/sweep_scheduler.cc
wdhawkins: Fixed throughout the file. I'll see if I can refactor the reaming pieces of `sweep` this week.

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/scheduler/sweep_scheduler.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/wgs_linear_solver.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed.

wdhawkins:
> Another request I have. At least the install instructions need to be updated on how to compile Caliper. We cannot rely on people knowing that. I, for example, compiled with `WITH_MPI=ON` (which was not on by default) and `WITH_KOKKOS=OFF` (which was ON by default). I see some 20+ options there, so just to be on the same page.

Yes... good call. I completely forgot about documentation.  I'll update install instructions and the configure_dependencies.py script.

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/angle_aggregation/angle_aggregation.cc
wdhawkins: Fixed.

framework/runtime.cc
```diff
@@ -69,7 +69,7 @@ Finalize()
   material_stack.clear();
   multigroup_xs_stack.clear();
   function_stack.clear();
- 
+
   CALI_MARK_END("Simulation");
```
andrsd, line 73: Should this be:
```
CALI_MARK_END(opensn::name.c_str());
```
so that is matches the `CALI_MARK_BEGIN`?

wdhawkins: Yes. I forgot to get that in my first round of fixes. I have a final commit to this PR that fixes that, updates the install instructions, and updates `configure_dependencies.py`.

ragusa (approved):

wdhawkins:
> Another request I have. At least the install instructions need to be updated on how to compile Caliper. We cannot rely on people knowing that. I, for example, compiled with `WITH_MPI=ON` (which was not on by default) and `WITH_KOKKOS=OFF` (which was ON by default). I see some 20+ options there, so just to be on the same page.

Done! I've updated the Linux and MacOS install instructions and configure_dependencies.py.

PR #202, opened 2024-04-03
Adding a layer for interaction with the Lua interpreter

### LINKED ISSUE #191: Improve API for interaction with Lua interpreter

andrsd:
- Better handling of lua arguments for basic types
- Adding API for setting global variables
- Adding API for pushing value onto Lua stack
- Adding API to process vector-valued arguments via lua
- Adding API to push vector of values onto Lua stack
- Adding API to push (key, value) pairs onto Lua stack
- Vector3 is treated as a table in the Lua layer
- Can store std:vector<opensn:TYPE> onto lua stack
- Adding API to call Lua functions from the Lua layer
- Adding LuaArg for std::vector<T>
- Adding API to check number and type of required Lua function arguments
- Adding API for returning values back into Lua interpreter
- Using LuaArg<std::vector<T>>() instead of LuaPopulateVectorFrom1DArray
- Adding argument checking into functions
- ParameterBlock can be used in the Lua API
- log.SetVerbosity takes 1 parameter, so it matches the C++ API
- aquad.GetProductQuadrature uses new way to return result to Lua interpreter
- ExportMultiFieldFunctionToVTK takes vector of field function handles as a 1st argument
- Adding support to handle std::map in the Lua API
- Args variable is set using new Lua API
- SolverGetFieldFunctionList is using new Lua API
- ComputeLeakage is using new Lua API
- FFInterpolationGetValue is using new Lua API
- LBSGetScalarFieldFunctionList is using new Lua API

Closes #191


wdhawkins (approved):

wdhawkins:
Very nice! 🎉

ragusa (approved):

Issue #87, opened 2024-01-22
Update installation instructions


andrsd:
The installation instructions needs to be updated.

PR #203, opened 2024-04-03
Update Linux install instructions and configure_dependencies script

wdhawkins:
This PR updates the Linux install instructions and configure_dependencies.py script. I've made some simplifications to configure_dependencies.py to make the automated install less error prone (forcing the use of curl for downloads, removing march=native compiler flags, etc.). Both sets of instructions were successfully tested on a clean install of Ubuntu 22.04.4. 

This should partially address #87, but the MacOS installation instructions may still need updating.

I've testing the MacOS instructions and made slight modifications.

### DIFF
--- a/doc/install_linux.md
+++ b/doc/install_linux.md
@@ -33,7 +33,7 @@ git cmake libopenmpi-dev flex bison \
 libncurses5-dev python3-pip doxygen sphinx
 ```
 
-To ensure that all installations downstream of this install with the required mpi wrappers, the
+To ensure that all downstream installations use the required MPI compiler wrappers, the
 following environment variables must be set:
 
 ```bash
@@ -64,7 +64,7 @@ To configure **PETSc** with the required build options, run:
 
 ```bash
 ./configure  \
---prefix=$PWD/../petsc-3.17.0-install  \
+--prefix=/path/to/dependencies/directory  \
 --download-hypre=1  \
 --with-ssl=0  \
 --with-debugging=0  \
@@ -92,12 +92,7 @@ PETSC_DIR=$PWD
 
 If the configuration fails, consult **PETSc**'s user documentation. 
 
-Follow the **PETSc** build
-prompts to complete the **PETSc** installation.
-
-Add **PETSc** to your `CMAKE_PREFIX_PATH` environment variable, e.g.:
-
-`export CMAKE_PREFIX_PATH=/path/to/dependencies/petsc-3.17.0-install:$CMAKE_PREFIX_PATH`
+Follow the **PETSc** build prompts to complete the **PETSc** installation.
 
 ## Step 3 - Install the Visualization Tool Kit (VTK)
 
@@ -111,7 +106,7 @@ tar -zxf VTK-9.3.0.tar.gz
 cd VTK-9.3.0
 mkdir build
 cd build
-cmake -DCMAKE_INSTALL_PREFIX=$PWD/../install  + \
+cmake -DCMAKE_INSTALL_PREFIX=/path/to/dependencies/directory  + \
 -DBUILD_SHARED_LIBS:BOOL=ON  + \
 -DVTK_Group_MPI:BOOL=ON  + \
 -DVTK_GROUP_ENABLE_Qt=NO  + \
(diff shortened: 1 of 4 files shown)


ragusa (approved):

doc/install_linux.md
```diff
@@ -1,98 +1,66 @@
...
-
-Now, install the remaining packages needed to build OpenSn and its dependencies:
-
-```bash
-sudo apt-get install cmake python git zlib1g-dev libx11-dev unzip
-```
-
-## Step 2 - An MPI Flavor
+1. A recent version of clang++/g++ that supports C++17
+2. gfotran (required by the BLAS component of PETSc)
```
andrsd, line 17: Typo: `gfotran` -> `gfortran`

doc/install_linux.md
```diff
@@ -171,134 +125,111 @@ cmake -DCMAKE_INSTALL_PREFIX=$PWD/../install  + \
...
+```
 
-To check if the code compiled correctly, execute the test scripts:
+To configure OpenSn for building the documentation, in addition to the OpenSn application, add the
+`-DOPENSN_WITH_DOCS` option to `cmake`:
 
 ```bash
-    $ test/run_tests -d test/ -j8
+    $ mkdir build
+    $ cd bukld
```
andrsd, line 200: Typo: `bukld` -> `build`

doc/install_linux_easy.md
```diff
@@ -1,176 +1,137 @@
...
-(which OpenSn uses as a linear algebra backend)
-- OpenGL has become optional and VTK can be used, through ViSiT and Paraview, for visualization.
+## Step 1 - Install Development Tools
 
-The above packages will therefore need to be installed.
+The following packages are required for OpenSn installation and development:
 
-Check to see if gcc is installed
+1. A recent version of clang++/g++ that supports C++17
+2. gfotran (required by the BLAS component of PETSc)
```
andrsd, line 17: Typo: `gfotran` -> `gfortran`

resources/configure_dependencies.py
```diff
@@ -662,53 +561,24 @@ def InstallVTK(pkg: str, ver: str, gold_file: str):
...
-    lua_version = package_info["lua"][VERSION]
-    petsc_version = package_info["petsc"][VERSION]
-    vtk_version = package_info["vtk"][VERSION]
-
-    petsc_dir = f"{install_dir}"
-    vtk_dir = f"{install_dir}"
+    #lua_version = package_info["lua"][VERSION]
+    #petsc_version = package_info["petsc"][VERSION]
+    #vtk_version = package_info["vtk"][VERSION]
+    #caliper_version = package_info["caliper"][VERSION]
```
andrsd, line 567: If this is not needed, then remove those lines...

andrsd (changes requested):

andrsd:
General comment on the use of `CMAKE_PREFIX_PATH`. Previously, the dependencies like Lua, PETSc and VTK were installed into 3 different locations and users had to set 3 env variables. This is quite silly, since all the packages we are using are designed such that they can be installed into single location, e.g. `/usr/local` but anything really. Note that the new `configure_dependencies.py` script installs all packages into a single location and then tells the user to use `export CMAKE_PREFIX_PATH=/location/with/deps`.

The `install_linux.md` instructions tell users to install into 2 locations, one for PETSc and one for VTK. They should really install into a single location and then set the CMAKE_PREFIX_PATH into that. It is simpler... Feel free to make the change @wdhawkins if you want to...

Moreover the normal and easy instructions seem to converge to each other, which is a good thing, since it means we would have just one set of instructions to maintain. At some point, we should also mention that Lua can be installed via a package manager - Ubuntu 20.04 comes with 5.3 which is good enough. And probably get rid of `configure_dependencies.py` script. Then, the instructions would be:
- you need these packages
- if your package manager have those, install them
- the stuff that is not available via package manager has to be build and this is how.
- then build OpenSn

But this would be something to do in the future. Does not have to be done in this PR.

doc/install_linux.md
wdhawkins: Fixed.

doc/install_linux_easy.md
wdhawkins: Fixed.

doc/install_linux.md
wdhawkins: Fixed.

resources/configure_dependencies.py
wdhawkins: Fixed.

wdhawkins:
> General comment on the use of `CMAKE_PREFIX_PATH`. Previously, the dependencies like Lua, PETSc and VTK were installed into 3 different locations and users had to set 3 env variables. This is quite silly, since all the packages we are using are designed such that they can be installed into single location, e.g. `/usr/local` but anything really. Note that the new `configure_dependencies.py` script installs all packages into a single location and then tells the user to use `export CMAKE_PREFIX_PATH=/location/with/deps`.
> 
> The `install_linux.md` instructions tell users to install into 2 locations, one for PETSc and one for VTK. They should really install into a single location and then set the CMAKE_PREFIX_PATH into that. It is simpler... Feel free to make the change @wdhawkins if you want to...

Yes, I agree. I've fixed this.

andrsd (approved):

Issue #197, opened 2024-04-01
Improve support for command-line options


wdhawkins:
Our current support for command-line options is ad-hoc and a bit fragile. We should at the very least be using `getopt_long` and preferably something like CLI11.

andrsd:
I have been using [cxxopts](https://github.com/jarro2783/cxxopts). It is basically just one header file that would go into `external/cxxopts`. 

I was using [TCLAP](https://tclap.sourceforge.net/) before, but I like `cxxopts` a little bit better - mostly because it is just a single header file.

andrsd:
I agree that the command line parsing code should be handled by a modern library.

wdhawkins:
I'll try to put in a couple hours this week... it should be pretty easy to clean up.

wdhawkins:
Closed with #204.

PR #204, opened 2024-04-05
Using cxxopts package for cli.

wdhawkins:
This PR refactors our CLI processing to use cxxopts. It should be much more robust and easier to maintain. I've divided the options into user-facing and a hidden developer-facing.  For example, `opensn -h` or `opensn --help` will display the user-facing options. `opensn --help-dev` will display developer-centric options.

This PR addresses #197.

### DIFF
--- a/lua/framework/lua_app.cc
+++ b/lua/framework/lua_app.cc
@@ -95,8 +95,7 @@ LuaApp::ProcessArguments(int argc, char** argv)
     options.add_options("User")
     ("h,help",                      "Help message")
     ("c,suppress-color",            "Suppress color output")
-    ("v,verbose",                   "Verbosity level (0 to 3). Default is 0.",
-      cxxopts::value<int>()->implicit_value("0"))
+    ("v,verbose",                   "Verbosity level (0 to 3). Default is 0.", cxxopts::value<int>())
     ("caliper",                     "Enable Caliper reporting",
       cxxopts::value<std::string>()->implicit_value("runtime-report(calc.inclusive=true),max_column_width=80"))
     ("positional",                  "Positional arugments", cxxopts::value<std::vector<std::string>>());
--- a/lua/framework/lua_app.cc
+++ b/lua/framework/lua_app.cc
@@ -20,8 +20,7 @@ namespace opensnlua
 {
 
 LuaApp::LuaApp(const mpi::Communicator& comm)
-  : sim_option_interactive_(true),
-    allow_petsc_error_handler_(false)
+  : sim_option_interactive_(true), allow_petsc_error_handler_(false)
 {
   opensn::mpi_comm = comm;
 }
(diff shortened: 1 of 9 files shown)


lua/framework/lua_app.cc
```diff
@@ -53,115 +40,118 @@ LuaApp::InitPetSc(int argc, char** argv)
...
-void
+int
 LuaApp::ParseArguments(int argc, char** argv)
 {
-  bool input_file_found = false;
-  for (int i = 1; i < argc; i++)
+  try
   {
-    std::string argument(argv[i]);
+    cxxopts::Options options("opensn", "");
```
andrsd, line 85: Ideally, this should be `LowerCase(opensn::name)`, where `LowerCase` is a function that converts a string into a lowercase string. I don't think there is one in OpenSn. And shame on C++ that it still does not have it (!). 

I usually have one like this:
```
std::string
LowerCase(const std::string & name)
{
    std::string lower(name);
    std::transform(lower.begin(), lower.end(), lower.begin(), ::tolower);
    return lower;
}
```

This could be used later on for case-insensitive string comparisons. I am saying this to prevent hard-coding the transformation in place...

lua/framework/lua_app.cc
```diff
@@ -53,115 +40,118 @@ LuaApp::InitPetSc(int argc, char** argv)
...
+                << " finished execution." << std::endl;
+    }
+  }
 
   return error_code;
 }
 
-void
+int
 LuaApp::ParseArguments(int argc, char** argv)
```
andrsd, line 81: I think I want to rename (read I want you to rename) this member function to `ProcessArguments`, because that's really what is happening. This function parses the arguments and processes them. Parsing to me means that the arguments are checked syntactically which is what `options.parse(argc, argv);` does.

lua/framework/lua_app.h
```diff
@@ -12,30 +12,24 @@ class LuaApp
 {
 public:
   LuaApp(const mpi::Communicator& comm);
-  ~LuaApp();
+
+  ~LuaApp()
+  {}
```
andrsd, line 17: Probably event delete this dtor.

andrsd (changes requested):
A few small changes.

Really like how using `cxxopts` cleaned up the processing of command line arguments 👍

lua/framework/lua_app.cc
wdhawkins: Agreed. Done.

lua/framework/lua_app.cc
wdhawkins: Fixed.

lua/framework/lua_app.cc
```diff
@@ -1,47 +1,34 @@
...
-  "\n"
-  "     -v                          Level of verbosity. Default 0.\n"
-  "                                 Can be either 0, 1 or 2.\n"
-  "     a=b                         Executes argument as a lua string. "
-  "i.e. x=2 or y=[[\"string\"]]\n"
-  "     --allow-petsc-error-handler Allows petsc error handler.\n"
-  "     --suppress-color            Suppresses the printing of color.\n"
-  "                                 useful for unit tests requiring a diff.\n"
-  "     --dump-object-registry      Dumps the object registry.\n"
-  "\n\n\n";
```
ragusa, line 30: so no more help being displayed ?? :-(

ragusa (approved):

lua/framework/lua_app.cc
wdhawkins: There is... it's just in the ProcessArguments() routine. It shortened it a bit, but I'll go back and make it a bit more expansive.

wdhawkins: I expanded the help a bit. I can also add to the description of the positional arguments if necessary.

andrsd (approved):

PR #207, opened 2024-04-09
gha: adding caliper into gh-pages workflow dependencies

andrsd:
`github-pages` workflow started to fail after #196 was merged. We don't test the workflow on PRs, so that slipped through the cracks.


wdhawkins (approved):

ragusa (approved):

andrsd:
The workflow is now [fixed](https://github.com/Open-Sn/opensn/actions/runs/8618795790) and we are generating gh-pages again. Thanks for the quick approvals!

PR #208, opened 2024-04-09
Adding SPDX tags with licensing information

andrsd:
- gha: adding caliper into gh-pages workflow dependencies
- Adding SPDX tags into source files
- gha: updating QA workflow to check for missing SPDX tags in source files


andrsd:
After this PR is merged you will need to include the SPDX tags into every new .cc and .h file. It is just 2 lines:
```
// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
```

The year should be the year when the file was created. So, for the rest of the year you can just copy and paste this.

There is a check in place that will make sure files are not missing the tags. It is called "qa / check spdx tags". So, if you see that failing, it means you forgot to include the above notice. The check will tell you which files are missing the tags, if you go look at the details/log.

wdhawkins (approved):

ragusa (approved):

PR #212, opened 2024-04-12
Improving error handling of invalid Lua arguments

andrsd:
- Refactoring lua wrapper function test
- Don't wait 2 seconds on startup in debug mode
- Better error handling of invalid Lua arguments
- Adding tests for testing invalid lua arguments


andrsd:
I created an `InvalidLuaArgument` exception to handle the invalid arguments. You can see that it is pretty slick to do it this way. This is just a beginning for the system of exceptions we are going to have in future. So, this is not a one-off thing. So far there is nothing like this in the code and all is done via `std` exceptions. I am envisioning a base `opensn::Exception` and then derived exceptions like the `InvalidLuaArgument` one.

andrsd:
I sneaked in 2 more patches:
1. I renamed the `example` test so that its name resembles what it is testing, `example` was misleading.
2. I removed the 2 second wait in the debug mode which had no purpose, really.

wdhawkins (approved):

lua/framework/lua_app.cc
```diff
@@ -222,17 +222,6 @@ LuaApp::RunInteractive(int argc, char** argv)
...
-  opensn::log.Log() << "Waiting...";
-  if (opensn::mpi_comm.rank() == 0)
-    for (int k = 0; k < 2; ++k)
-    {
-      usleep(1000000);
-      opensn::log.Log() << k;
-    }
-  mpi_comm.barrier();
-#endif
-
```
ragusa, line 235: could this be due to the fact that it takes a bit of time for the the debugger (totalview, for instance) to start in parallel?

ragusa (approved):

lua/framework/lua_app.cc
andrsd: I typically do:
```
mpirun -np <n> ./binary params -- -start_in_debugger
```
the `-start_in_debugger` is a PETSc option that will print the commands how to attach a debugger to the new processes. Then, I open <n> terminals, paste these commands there and run.

But, I think I want to keep this:
```
#ifndef NDEBUG
  mpi_comm.barrier();
#endif
```
so that the processes waiting until all are started.

andrsd: I checked and the barrier is not needed, PETSc does the right thing...

PR #210, opened 2024-04-11
Initial Support for OpenMC MGXS Libraries

wdhawkins:
I think this is ready for merging. I might add an additional regression test or two.

This commit supports OpenMC libraries with sigt, siga, transfer, and simple fission. It does not support delayed/prompt and other more complicated processes.  These are very easy to add.

I would like to move to a header-only HDF library like h5pp. I haven't quite gotten it to behave like I want, so I've added a simple wrapper for some HDF functions for the time being. We are likely to need this functionality in a couple of weeks, and I don't want to hold things up any longer. 

It has had limited testing, but appears to be working correctly. For simple k-eignvalue problems, I match k given by OpenMC reasonably well. 

I have a Jupyter notebook that creates the OpenMC MGXS libraries. I'll add it to a tools directory in a future commit. I'm not an MC (and definitely not an OpenMC) person, so it likely needs careful inspection.

Issue #111, opened 2024-02-06
Rework of XS Classes


zhardy-lanl:
There are a lot of things we could do with the cross sections class... Here are some things to discuss:

- [x] Use an overloaded `InitOpenSnXS` routine to replace the various `Make<...>` routines. Completed in #214.
- [ ] Hook up the classes to input parameters, make the `InitOpenSnXS` routines protected, and call the appropriate one based on parameters. This could easily be ported over to constructors once we move away from input parameters (or left as init routines)
- [x] Naming: The name `SingleStateXS` originally came from the idea of having `MultiStateXS` which would essentially do interpolations on `SingleStateXS`. Seeing as that we have an `AdjointMGXS` and not a `SingleStateAdjointMGXS`, we are a bit inconsistent. I would favor getting rid of the inheritance structure for cross sections and having a single `MultiGroupXS` class where we do adjoint transposes in house (tracking with a flag).  Completed in #214.
 
Refs #109

ragusa:
i believe the second item was done when binding to python

PR #214, opened 2024-04-14
Refactoring XS classes

wdhawkins:
This PR refactors our xs classes and addresses the majority of issues identified in #111. 

`SingleStateXS` and `AdjointMGXS` have been deprecated in favor of a single `MGXS` class that handles all xs functionality. The various `Make...` routines have been refactored into an overloaded `Initialize` routine. The massive single xs source file has been broken up to facilitate maintenance. As Zach mentions in #111, it would be preferable to move from initialization routines to constructors, but this breaks our add/set paradigm for properties. It's possible to work around this problem, but it's messy.  I would prefer getting rid of add/set in favor of add with initialization.  

We still need to develop a long-term strategy for handling multi-temperature and interpolation, but I think this is a step in the right direction. 

I will rebase #210 on top of this if/when it's committed.


wdhawkins:
Do we need to keep the MC functionality?

framework/physics/physics_material/mgxs/mgxs_combined.cc
```diff
@@ -0,0 +1,193 @@
...
+      } // for j
+
+      precursor_count += xsecs[x]->NumPrecursors();
+    }
+
+    // Set inverse velocity data
+    if (x == 0 and xsecs[x]->InverseVelocity().empty())
+      inv_velocity_ = xsecs[x]->InverseVelocity();
+    OpenSnLogicalErrorIf(
+      xsecs[x]->InverseVelocity() != inv_velocity_,
```
ragusa, line 163: are we looking for equality between two reals? no tiny fuzzy = ?

framework/physics/physics_material/mgxs/mgxs_read_opensn.cc
```diff
@@ -798,15 +516,15 @@ SingleStateMGXS::MakeFromOpenSnXSFile(const std::string& file_name)
...
+      throw std::runtime_error("Error reading Chi cross section file "
                                "\"" +
                                file_name + "\".\n" + "Line number " + std::to_string(line_number) +
                                "\n" + err.what());
     }
 
     catch (const std::logic_error& err)
     {
-      throw std::logic_error("Error reading OpenSn cross section file "
+      throw std::logic_error("Error reading Chi cross section file "
```
ragusa, line 527: github's diff shows that the old file had "OpenSn cross section" and that the new version of the file has "Chi cross section". Weird.

framework/physics/physics_material/mgxs/mgxs_simple.cc
```diff
@@ -0,0 +1,70 @@
...
+  // Set the up/down-scattering cross sections.
+  // Summary:
+  //     1) The half of groups with higher energies down-scatter to the next
+  //        lowest energy group half the time and to the same group half the
+  //        time.
+  //     2) The half of groups with lower energies less the last group
+  //        down-scatter to the next lowest energy group three quarters of the
+  //        time and up-scatter to the next highest energy group one quarter
+  //        of the time.
+  //     3) The lowest energy group has the same form as 1).
```
ragusa, line 45: why does not the last group upscatter? Strange. Also, how can that group downscatter? It's the last group!!

framework/physics/physics_material/mgxs/mgxs_simple.cc
```diff
@@ -0,0 +1,70 @@
...
+    // Up-scattering
+    if (g > num_groups_ / 2)
+    {
+      if (g < num_groups_ - 1)
+      {
+        S.Insert(g, g - 1, sigma_t * c * 0.25);
+        S.Insert(g, g + 1, sigma_t * c * 0.25);
+      }
+      else
+        S.Insert(g, g - 1, sigma_t * c * 0.5);
```
ragusa, line 62: this line. see previous comment

lua/framework/physics/lua_material.cc
```diff
@@ -282,8 +282,7 @@ MatSetProperty(lua_State* L)
     // If the property is valid
     if (location_of_prop >= 0)
     {
-      auto prop =
-        std::static_pointer_cast<SingleStateMGXS>(cur_material->properties_[location_of_prop]);
+      auto prop = std::static_pointer_cast<MGXS>(cur_material->properties_[location_of_prop]);
 
       // Process operation
       if (operation_index == static_cast<int>(OpType::SIMPLEXS0))
```
ragusa, line 288: the name SIMPLEXS0 is cryptic. Better name ?

lua/framework/physics/lua_material.cc
```diff
@@ -294,7 +293,7 @@ MatSetProperty(lua_State* L)
         auto G = LuaArg<int>(L, 4);
         auto sigma_t = LuaArg<double>(L, 5);
 
-        prop->MakeSimple0(G, sigma_t);
+        prop->Initialize(G, sigma_t);
       }
       else if (operation_index == static_cast<int>(OpType::SIMPLEXS1))
```
ragusa, line 298: the name SIMPLEXS1 is cryptic. Better name ?

lua/framework/physics/lua_material.cc
```diff
@@ -314,7 +313,7 @@ MatSetProperty(lua_State* L)
 
         const auto file_name = LuaArg<std::string>(L, 4);
 
-        prop->MakeFromOpenSnXSFile(file_name);
+        prop->Initialize(file_name);
       }
       else if (operation_index == static_cast<int>(OpType::EXISTING))
```
ragusa, line 318: still confused by exisiting? Does this mean that a XS already exists in memory?

framework/physics/physics_material/mgxs/mgxs.cc
```diff
@@ -0,0 +1,228 @@
...
+            sigma_1 += vals[t];
+            break;
+          }
+      } // for gp
+    }   // if moment 1 available
+
+    // Compute transport cross section
+    if (sigma_1 >= sigma_t_[g])
+    {
+      log.Log0Warning() << "Transport cross section found to be less than zero for "
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

framework/physics/physics_material/mgxs/mgxs.cc
```diff
@@ -0,0 +1,228 @@
...
+    const auto& S_ell = transfer_matrices_[ell];
+    SparseMatrix S_ell_transpose(num_groups_, num_groups_);
+    for (size_t g = 0; g < num_groups_; ++g)
+    {
+      const size_t row_len = S_ell.rowI_indices_[g].size();
+      const size_t* col_ptr = S_ell.rowI_indices_[g].data();
+      const double* val_ptr = S_ell.rowI_values_[g].data();
+
+      for (size_t j = 0; j < row_len; ++j)
+        S_ell_transpose.Insert(*col_ptr++, g, *val_ptr++);
```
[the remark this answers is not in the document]
wdhawkins: Good question. I'm not familiar with the internals of SparseMatrix, but, at a quick glance, it looks like it's using `push_back` internally.

framework/physics/physics_material/mgxs/mgxs.cc
```diff
@@ -0,0 +1,228 @@
...
+  // Transpose production matrices
+  if (is_fissionable_)
+  {
+    transposed_production_matrix_.clear();
+    transposed_production_matrix_.resize(num_groups_);
+    const auto& F = production_matrix_;
+    for (size_t g = 0; g < num_groups_; ++g)
+      for (size_t gp = 0; gp < num_groups_; ++gp)
+        transposed_production_matrix_[g].push_back(F[gp][g]);
+  }
```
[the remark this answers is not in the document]
wdhawkins: In this case, the `production_matrix` (and the transpose) are `std::vector<std::vector<double>>` instead of `std::vector<SparseMatrix>`.

framework/physics/physics_material/mgxs/mgxs_combined.cc
```diff
@@ -0,0 +1,193 @@
...
+    {
+      is_fissionable_ = true;
+      Nf_total += xs->ScalingFactor();
+    }
+
+    // Define and check number of groups
+    if (xsecs.size() == 1)
+      n_grps = xs->NumGroups();
+    OpenSnLogicalErrorIf(xs->NumGroups() != n_grps,
+                         "All cross sections being combined must have the same group structure.");
```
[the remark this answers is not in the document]
wdhawkins: Yes. We should improve this check. Although, I'm really of the opinion we should deprecate this capability. It seems very fragile... and has probably never been used.

framework/physics/physics_material/mgxs/mgxs_combined.cc
wdhawkins: Yep, it certainly looks like it. Not very robust. I'm really beginning to wonder if this code has ever been used.

framework/physics/physics_material/mgxs/mgxs_read_opensn.cc
wdhawkins: That's weird. I'll double check that.

lua/framework/physics/lua_material.cc
wdhawkins: Yes, existing means a xs data that's already on the stack.

lua/framework/physics/lua_material.cc
wdhawkins: I think `SIMPLEXS0` is just a pure absorber. We could change it to PURE_ABSORBER.

framework/physics/physics_material/mgxs/mgxs_simple.cc
wdhawkins: I don't think the description is accurate.  It looks to me like that line should be:
`S.Insert(g, g + 1, sigma_t * c* 0.5);`

framework/physics/physics_material/mgxs/mgxs_combined.cc
```diff
@@ -0,0 +1,193 @@
...
+  // Pickup all xs and make sure they are valid
+  std::vector<std::shared_ptr<MGXS>> xsecs;
+  xsecs.reserve(combinations.size());
+
+  unsigned int n_grps = 0;
+  unsigned int n_precs = 0;
+  double Nf_total = 0.0; // Total density of fissile materials
+
+  // Loop over cross sections
+  for (auto combo : combinations)
```
andrsd, line 24: may be `auto & combo` so it does not do a local copy, i.e. forgotten `&`...

framework/physics/physics_material/mgxs/mgxs_write_opensn.cc
```diff
@@ -26,11 +23,13 @@ MultiGroupXS::ExportToOpenSnXSFile(const std::string& file_name,
     if (min_value >= 0.0)
     {
       for (auto val : xs)
+      {
```
andrsd, line 26: Does the `for` above missing `&`?

andrsd (changes requested):

lua/framework/physics/lua_material.cc
wdhawkins: Changed to PURE_ABSORBER.

framework/physics/physics_material/mgxs/mgxs_combined.cc
wdhawkins: Yep! Fixed.

framework/physics/physics_material/mgxs/mgxs_write_opensn.cc
wdhawkins: Fixed.

zhardy-lanl:
> Do we need to keep the MC functionality?

I don't think we do. It isn't used anywhere to my knowledge.

ragusa:
regarding the MC part, can someone point out to the lines of code? I can't find them. Thanks!

wdhawkins:
> regarding the MC part, can someone point out to the lines of code? I can't find them. Thanks!

Lines 171-173 in `mgxs.h`.

zhardy-lanl:
> regarding the MC part, can someone point out to the lines of code? I can't find them. Thanks!

@ragusa Checkout line 171 of `mgxs.h`.

ragusa:
Got it. I couldn't find a use from `scat_angles_gprime_g_`. I think we can remove

framework/physics/physics_material/mgxs/mgxs_simple.cc
```diff
@@ -0,0 +1,70 @@
...
+  Reset();
+
+  num_groups_ = num_groups;
+  sigma_t_.resize(num_groups, sigma_t);
+  sigma_a_.resize(num_groups, sigma_t);
+  ComputeDiffusionParameters();
+}
+
+void
+MGXS::Initialize(unsigned int num_groups, double sigma_t, double c)
```
wdhawkins, line 21: This entire routine will be replaced with a new version.

framework/physics/physics_material/mgxs/mgxs_combined.cc
```diff
@@ -0,0 +1,193 @@
...
+        for (size_t gp = 0; gp < num_groups_; ++gp)
+          production_matrix_[g][gp] += F[g][gp];
+
+        if (n_precs > 0)
+        {
+          nu_prompt_sigma_f_[g] += nu_p_sig_f[g];
+          nu_delayed_sigma_f_[g] += nu_d_sig_f[g];
+        }
+      }
+    } // for g
```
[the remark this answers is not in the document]
wdhawkins: Resolving this in favor of my comment on replacing the entire `Initialize(unsigned int num_groups, double sigma_t, double c)` routine.

framework/physics/physics_material/mgxs/mgxs_simple.cc
wdhawkins: Resolving this in favor of my comment on replacing the entire Initialize(unsigned int num_groups, double sigma_t, double c) routine.

framework/physics/physics_material/mgxs/mgxs_simple.cc
wdhawkins: Resolving this in favor of my comment on replacing the entire Initialize(unsigned int num_groups, double sigma_t, double c) routine.

zhardy-lanl, line 45:
```suggestion
  // Summary:
  //     1) The half of groups with higher energies down-scatters to the next
  //        group half the time and to the same group half of the time.
  //     2) The half of groups with lower energies excluding the last group
  //        down-scatters to the next group one quarter of the
  //        time, up-scatters to the previous group one quarter of the time,
  //        and to the same group half of the time.
  //     3) The lowest energy group up-scatters to the previous group half
  //        of the time and to the same group half of the time.
```

framework/physics/physics_material/mgxs/mgxs_simple.cc
zhardy-lanl: See suggested change.

framework/physics/physics_material/mgxs/mgxs_write_opensn.cc
zhardy-lanl, line 1: Should this be in `mgxs.h`?

framework/physics/physics_namespace.h
```diff
@@ -13,7 +13,7 @@ enum class OperationType
 {
   SINGLE_VALUE = 0,
   FROM_ARRAY = 1,
-  SIMPLEXS0 = 20,
+  PURE_ABSORBER = 20,
   SIMPLEXS1 = 21,
```
zhardy-lanl, line 17: The name here is a bit tricky because the XS are not in a standard form that has a common name. It may be good to add a `DOWN_SCATTERER` option and call this `FULL_SCATTERER`, `UPDOWN_SCATTERER`, or `MIXED_SCATTERER` or something like that.

lua/framework/physics/lua_material.h
```diff
@@ -106,9 +106,9 @@ int MatAddProperty(lua_State* L);
  * will be converted to 0 to N-1. This method can be used to set mutligroup
  * cross-sections or sources.
  * \n
- * SIMPLEXS0\n
- * Makes a simple material with no transfer matrix just \f$\sigma_t \f$. Expects
- * two values: \n
+ * PURE_ABSORBER\n
+ * Makes a simple material with no transfer matrix and \f$\sigma_t \f$ =
+ * \f$\sigma_a \f$. Expects two values: \n
```
zhardy-lanl, line 111:
```suggestion
 * Makes a simple material with no transfer matrix and \f$\sigma_t =
 * \sigma_a \f$. Expects two values: \n
```
Any reason for the `=` to not be in math font?

lua/framework/physics/physics_material/mgxs/mgxs_lua_utils.h
```diff
@@ -45,9 +45,9 @@ int PhysicsTransportXSCreate(lua_State* L);
  * will be converted to 0 to N-1. This method can be used to set mutli-group
  * cross sections or sources.
  * \n
- * SIMPLEXS0\n
- * Makes a simple material with no transfer matrix just \f$\sigma_t \f$. Expects
- * two values: \n
+ * PURE_ABSORBER\n
+ * Makes a simple material with no transfer matrix and \f$\sigma_t \f$ = \f$\sigma_a \f$.
```
zhardy-lanl, line 49:
```suggestion
 * Makes a simple material with no transfer matrix and \f$\sigma_t = \sigma_a \f$.
```
See above.

zhardy-lanl (changes requested):
Several small items and comments on previous discussions. On a larger note, why are we moving to a completely abbreviated class name in `MGXS` from `MultiGroupXS`? I much prefer the latter. Further, shouldn't we consolidate all of the class routines into a single source file? The end product would then be `.../physics_material/m(ulti_)g(roup_)xs.h(cc)`.

There is a lot that can and should be done in the near future with the XS to support linking to XS libraries and having variable densities and things of that sort. This is a good step in that direction rather than having several different XS flavors.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
zhardy-lanl: This is the combine routine though?

lua/framework/physics/lua_material.cc
zhardy-lanl: From my review:

> The name here is a bit tricky because the XS are not in a standard form that has a common name. It may be good to add a DOWN_SCATTERER option and call this FULL_SCATTERER, UPDOWN_SCATTERER, or MIXED_SCATTERER or something like that.

framework/physics/physics_material/mgxs/mgxs_simple.cc
zhardy-lanl: For my own edification, what is the change going to be?

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: I meant deprecate the combine routine, but I see your point about combining lists of isotopes. I won't tackle improving this check in this PR, but I do still think it's needs work since all it's checking is the number of groups in each xs object and not the actual group structure itself.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: Yes... sorry, I must have selected the wrong piece of code. Jean and I discussed this, and essentially came up with the same solution as you. Add an additional parameter to this `Initialize` routine that would all the user to select pure downscatter, or some kind of mix of downscatter and upscatter.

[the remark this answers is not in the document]
wdhawkins: If you or Jean want to send me a routine, I'm happy to include it here.  I suspect you two know better what you would like to see here. 😄

framework/physics/physics_material/mgxs/mgxs.h
```diff
@@ -0,0 +1,184 @@
...
+
+#pragma once
+
+#include "framework/physics/physics_material/material_property_base.h"
+#include "framework/math/sparse_matrix/math_sparse_matrix.h"
+
+namespace opensn
+{
+
+class MGXS : public PhysicsMaterialProperty
```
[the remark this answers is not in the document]
wdhawkins: I'm ok with this. I'll make this change. I don't really have strong feelings between MGXS and MultiGroupXS.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
zhardy-lanl: Totally agree. Right now the only checks are on number of groups and velocities (which are only required for transients which we don't have). We do have the `e_bounds` attribute (although it should probably be made into a `std::vector<double>` of length `num_groups + 1` instead of a `std::vector<std::vector<double>>`.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: I purposely split these routine up into individual source files to try to prevent having one source with 1000+ lines. I'm ok combining some of the routines, but I'd still like to try to keep the size of each of the source files to something manageable. The `Initialize` routine for reading OpenSn native files is ~1000 lines itself, and there will soon be a routine to read OpenMC libraries that will likely be quite large. We could put the smaller `Initialize` routines into `multi_group_xs.cc`.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
zhardy-lanl: Once I get caught up with the new dependencies, I may try to knock one out. I couldn't give you a timeline right now though.

Here are a few thoughts:

- For multi-group, the pure absorber and this XS type aren't physical (same `sigma_t` and/or `c` for all groups). We could enforce that these types are for one group XS only and require an input file for multi-group problems. That would get rid of the problem of the made up up/down-scattering ratios.
- We could change the inputs for the pure absorber and this XS type to take a `std::vector<double>` for `sigma_t` and `c`. This still begs the question about up/down-scattering. For this, I would say that we should add 2 inputs: `std::vector<double> downscatter_fraction` and `std::vector<double> upscatter_fraction`. This then defines the within-group fraction as well (with checks for the sum being less than 1.0 per group). The Lua API may be a bit tricky if you want to keep the same input signatures where they may already exist, but it's doable. If the Lua input is a number and not a table, you would assume the same behavior as the existing routines and you would have to create the vector inputs in the Lua layer.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
zhardy-lanl: I figured that was the case and I am definitely agreeable to keeping source files manageable. The reason I bring this up is because this is what was we did in Chi-Tech, and we subsequently abandoned that when starting OpenSn. For example `lbs_solver.cc` is 3.5k lines of code. What are your thoughts on this @andrsd ?

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: I don't want to introduce too many more changes to the internals in this PR, but I hated `e_bounds` so much that I went ahead and fixed it as you suggested. It is not currently used, so it won't affect anything downstream of the mgxs class. In an OpenSn xs file, what we now expect is a list of `num_groups + 1` boundaries:

```
GROUP_STRUCTURE_BEGIN
...
...
...
GROUP_STRUCTURE_END
```

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: I've consolidated the bulk of the code into `multi_group_xs.cc` with the exception of reading and exporting OpenSn files. This keeps`multi_group_xs.cc` from being too terribly large. I expect the end result will be the following:

```
multi_group_xs.h
multi_group_xs.cc
import_opensn.cc
export_opensn.cc
import_openmc.cc
```
When #210 is merged, the `Initialize` routine that currently reads OpenSn files will move into `multi_group_xs.cc`. It will be a short routine that looks to see whether the supplied file is an OpenSn file or an OpenMC file and then calls the appropriate routine, so I think the structure above makes good sense.

Does this seem like a good compromise?

framework/physics/physics_material/mgxs/mgxs_combined.cc
```diff
@@ -0,0 +1,193 @@
...
+    // Combine cross sections
+    const auto& sig_t = xsecs[x]->SigmaTotal();
+    const auto& sig_a = xsecs[x]->SigmaAbsorption();
+    const auto& sig_f = xsecs[x]->SigmaFission();
+    const auto& nu_p_sig_f = xsecs[x]->NuPromptSigmaF();
+    const auto& nu_d_sig_f = xsecs[x]->NuDelayedSigmaF();
+    const auto& F = xsecs[x]->ProductionMatrix();
+
+    // Here, raw cross sections are scaled by densities and spectra by
+    // fractional densities. The latter is done to preserve a unit spectra.
```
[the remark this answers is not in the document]
wdhawkins: Removed.

framework/physics/physics_material/mgxs/mgxs_write_opensn.cc
wdhawkins: See my comment above about my thoughts for the code layout.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
zhardy-lanl: I think that is fine, I would just like to get @andrsd take on deviating from what we are doing throughout the remainder of the codebase.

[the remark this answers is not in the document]
andrsd: You don't need to split the file here. The problem that needs to be solved is to refactor the parser/exporter of OpenSn XS file into its own layer, so you can do things like:
```
opensn::XSFile file;
file.open(...);
file.parser();
// pick bits from file and create your XS data structure
```

With the HDF5 reader for OpenMC, I am not sure how big the code gets. I would not think it would be terrible, since the h5pp can load the datasets into C++ data structures like `std::vector<>` and all error handling is done via exception, so you really just need to wrap the code into a try-catch block.

However, if these routines become too big, they can be moved into "factory"(*) classes that will be building the XS data structures, so instead of `Initialize`, we could do something like `opensn::XS xs = OpenMCFactory::FromFile(file_name)`. This is just a sketch of an idea what can be done.

(*) factory here is not the factory design pattern we use building `Object`s - hence the `"`.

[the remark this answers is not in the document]
wdhawkins: @andrsd 100% agree with refactoring the OpenSn XS  from-file functionality into it's own layer. That would certainly be the way to go.  Unfortunately, it's unlikely to be some time before we get there given the number of things that need to be done.  In the meantime, I still think we should try to avoid multi-thousand line source files.

lua/framework/physics/physics_material/mgxs/mgxs_lua_utils.h
```diff
@@ -45,9 +45,9 @@ int PhysicsTransportXSCreate(lua_State* L);
  * will be converted to 0 to N-1. This method can be used to set mutli-group
```
ragusa, line 45: mutli --> multi

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: I've replaced the SIMPLEXS0/PURE_ABSORBER and SIMPLEXS1 routines with a single MODEL1G routine. For those test problems that needed the multi-group capability of SIMPLEXS1, I exported the necessary cross sections to an OpenSn XS file.

framework/physics/physics_material/mgxs/mgxs_simple.cc
wdhawkins: Resolved with MODEL1G.

framework/physics/physics_material/mgxs/mgxs_combined.cc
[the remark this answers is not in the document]
wdhawkins: I've merged all of the routines back into `multi_group_xs.cc` with the exception of the i/o routines which will become standalone classes.

lua/framework/physics/lua_material.cc
wdhawkins: Resolved with MODEL1G.

ragusa (approved):

andrsd (approved):

framework/physics/physics_namespace.h
```diff
@@ -13,8 +13,7 @@ enum class OperationType
 {
   SINGLE_VALUE = 0,
   FROM_ARRAY = 1,
-  SIMPLEXS0 = 20,
-  SIMPLEXS1 = 21,
+  MODEL1G = 20,
```
zhardy-lanl, line 16: How about something like
```suggestion
  SIMPLE_ONE_GROUP = 20,
```
or
```suggestion
  SIMPLE_1G = 20,
```
What does model mean here?

lua/framework/physics/lua_material.h
```diff
@@ -106,20 +106,9 @@ int MatAddProperty(lua_State* L);
...
- *
- * ####_
- *
- * SIMPLEXS1\n
- * Makes a simple material with isotropic transfer matrix (L=0)
- * and mostly down scattering but with a few of the last groups
- * subject to up-scattering. Expects three values
- * values: \n
- *  - int number of groups (\f$G \f$),
+ * MODEL1G\n
```
zhardy-lanl, line 109: See other comment.

lua/framework/physics/lua_material.h
```diff
@@ -106,20 +106,9 @@ int MatAddProperty(lua_State* L);
...
+ * Makes a simple, 1-group material with a user-selectable amount of scattering.
```
zhardy-lanl, line 110:
```suggestion
 * Makes a simple, one group material with a user-selectable scattering ratio.
```

lua/framework/physics/physics_material/multi_group_xs/multi_group_xs_lua_utils.h
```diff
@@ -42,23 +42,12 @@ int PhysicsTransportXSCreate(lua_State* L);
...
- *
- * ####_
- *
- * SIMPLEXS1\n
- * Makes a simple material with isotropic transfer matrix (L=0)
- * and mostly down scattering but with a few of the last groups
- * subject to up-scattering. Expects three values
- * values: \n
- *  - int number of groups (\f$G \f$),
+ * MODEL1G\n
```
zhardy-lanl, line 48: See other comment.

lua/framework/physics/physics_material/multi_group_xs/multi_group_xs_lua_utils.h
```diff
@@ -42,23 +42,12 @@ int PhysicsTransportXSCreate(lua_State* L);
...
+ * Makes a simple, 1-group material with a user-selectable amount of scattering.
```
zhardy-lanl, line 49:
```suggestion
 * Makes a simple, one group material with a user-selectable scattering ratio.
```

lua/framework/physics/physics_material/multi_group_xs/multi_group_xs_lua_utils.cc
```diff
@@ -23,8 +23,7 @@ RegisterLuaFunctionNamespace(PhysicsTransportXSExportToOpenSnFormat, xs, ExportT
 
 RegisterLuaConstantAsIs(SINGLE_VALUE, Varying(0));
 RegisterLuaConstantAsIs(FROM_ARRAY, Varying(1));
-RegisterLuaConstantAsIs(SIMPLEXS0, Varying(20));
-RegisterLuaConstantAsIs(SIMPLEXS1, Varying(21));
+RegisterLuaConstantAsIs(MODEL1G, Varying(20));
```
zhardy-lanl, line 26: See other comment.

zhardy-lanl (changes requested):
Not to hold things up, but I'm not sure I'm a big fan of `MODEL1G` for a new name. Please see my alternate suggestions in the comments. Since the documentation calls the cross sections simple, one group cross sections I think the name should follow suit.

lua/framework/physics/lua_material.h
ragusa: i'm good with this. I was say: one-group material (needs a hyphen)

lua/framework/physics/physics_material/multi_group_xs/multi_group_xs_lua_utils.h
ragusa: please add hyphen --> one-group

lua/framework/physics/lua_material.h
zhardy-lanl: Fixed the suggested change.

lua/framework/physics/physics_material/multi_group_xs/multi_group_xs_lua_utils.h
zhardy-lanl: Fixed the suggested change.

framework/physics/physics_namespace.h
ragusa: "simple" is not that important. I would say that the isotropic nature of scattering allowed in that simple model is a more important information to convey. How about `ISOTROPIC_ONE_GROUP` or `ISOTROPIC_1G` ?

zhardy-lanl: I would vote no on `ISOTROPIC` because you can set this to be a pure absorber. To me `SIMPLE` is indicative of there only being absorption and scatter (i.e. no fission).

ragusa: ok

wdhawkins: I went with `MODEL` because it really is a model xs.  I'm ok with `SIMPLE` as well, so I've changed it to `SIMPLE_ONE_GROUP`.

lua/framework/physics/lua_material.h
wdhawkins: Fixed.

zhardy-lanl (approved):

PR #215, opened 2024-04-16
Removing positional arguments from CLI

### LINKED ISSUE #213: Command line arguments inconsistencies
There are currently 2 modes of operation: 

1. Interactive mode

   Users do:
   ```
   ./opensn [lua-arg]
   ```

2. Batch mode

   Users do:
   ```
   ./opensn file-name [lua-arg]
   ```

We treat `file-name` and `lua-arg` as positional arguments, which means they have to be supplied. But, as you can see, both are actually **optional** arguments:
- If `file-name` is specified, we run batch mode. If not, we run in interactive mode.
- `lua-arg`s are just checked if they include `=` and if so they are added to the lua processing buffer, but in general this can actually be any lua command.

## Proposed change

This can be changed to:
1. Using `-i` option for an input file, this would be setting the batch/interactive mode
2. Using `-l` (or whatever else) for the lua arguments - could be supplied multiple times, like so `-l var1=100 -l var2=102`

wdhawkins:
This PR removes all positional arguments from the CLI.  -i or --input is required for input files, and -l or --lua is required for Lua expressions. cxxopts allows specifying multiple options of the same type, so we only use the last option of each type found on the command line. We could probably flag that as an error if we want to.

Closes #213

### DIFF
--- a/lua/framework/lua_app.cc
+++ b/lua/framework/lua_app.cc
@@ -101,7 +101,9 @@ LuaApp::ProcessArguments(int argc, char** argv)
     ("v,verbose",                   "Verbosity level (0 to 3). Default is 0.", cxxopts::value<int>())
     ("caliper",                     "Enable Caliper reporting",
       cxxopts::value<std::string>()->implicit_value("runtime-report(calc.inclusive=true),max_column_width=80"))
-    ("positional",                  "Positional arugments", cxxopts::value<std::vector<std::string>>());
+    ("i,input",                     "Input file", cxxopts::value<std::string>())
+    ("l,lua",                       "Lua expression",
+      cxxopts::value<std::vector<std::string>>());
 
     options.add_options("Dev")
       ("help-dev",                  "Developer options help")
@@ -109,10 +111,6 @@ LuaApp::ProcessArguments(int argc, char** argv)
       ("dump-object-registry",      "Dump object registry");
     /* clang-format on */
 
-    options.positional_help("[FILE] [LUA ASSIGNMENT]");
-    options.parse_positional("positional");
-
-    bool found_filename = false;
     auto result = options.parse(argc, argv);
 
     if (result.count("help"))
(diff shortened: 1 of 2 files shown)


andrsd:
>cxxopts allows specifying multiple options of the same type, so we only use the last option of each type found on the command line. We could probably flag that as an error if we want to.

I don't think we want to flag that as an error. For example, compilers behave this way - they honor the last specified flag. If somebody does `-march=haswell -march=native`, native is used. No error thrown. And it is not only compilers...

lua/framework/lua_app.cc
```diff
@@ -161,25 +158,17 @@ LuaApp::ProcessArguments(int argc, char** argv)
...
-            opensn::input_path = arg;
-            sim_option_interactive_ = false;
-            found_filename = true;
-          }
-          else
-            throw std::runtime_error("Invalid option " + arg);
-        }
-      }
+      auto path = result["input"].as<std::string>();
+      opensn::input_path = path;
```
andrsd, line 164: `opensn::input_path = result["input"].as<std::string>();`

lua/framework/lua_app.cc
```diff
@@ -161,25 +158,17 @@ LuaApp::ProcessArguments(int argc, char** argv)
...
+      sim_option_interactive_ = false;
+    }
+
+    if (result.count("lua"))
+    {
+      auto args = result["lua"].as<std::string>();
+      console.GetCommandBuffer().push_back(args);
```
andrsd, line 171: This can be then:
```
for (auto& l : result["lua"].as<std::vector<std::string>>())
    console.GetCommandBuffer().push_back(l);
```

test/src/test_slot.py
```diff
@@ -25,10 +25,10 @@ def _Run(self):
             return
 
         cmd = self.argv.mpi_cmd + " " + str(test.num_procs) + " "
-        cmd += self.argv.exe + " "
+        cmd += self.argv.exe + " -i "
         cmd += test.filename + " "
```
andrsd, line 29: This is cosmetic, but I would personally keep the flag and the value on the same line:
```
cmd += "-i " + test.filename + " "
```

Or even better (if you feel like it). Something like:
```
cmd = " ".join([ params ])
```

But maybe, we could change this in a different PR. Totally optional...

andrsd (changes requested):

lua/framework/lua_app.cc
```diff
@@ -101,18 +101,15 @@ LuaApp::ProcessArguments(int argc, char** argv)
     ("v,verbose",                   "Verbosity level (0 to 3). Default is 0.", cxxopts::value<int>())
     ("caliper",                     "Enable Caliper reporting",
       cxxopts::value<std::string>()->implicit_value("runtime-report(calc.inclusive=true),max_column_width=80"))
-    ("positional",                  "Positional arugments", cxxopts::value<std::vector<std::string>>());
+    ("i,input",                     "Input file", cxxopts::value<std::string>())
+    ("l,lua",                       "Lua expression", cxxopts::value<std::string>());
```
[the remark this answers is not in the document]
wdhawkins: Of course. I completely spaced on that one. Fixed.

test/src/test_slot.py
wdhawkins: I kept the flag and value on the same line. I'll defer the `join` another PR. There's quite a bit of cleanup that I think we could do there.

andrsd (approved):

ragusa (approved):

PR #219, opened 2024-04-17
Refactoring mesh reading capability

### LINKED ISSUE #217: Separate I/O capability from `UnpartitionedMesh`
`UnpartitionedMesh` is a data container. The various ways of creating it should be moved in its own layer. This layer would be creating `UnpartitionedMesh` objects. And this layer would be used by the `FromFileMeshGenerator`.

This new layer would contain code for:
- working with VTK, i.e. using VTK to read different file formats like `vtu`, `pvtu`, `ExodusII`, etc.
- working with `msh` file created by `gmsh`
- working with wavefront `obj` files

In future, new formats can be easily added into this layer. This will keep things organized and well separated.

andrsd:
- Refactoring mesh reading capability into an I/O layer
- Dimension is not a `MeshAttribute`
- Clean up how orthogonal attributes are done
- Splitting ComputeCentroids and CheckQuality into individual functions

Closes #217


framework/io/gmsh_io.cc
```diff
@@ -0,0 +1,412 @@
...
+{
+  const std::string fname = "GmshIO::ReadFromMsh";
+
+  // Opening the file
+  std::ifstream file;
+  file.open(options.file_name);
+  if (not file.is_open())
+  {
+    log.LogAllError() << "Failed to open file: " << options.file_name << " in call "
+                      << "to ReadFromMsh \n";
```
ragusa, line 23: do you want to keep ReadFromMsh as is in the error message?

framework/io/gmsh_io.cc
```diff
@@ -0,0 +1,412 @@
...
+  /**Lamda for checking if an element is 2D.*/
+  auto IsElementType2D = [](int element_type)
+  {
+    if (element_type == 2 or element_type == 3)
+      return true;
+
+    return false;
+  };
+
+  /**Lamda for checking if an element is 2D.*/
```
ragusa, line 120: update comment 2D -> 3D

framework/io/gmsh_io.cc
```diff
@@ -0,0 +1,412 @@
...
+      cell_type = CellType::TRIANGLE;
+    else if (element_type == 3)
+      cell_type = CellType::QUADRILATERAL;
+    else if (element_type == 4)
+      cell_type = CellType::TETRAHEDRON;
+    else if (element_type == 5)
+      cell_type = CellType::HEXAHEDRON;
+    else if (element_type == 6 or // Prism
+             element_type == 7)
+      cell_type = CellType::POLYHEDRON; // Pyramid
```
ragusa, line 155: should // Pyramid but moved up one line to be next to type 7 ?

framework/io/obj_io.cc
```diff
@@ -0,0 +1,392 @@
...
+        } // for face
+      }   // for edge
+
+      log.Log() << "UnpartitionedMesh: assigned " << num_faces_boundarified
+                << " faces to boundary id " << bndry_id << " with name " << block_data[bid].name;
+
+      mesh->AddBoundary(bndry_id, block_data[bid].name);
+
+      ++bndry_id;
+    } // for boundary block
```
ragusa, line 386: do we have a preference between `++a` and `a++`. should we specify a coding convention too?

framework/mesh/mesh_generator/extruder_mesh_generator.cc
```diff
@@ -101,17 +101,17 @@ ExtruderMeshGenerator::ExtruderMeshGenerator(const InputParameters& params)
   } // layer_block in layers_param
 }
 
-std::unique_ptr<UnpartitionedMesh>
-ExtruderMeshGenerator::GenerateUnpartitionedMesh(std::unique_ptr<UnpartitionedMesh> input_umesh)
+std::shared_ptr<UnpartitionedMesh>
```
ragusa, line 104: what are the consequences (or reasons) for going from a unique ptr to a shared ptr ?

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/sweep_chunks/lbs_curvilinear_sweep_chunk_pwl.cc
```diff
@@ -66,7 +66,7 @@ SweepChunkPwlrz::SweepChunkPwlrz(
       map_polar_level_.emplace(dir_idx, dir_set.first);
 
   //  set normal vector for symmetric boundary condition
-  const int d = (grid_.Attributes() & DIMENSION_1) ? 2 : 0;
+  const int d = (grid_.Dimension() == 1) ? 2 : 0;
   normal_vector_boundary_ = Vector3(0.0, 0.0, 0.0);
   normal_vector_boundary_(d) = 1;
```
ragusa, line 71: I do not fully follow this. When dim=1, th normal is [0,0,1], otherwise it is [1,0,0]. Why [1,0,0] for dim=2/3 ?

framework/io/gmsh_io.cc
andrsd: Strictly speaking those magic numbers should not be there at all. They should be an enum. But, I only moved the code. This routine will be replaced by [gmshparsercpp](https://github.com/andrsd/gmshparsercpp) which does the right thing with the enum, see https://github.com/andrsd/gmshparsercpp/blob/main/include/gmshparsercpp/Enums.h#L11

framework/io/obj_io.cc
andrsd: I don't think this should be in coding conventions. There are cases where you should be using pre-fix version and there cases where post-fix is just fine...

framework/mesh/mesh_generator/extruder_mesh_generator.cc
andrsd: `UnpartitionedMesh`es are stored as `shared_ptr`. When mesh generators were executed there were some `std::move`s so that the `shared_ptr` could be passed into the API as a `unique_ptr`. If I understand correctly `std::move` invalidates the source memory, i.e. the stored `shared_ptr` would no longer valid. But, the lua layer might be keeping a lua handle into that stack, so when it would want to access it again (i.e. after the mesh generators were executed) the expected `UnpartitionedMesh` would not be there. This was my line of thinking (but I did not test it).

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/sweep_chunks/lbs_curvilinear_sweep_chunk_pwl.cc
andrsd: I think it must have something to do with the fact that the 1D orthogonal mesh generator, creates the mesh in the z-coordinates https://github.com/Open-Sn/opensn/blob/0f4fee4410b6379fab86a4b9367d33d026a55241/framework/mesh/mesh_generator/orthogonal_mesh_generator.cc#L121-L125

We should change that so it is in x? If so, I can open an issue for that task.

framework/mesh/mesh_generator/extruder_mesh_generator.cc
ragusa: ok. tagging @wdhawkins and @zhardy-lanl for awareness and maybe feedback

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/sweep_chunks/lbs_curvilinear_sweep_chunk_pwl.cc
ragusa: this file is for the cylindrical coordinate system (given its name `lbs_curvilinear....`). I am not even sure what 1D should mean there, given what the lines 68-71 do. Tagging @zhardy-lanl  and @wdhawkins 
1. in 1D cylindrical, the coordinate system should be \vec{e_r} (the radial dimension). The normal should be [1,0,0] as I expect the radial coordinate to be the first one. What those lines above do is treat 1D cylindrical like 1D Cartesian and I agree that in 1D Cartesian, we have selected z to be the axis and not x. In cylindrical coordinates, the polar angle is really with respect to the cylindrical axis.  Implementation wise, I am still confused why 1D cyl coord has a normal as [0,0,1]. The comments states symmetric boundary condition, and indeed, the axis of the cylinder is a symmetric BC. 
2. In multi-D cylindrical, I *do* understand the choice of normal to be [1,0,0] as it fits what I described above. 
3. So, I must conclude that in the starting direction case of cylindrical, which is a 1D slab simulation, the code calls the 1D solver of Cartesian coordinates, which has z as its only axis hence the normal [0,0,1] and for the other sweeping directions, i.e., in true cylindrical sweeps, the normal is set to [1,0,0]. 
I will wait for others to comment. My explanation or understanding is too convoluted. I hope I am not right as this would be a hacky implementation

framework/io/obj_io.cc
wdhawkins: I think the safe thing to say in our coding standard is: "prefer prefix operator for non-pod types". 99.9% of the time it won't matter which you use because the compiler is going to see what's happening and optimize appropriately. But, there are situations where it can make a difference, and, for inexperienced programmers like we're likely to have on OpenSn, it's probably worth mentioning.

andrsd: To me, this is C++ 101 and thus it does not belong into coding standards.

framework/mesh/mesh_generator/extruder_mesh_generator.cc
wdhawkins: I would agree with this assessment. Was there a specific reason (that you can tell) that we go from `shared_ptr` to `unique_ptr` for the mesh generator? I haven't looked very closely at this bit of the code.

framework/mesh/mesh.h
```diff
@@ -70,6 +67,13 @@ operator|(const MeshAttributes f1, const MeshAttributes f2)
   return static_cast<MeshAttributes>(static_cast<int>(f1) | static_cast<int>(f2));
 }
 
+struct MeshOrthoAttributes
```
zhardy-lanl, line 70:
```suggestion
struct OrthoMeshAttributes
```

framework/mesh/mesh.h
```diff
@@ -56,9 +56,6 @@ class VolumeMesherPredefinedUnpartitioned;
 enum MeshAttributes : int
 {
   NONE = 0,
-  DIMENSION_1 = (1 << 0),
-  DIMENSION_2 = (1 << 1),
-  DIMENSION_3 = (1 << 2),
   ORTHOGONAL = (1 << 3),
   EXTRUDED = (1 << 4),
   UNSTRUCTURED = (1 << 5)
```
zhardy-lanl, line 61: To be honest, this whole enum is a bit weird. I'm not super familiar with how these flags work, but couldn't someone set the mesh attributes to both orthogonal and unstructured here? We should rework this and maybe have an enum class for mesh type as `ORTHOGONAL`/`UNSTRUCTURED` and a boolean attribute in whatever class for extruded.

framework/mesh/mesh.cc
```diff
@@ -28,21 +28,18 @@ CreateUnpartitioned1DOrthoMesh(std::vector<double>& vertices)
...
   size_t Nz = vertices.size();
 
-  umesh->GetMeshOptions().ortho_Nx = 1;
-  umesh->GetMeshOptions().ortho_Ny = 1;
-  umesh->GetMeshOptions().ortho_Nz = Nz - 1;
-  umesh->GetMeshOptions().boundary_id_map[4] = "ZMAX";
-  umesh->GetMeshOptions().boundary_id_map[5] = "ZMIN";
+  umesh->SetOrthoAttributes(1, 1, Nz - 1);
+  umesh->AddBoundary(4, "ZMAX");
+  umesh->AddBoundary(5, "ZMIN");
```
zhardy-lanl, line 38: Not suggesting to fix this now, but it has always pained me that max comes first in the integer ordering...

framework/mesh/mesh_generator/from_file_mesh_generator.cc
```diff
@@ -62,31 +58,26 @@ FromFileMeshGenerator::GenerateUnpartitionedMesh(std::unique_ptr<UnpartitionedMe
...
+    return VtkIO::FromExodus(options);
   else if (extension == ".vtu")
-    umesh->ReadFromVTU(options);
+    return VtkIO::FromVTU(options);
   else if (extension == ".pvtu")
-    umesh->ReadFromPVTU(options);
+    return VtkIO::FromPVTU(options);
   else if (extension == ".case")
-    umesh->ReadFromEnsightGold(options);
+    return VtkIO::FromEnsightGold(options);
```
zhardy-lanl, line 76: Would it make more sense to have `ObjIO::FromOBJ` and `GmshIO::FromGmsh` instead to be more consistent with routines in `VtkIO`?

zhardy-lanl (changes requested):
To me it really feels as though `framework/io` should be moved to `framework/mesh/io` as that everything in it has to do with meshing. What are your thoughts? Other than that, a few minor suggestions.

framework/mesh/mesh_generator/extruder_mesh_generator.cc
zhardy-lanl: If there was a reason for `unique_ptr` I don't know it. If it isn't obvious why it is used, there is no reason to not move to the less restrictive `shared_ptr` IMO.

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/sweep_chunks/lbs_curvilinear_sweep_chunk_pwl.cc
zhardy-lanl: If I am reading it correctly, `DiscreteOrdinatesCurvilinearSolver::PerformInputChecks()` has a check that only allows 2D cylindrical.

framework/mesh/mesh_generator/from_file_mesh_generator.cc
zhardy-lanl: Also, should we change `VtkIO` to `VTKIO` since VTK is an all caps acronym?

framework/mesh/mesh.h
ragusa: those bitwise operations. argh. I am sure they are useful but in what context? aren't those here equivalent to 1,2,4,8,16,32? why is bitwise better?

andrsd: 1. I agree that this enum is weird as hell. I will be looking at removing it in future. Right now, I just undid the dimension non-sense.
2. one big thing that I think was a problem is that this way you can easily have mesh that is 1D, 2D and 3D at the same time (3 lower bits sets)

framework/mesh/mesh_generator/from_file_mesh_generator.cc
andrsd: > Would it make more sense to have ObjIO::FromOBJ and GmshIO::FromGmsh instead to be more consistent with routines in VtkIO?

Or, I could do it the other way around and have `VtkIO::FromFile()` and do the split on the type inside 🤔 There is quite a bit of overlap inside the VtkIO class that I am not dealing with in the PR, but it is something that deserves attention.

framework/mesh/mesh_generator/extruder_mesh_generator.cc
andrsd: This could be the reason for choosing `unique_ptr`:

https://github.com/idaholab/moose/blob/0814fb339a029de034bb8b2de55749ef6d8b3426/framework/include/meshgenerators/MeshGenerator.h#L69

and

https://github.com/idaholab/moose/blob/0814fb339a029de034bb8b2de55749ef6d8b3426/framework/include/meshgenerators/MeshGenerator.h#L265

framework/mesh/mesh.cc
andrsd: I would personally change the boundary names to:
- left, right (1D)
- left, right, top, bottom (2D)
- left, right, top, bottom, front, back (3D)

Min/max has a different meaning to me. I don't associate it with boundary names...

framework/mesh/mesh.h
wdhawkins: I agree. I honestly don't see any good reason for us to ever use something like this.

wdhawkins:
I don't mind `framework/io`. As @andrsd says, all of the xs i/o should go there as well. I envision something similar in the solvers for restarts, moments, etc.

framework/mesh/mesh.h
zhardy-lanl: Bitwise operations are not in of themselves something to avoid but they should be used carefully. They are good when you want to specify a set of flags, which is probably why they were used her in the first place (i.e. dimension and orthogonal/unstructured are separate attributes). We use them for `SourceFlags` to specify sets of sources to `SetSource`. These are the `APPLY_WGS_SCATTER` type flags you see. In deal.ii, they are used to specify which unit cell matrices to construct. Here... It makes little sense though.

zhardy-lanl:
> > To me it really feels as though framework/io should be moved to framework/mesh/io as that everything in it has to do with meshing.
> 
> I started this way, but then I realized this layer could serve bigger purpose. For example, the code that outputs field function can be refactored to this approach as well, i.e. this new API would have for example `VtkIO::ToFile(FieldFunction, ...)`.
> 
> And the cross-section readers/writers can live here as well. Thoughts?

I personally still like mesh stuff living with mesh, XS stuff living with XS, etc. I don't really like the idea of having to navigate to another part of the code base to find where things are written/read. The disjointedness gives me the same feeling of the sweeper stuff living in `framework/mesh` but being used exclusively in LBS. To the point about field functions... Having something like `MeshIO`, `XSIO`, `FieldFunctionIO` classes that contain all supported filetypes sounds appealing to me. This ties into my other comment about `FromFile`. In the classes mentioned above, we would have routines like `To(From)<file type>`.

framework/mesh/mesh.cc
andrsd: I was basing the left, right for 1D on the fact  that 1D would be in the X-coordinate. Maybe, I misunderstood, but I thought that was also a wanted change. Should 1D stay in the Z-coordinate?

framework/mesh/mesh.h
wdhawkins: I agree they have their place. I just don't see a necessary use case for us. It's too easy for inexperienced programmers to get things wrong (case in point... where we could have a mesh that's 1D, 2D, 3D all at the same time). For a code that may have a fair number of student programmers with limited experience, I'm just not a fan.

zhardy-lanl: What would you do with the `SourceFlags` then? Use a lot of boolean function arguments for `SetSource` routines?

framework/mesh/mesh.cc
zhardy-lanl: I would be totally fine with 1D being the x-coordinate, did we make that decision?

framework/mesh/mesh_generator/from_file_mesh_generator.cc
zhardy-lanl: `FromOBJ` should have been `FromObj`. What are your thoughts on my proposal from my comment?

> Having something like MeshIO, XSIO, FieldFunctionIO classes that contain all supported filetypes sounds appealing to me. This ties into my other comment about FromFile. In the classes mentioned above, we would have routines like To(From)<file type>.

framework/io/obj_io.cc
```diff
@@ -0,0 +1,392 @@
...
+
+        // Convert word to number (Vertex)
+        try
+        {
+          int numValue = std::stoi(vert_word);
+          cell->vertex_ids.push_back(numValue - 1);
+        }
+        catch (const std::invalid_argument& ia)
+        {
+          log.Log0Warning() << "Failed converting work to number in line " << file_line
```
wdhawkins, line 147: work->word

framework/mesh/mesh.h
wdhawkins: I haven't looked at the `SetSource` routines, so I can't offer an opinion. It could be they are the use case that pushes us to a bitset. If that turned out to be the case, it might make sense to write a flags class that makes it easier for our more inexperienced users. I've seen code where scoped enums are are used on top of a bitset to make something that works pretty well and hides all of the bitset complications.

framework/mesh/mesh_generator/from_file_mesh_generator.cc
andrsd: > Having something like MeshIO, XSIO, FieldFunctionIO classes...

This works too. I am assuming, the code would go into `io` dir in the appropriate location, e.g. `framework/mesh/io`. I am thinking one header file, but multiple `.cc` file, per file format...

Also, if `OBJ` becomes `Obj`, then `VTU` and `PVTU` should be `Vtu` and `Pvtu`? Or do you see that as an acronym, thus keep it all-caps?

framework/io/obj_io.cc
```diff
@@ -0,0 +1,392 @@
...
+  std::ifstream file;
+  file.open(options.file_name);
+  if (not file.is_open())
+  {
+    log.LogAllError() << "Failed to open file: " << options.file_name << " in call "
+                      << "to ImportFromOBJFile \n";
+    Exit(EXIT_FAILURE);
+  }
+
+  opensn::mpi_comm.barrier();
```
wdhawkins, line 28: This can be removed.

framework/io/obj_io.cc
```diff
@@ -0,0 +1,392 @@
...
+          if (k == 1)
+            edge.first = vertex_id - 1;
+          if (k == 2)
+            edge.second = vertex_id - 1;
+        }
+
+        // Catch conversion error
+        catch (const std::invalid_argument& ia)
+        {
+          log.Log0Warning() << "Failed to text to integer in line " << file_line << std::endl;
```
wdhawkins, line 203: I think we're missing a `convert` here. Maybe `Failed to convert text to integer in line`.

framework/io/obj_io.cc
```diff
@@ -0,0 +1,392 @@
...
+      vertex_map.push_back(m);
+
+    // Build set of cell vertices
+    std::set<size_t> cell_vertex_id_set;
+    for (const auto& cell_ptr : block_data.at(main_block_id).cells)
+      for (size_t vid : cell_ptr->vertex_ids)
+        cell_vertex_id_set.insert(vid);
+
+    // Make cell_vertices and edit map
+    {
```
wdhawkins, line 267: Why do we restrict scope here?

framework/io/obj_io.cc
```diff
@@ -0,0 +1,392 @@
...
+ObjIO::FromFile(const UnpartitionedMesh::Options& options)
+{
+  const std::string fname = "ObjIO::ReadFromWavefrontOBJ";
+
+  // Opening the file
+  std::ifstream file;
+  file.open(options.file_name);
+  if (not file.is_open())
+  {
+    log.LogAllError() << "Failed to open file: " << options.file_name << " in call "
```
wdhawkins, line 23: Show we `throw` here? I think this is the only place I see where`Exit` is called instead of a `throw`.

framework/mesh/mesh.h
zhardy-lanl: This is the call signature:
```
active_set_source_function_(groupset,
                            q_moments_local_,
                            phi_old_local_,
                            densities_local_,
                            APPLY_FIXED_SOURCES | APPLY_AGS_FISSION_SOURCES |
                              APPLY_WGS_FISSION_SOURCES);
```
The way it is used is quite user-readable and makes it easy to mix and match sources, which is particularly useful for the k-eigenvalue solvers. It would be good to keep a call signature as readable as this.

framework/mesh/mesh_generator/from_file_mesh_generator.cc
zhardy-lanl: > This works too. I am assuming, the code would go into io dir in the appropriate location, e.g. framework/mesh/io. I am thinking one header file, but multiple .cc file, per file format...

I would say either multiple `.cc` files or a single file `mesh_io.h(cc)`. 

> Also, if OBJ becomes Obj, then VTU and PVTU should be Vtu and Pvtu? Or do you see that as an acronym, thus keep it all-caps?

I would try to name everything as it is commonly written. If `OBJ` is an acronym, then it should be all caps if that is how it is normally written. I don't lean heavily one way or the other, I just think it is good to establish a standard for it.

ragusa: FWIIW, on Wikipedia, it's VTK (hence I assume also VTU) and OBJ.

framework/io/obj_io.cc
andrsd: This way handling the error of "unable to open file" seems to be used in more places in the whole code. I will patch it with `throw` where I think it is appropriate - like these IO routines. The VTK reader does throw, but OBJ and Gmsh `Exit()`.

framework/io/obj_io.cc
andrsd: I think somebody wanted to use range-based for loop, but they needed an iteration index, but did not want to propagate the name further in the code... This is just 🤦‍♂️

andrsd: nvm, `cell_vertex_id_set` is a `std::set`, so only to restrict the scope of the `new_id`.

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/sweep_chunks/lbs_curvilinear_sweep_chunk_pwl.cc
ragusa: @zhardy-lanl @andrsd I've updated the curvilinear issue to include this to check.

ragusa (approved):
I approve, pending the changes requesting by others

framework/mesh/mesh.h
zhardy-lanl, line 61: I take it this is going to be handled in a subsequent PR?

framework/mesh/mesh_continuum/mesh_continuum.h
```diff
@@ -38,27 +38,11 @@ class MeshContinuum
...
-    : local_cells(local_cells_),
-      cells(local_cells_,
-            ghost_cells_,
-            global_cell_id_to_local_id_map_,
-            global_cell_id_to_nonlocal_id_map_)
-  {
-  }
+  MeshContinuum();
+
+  unsigned Dimension() const { return dim_; }
```
zhardy-lanl, line 44:
```suggestion
  unsigned int Dimension() const { return dim_; }
```
Not that it makes a difference, but I don't believe we've used `unsigned` anywhere else in the code. This could maybe be added to the coding standards (either with or without `int`). I'm indifferent as to which to use.

framework/mesh/mesh_continuum/mesh_continuum.h
zhardy-lanl, line 1: Please move all of the class attributes to the bottom of the class declaration instead of having them split up.

zhardy-lanl (approved):
Thank you for the revisions! Please address the small number of formatting/syntactic based comments before merging.

wdhawkins (approved):

framework/mesh/mesh.h
andrsd: Yes

framework/mesh/mesh_continuum/mesh_continuum.h
```diff
@@ -228,14 +198,30 @@ class MeshContinuum
...
+  std::map<uint64_t, uint64_t> global_cell_id_to_nonlocal_id_map_;
+
+  uint64_t global_vertex_count_ = 0;
+
+  MeshAttributes attributes_ = NONE;
+  OrthoMeshAttributes ortho_attributes_;
+
+  std::map<uint64_t, std::string> boundary_id_map_;
+  /// Spatial dimension
+  unsigned int dim_;
```
zhardy-lanl, line 224:
```suggestion
  unsigned int dim_;  ///< Spatial dimension
  MeshAttributes attributes_ = NONE;
  OrthoMeshAttributes ortho_attributes_;
  std::map<uint64_t, std::string> boundary_id_map_;
  
   uint64_t global_vertex_count_ = 0;
   
  std::vector<std::unique_ptr<Cell>> local_cells_; ///< Actual local cells
  std::vector<std::unique_ptr<Cell>> ghost_cells_; ///< Locally stored ghosts

  std::map<uint64_t, uint64_t> global_cell_id_to_local_id_map_;
  std::map<uint64_t, uint64_t> global_cell_id_to_nonlocal_id_map_;
```

framework/mesh/unpartitioned_mesh/unpartitioned_mesh.h
```diff
@@ -191,23 +124,21 @@ class UnpartitionedMesh
...
+  std::vector<LightWeightCell*> raw_boundary_cells_;
+  std::vector<std::set<uint64_t>> vertex_cell_subscriptions_;
+
+  /// Spatial mesh dimension
+  unsigned int dim_;
+  struct OrthoMeshAttributes ortho_attrs_;
+  MeshAttributes attributes_ = NONE;
+  Options mesh_options_;
+  BoundBox bound_box_;
+  std::map<uint64_t, std::string> boundary_id_map_;
```
zhardy-lanl, line 141:
```suggestion
  unsigned int dim_;  ///< Spatial dimension
  struct OrthoMeshAttributes ortho_attrs_;
  MeshAttributes attributes_ = NONE;
  Options mesh_options_;
  BoundBox bound_box_;
  std::map<uint64_t, std::string> boundary_id_map_;
  
  std::vector<Vertex> vertices_;
  std::vector<LightWeightCell*> raw_cells_;
  std::vector<LightWeightCell*> raw_boundary_cells_;
  std::vector<std::set<uint64_t>> vertex_cell_subscriptions_;
```

zhardy-lanl: Should it be `struct OrthoMeshAttributes` or just `OrthoMeshAttributes`?

zhardy-lanl (changes requested):
I won't be offended if you ignore this and merge anyways. Just some attribute reorganization to put informative versus working data structures together.

framework/mesh/unpartitioned_mesh/unpartitioned_mesh.h
andrsd: `struct` is not needed, I will remove it...

zhardy-lanl (approved):

PR #225, opened 2024-04-23
Updating regression ci workflow. Adding build test workflow.

wdhawkins:
I've created multiple build environments on our CI machine to test various compilers we support. Initially, we are testing the following:

GCC: 9.5, 10.5, 12.3, 13.2
Clang: 12.0.1, 14.0.6, 16.0.6, 17.0.6
Intel: 2024.1

Since we need c++17 support, gcc 9.5 is about as far back as we can go. 8.x would probably work with a bit of effort. We could almost certainly go back further with clang, but I think five versions (actually 6 version now) from current is probably good. oneAPI is only just getting stable in my opinion, so the current version is good enough, I think.

I've never tried using a matrix with GitHub workflows, so I'm hoping @andrsd can weigh in on that. There will probably be some tweaking needed for the workflow.


.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test_gcc:
+    runs-on: [self-hosted]
+    matrix:
+      gcc: [9.5.0, 10.5.0, 12.3.0, 13.2.0]
```
andrsd, line 17: Do this instead:
```
compiler:
  - gcc/9.5.0
  - gcc/10.5.0
  - gcc/12.3.0
  - gcc/13.2.0
  - clang/12.0.1
  - clang/14.0.6
  - clang/16.0.6
  - clang/17.0.6
  - intel/oneAPI-2021.4
```

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: configure
+        shell: bash
+        run: |
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/gcc/${{ matrix.gcc }}
```
andrsd, line 24:
```
module load opensn/${{ matrix.compiler }}
```

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+          ./configure.sh
+      - name: make
+        shell: bash
+        run: |
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/gcc/${{ matrix.gcc }}
```
andrsd, line 30:
```
module load opensn/${{ matrix.compiler }}
```

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+          cmake --build build --parallel 64
+      - name: test
+        shell: bash
+        run: |
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/gcc/${{ matrix.gcc }}
```
andrsd, line 36:
```
module load opensn/${{ matrix.compiler }}
```

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+    - cron: "0 2 * * *"
+  push:
+    branches: [main]
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test_gcc:
```
andrsd, line 14: `test_gcc` -> `build`

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+    runs-on: [self-hosted]
```
andrsd, line 15: above this, add (same level as `runs-on`) :
```
name: "${{ matrix.compiler }}"
```

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+    matrix:
```
andrsd, line 16:
```
strategy:
  fail-fast: false
  matrix:
     ...    
```

`strategy` is at the same level as `runs-on`.

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/gcc/${{ matrix.gcc }}
+          test/run_tests -d test/ -j 32 -v 1 -w 3
+      - name: test tutorials
+        shell: bash
+        run: |
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/gcc/${{ matrix.gcc }}
+          test/run_tests -d tutorials -j 32 -v 1 -w 3
+  test_clang:
```
andrsd, line 44: Now, you don't need this, since it is already in the test matrix

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,105 @@
...
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/clang/${{ matrix.clang }}
+          test/run_tests -d test/ -j 32 -v 1 -w 3
+      - name: test tutorials
+        shell: bash
+        run: |
+          export MODULEPATH=/scratch-local/software/modulefiles
+          module load opensn/clang/${{ matrix.clang }}
+          test/run_tests -d tutorials -j 32 -v 1 -w 3
+  test_intel:
```
andrsd, line 74: And you don't need this either.

andrsd (changes requested):

andrsd:
You could also lump the multiple builds into regression workflow and run tests only on a specific configuration via:

the test step would look like

```
- name: test
  if: ${{ matrix.compiler == 'gcc/12.3.0')
  shell: bash
  run: |
    export MODULEPATH=/scratch-local/software/modulefiles
    module load opensn/${{ matrix.compiler }}
    test/run_tests -d test/ -j 32 -v 1 -w 3
```

The line with `if` does the magic. `test tutorials` step would gain that condition as well

andrsd:
I just noticed that tests and tutorials are execute on every single configuration. If that is desired, then the regression test work flow can be replaced by the new one. Or the new one should be integrated into regression tests.

If you want to just test on one selected configuration, then the above comment would be the way to do it.

.github/workflows/builds.yaml
wdhawkins: I think I correctly addressed all of the comments. Let's me know what you think of the latest version. I also removed the use of `configure.sh`.

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,43 @@
+name: Build Tests
```
andrsd, line 1: This should be "Nightly builds"

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,43 @@
...
+    compiler:
+      - gcc/9.5.0
+      - gcc/10.5.0
+      - gcc/12.3.0
+      - gcc/13.2.0
+      - clang/12.0.1
+      - clang/14.0.6
+      - clang/16.0.6
+      - clang/17.0.6
+      - intel/oneAPI-2021.4
```
andrsd, line 22: this is still not correct
```
strategy:
  fail-fast: false
  matrix:
    compiler:
      - gcc/...
      - ...
```
See [doco](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy)

andrsd (changes requested):

andrsd (approved):

ragusa (approved):

zhardy-lanl (approved):

PR #226, opened 2024-04-24
Adding file/line info to additional Lua error messages.

wdhawkins:


lua/framework/lua.cc
```diff
@@ -12,20 +12,20 @@
 static int a = 15;
 
 void
-LuaPostArgAmountError(const std::string& func_name, int expected, int given)
+LuaPostArgAmountError(const std::string& func_name, lua_State* L, int expected, int given)
 {
-  throw std::invalid_argument("Incorrect amount of arguments supplied in " + func_name +
-                              " expected " + std::to_string(expected) + " arguments " + " but " +
-                              std::to_string(given) + " provided");
+  throw std::invalid_argument(
```
andrsd, line 17: There is dedicated `InvalidLuaArgument` exception that you can use instead. But, this is ok too. I will be refactoring/removing most of these functions in future. The is some code that still relies on these, so they are still around...

ragusa (approved):

zhardy-lanl (approved):

andrsd (approved):

PR #233, opened 2024-04-27
Throw an exception if we generate a partition with zero cells

wdhawkins:
Throw an error instead of crashing if we generate a partition with zero cells. This can easily happen if we have more ranks than cells. For us to exit gracefully in this case, it's necessary to have every location calculate min/max/avg cells per partition and throw if min is zero.

### DIFF
--- a/framework/mesh/mesh_generator/mesh_generator.cc
+++ b/framework/mesh/mesh_generator/mesh_generator.cc
@@ -91,11 +91,32 @@ MeshGenerator::Execute()
   // Generate final umesh and convert it
   current_umesh = GenerateUnpartitionedMesh(std::move(current_umesh));
 
+  auto num_partitions = opensn::mpi_comm.size();
   std::vector<int64_t> cell_pids;
   if (opensn::mpi_comm.rank() == 0)
-    cell_pids = PartitionMesh(*current_umesh, opensn::mpi_comm.size());
-
+    cell_pids = PartitionMesh(*current_umesh, num_partitions);
   BroadcastPIDs(cell_pids, 0, mpi_comm);
+
+  std::vector<size_t> partI_num_cells(num_partitions, 0);
+  for (int64_t pid : cell_pids)
+    partI_num_cells[pid] += 1;
+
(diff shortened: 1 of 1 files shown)


framework/mesh/mesh_generator/mesh_generator.cc
```diff
@@ -177,31 +198,13 @@ MeshGenerator::PartitionMesh(const UnpartitionedMesh& input_umesh, int num_parti
     }
   }
 
-  // Note A: We do not add the diagonal here. If we do it, ParMETIS seems
+  // Note A: We do not add the diagonal here. If we do, ParMETIS seems
```
ragusa, line 201: I know it is not your comment, but what does this mean? And what about PTSCOTCH then? Should we create an issue to remember to investigate the meaning of that comment? It seems important for parallel efficiency.

andrsd:
Should partition with zero cells really be a stopping condition?

framework/mesh/mesh_generator/mesh_generator.cc
wdhawkins: I'm not sure what this comment means. I'm not even sure it's still relevant to the code that it references. It doesn't look to me like we are excluding anything. I'll open an issue to investigate what's going on here.

wdhawkins:
> Should partition with zero cells really be a stopping condition?

I believe it should. I think from a "know what you're doing and use the code correctly" perspective, it always should. But, I can see the argument for robustness. I think there would be a number of fixes that would be necessary to support this. With a few minor fixes, I can get get the code into the topological sorting for the sweep, but then things go sideways again. At some point, I'll probably rewrite that chunk of code, and I'll try to make it work for cases where we have partitions with no cells.

zhardy-lanl (approved):

andrsd (approved):

framework/mesh/mesh_generator/mesh_generator.cc
ragusa: see email

ragusa (approved):

PR #234, opened 2024-04-29
Adding hdf5 package to gh-pages workflow.

wdhawkins:
I don't use `conda`, but I believe this change will fix the gh-pages workflow.  Maybe someone more familiar with it can weight in.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -35,7 +35,8 @@ jobs:
             petsc==3.17.5 \
             vtk \
             lua \
-            caliper
+            caliper \
+            hdf5
           sudo pip install -r doc/requirements.txt
 
       - name: Build doco


andrsd (approved):
This is a correct fix

ragusa (approved):

andrsd:
Maybe, I spoke too soon. HDF5 is installed as a dependency, see [here](https://github.com/Open-Sn/opensn/actions/runs/8880183331/job/24379658350#step:4:61). It is because I built PETSc with HDF5 support. Let me pull `main` and see if I can replicate this problem...

andrsd:
The problem is that `opensn` is trying to use the C++ component from HDF5, but the `hdf5` package in conda is not build with c++ support. There are 2 courses of action: 1. Update opensn to use `h5pp`. IIRC, is does not need the c++ part of hdf5. 2. rebuild the hdf5 package in conda so that is has the c++ component.

andrsd:
I am trying to re-build hdf5 with c++ support, and I see this:
```
configure: error: --enable-cxx and --enable-parallel flags are incompatible. Use --enable-unsupported to override this error.
```
Note that my configure line is taken from PETSc, so that I have HDF5 compatible with PETSc. I read that most hdf5 builds are not done correctly to work with PETSc. However, we are not really using anything in PETSc that needs HDF5, but this is an interesting observation. I also, wonder if `1.14.3` (which is what we have in `opensn`) fixed that - I am still on `1.12.1` in the conda package - because that's what is being used in PETSc 3.17.x.

zhardy-lanl:
> I am trying to re-build hdf5 with c++ support, and I see this:
> 
> ```
> configure: error: --enable-cxx and --enable-parallel flags are incompatible. Use --enable-unsupported to override this error.
> ```
> 
> Note that my configure line is taken from PETSc, so that I have HDF5 compatible with PETSc. I read that most hdf5 builds are not done correctly to work with PETSc. However, we are not really using anything in PETSc that needs HDF5, but this is an interesting observation. I also, wonder if `1.14.3` (which is what we have in `opensn`) fixed that - I am still on `1.12.1` in the conda package - because that's what is being used in PETSc 3.17.x.

@andrsd Please let me know what the appropriate fix for this is as that I have the same issue with conda...

andrsd:
I rebuilt the hdf5 package with the `--enable-unsupported` flag and all seems to be fine now. The gh-pages workflow is now green, see [here](https://github.com/Open-Sn/opensn/actions/runs/8880183331).

So, I think this can be closed.

zhardy-lanl (approved):

PR #241, opened 2024-05-01
Removing dead and duplicate code

andrsd:


framework/mesh/mesh.h
```diff
@@ -90,44 +90,6 @@ ComputeLBF(std::vector<Vector3>& points, std::vector<double>& x_cuts, std::vecto
 /** Decomposes a 2D surface mesh using the centroids in a Px-Py fashion.*/
 void DecomposeSurfaceMeshPxPy(const SurfaceMesh& smesh, int Px, int Py);
```
ragusa, line 91: is this (`DecomposeSurfaceMeshPxPy`) still used?

framework/mesh/mesh.cc
```diff
@@ -15,297 +15,4 @@ GetCurrentMesh()
...
-    {
-      auto cell =
-        new UnpartitionedMesh::LightWeightCell(CellType::POLYGON, CellType::QUADRILATERAL);
-
-      // vertex ids:   face ids:
-      //                 2
-      //    3---2      x---x
-      //    |   |     3|   |1
-      //    0---1      x---x
-      //                 0
```
ragusa, line 125: can we make sure that the comments that describe the vertex and face IDs are still somewhere appropriate in the code?

framework/mesh/mesh.cc
```diff
@@ -15,297 +15,4 @@ GetCurrentMesh()
...
-                                                 vmap[i][j + 1][k],
-                                                 vmap[i + 1][j + 1][k],
-                                                 vmap[i + 1][j][k],
-
-                                                 vmap[i][j][k + 1],
-                                                 vmap[i][j + 1][k + 1],
-                                                 vmap[i + 1][j + 1][k + 1],
-                                                 vmap[i + 1][j][k + 1]};
-
-        // East face
```
ragusa, line 236: I also liked the comments about East face, West face, ..... do we still have then in the code that those lines are duplicates of?

framework/mesh/mesh.cc
andrsd: What was in `mesh.cc` was copy pasted into `orthogonal_mesh_generator.cc`, see https://github.com/Open-Sn/opensn/blob/c818552e2e8acd0631bdd37d4e3312e4bac742e5/framework/mesh/mesh_generator/orthogonal_mesh_generator.cc#L240-L245

framework/mesh/mesh.h
andrsd: No, that is dead code as well. Do you want me to remove it as well?

andrsd: The code was removed but the declaration stayed. The same is true for `ComputeLBF`, so I will remove that as well...

ragusa (approved):

wdhawkins (approved):

zhardy-lanl (approved):

PR #253, opened 2024-05-07
Removing PETSc forward declarations

andrsd:


andrsd:
It was there so the code would not have to `#include` PETSc header files (my guess)

wdhawkins (approved):

zhardy-lanl (approved):

Issue #252, opened 2024-05-06
Remove `mat.AddProperty` Lua API and Refactor `mat.SetProperty`


zhardy-lanl:
- [x] Remove the `mat.AddProperty` Lua API
- [x] Refactor `mat.SetProperty` so that it 1) creates a property on a material if the material does not contain the specified property type, and 2) overwrites an existing property of the same type.
- [x] Decide whether to remove property names and whether multiple properties of the same type should be allowed.

Refs #250

PR #256, opened 2024-05-07
Formatting input files for consistency

wdhawkins:
This PR uses ~https://github.com/Koihik/LuaFormatter~ https://github.com/JohnnyMorganz/StyLua  to bring some consistency to the format of our input files. If we want to start enforcing format checks on our `*.lua` files using this tool or another, we can address that as a separate issue/PR.

`LuaFormatter` mangles some comment blocks, and I can't seem to prevent that from happening. Instead of adding `noformat` tags to all of the input files, I decided to switch to `stylua`. It has less configuration options, but I think the formatting is more consistent. It also seems to be the go-to Lua formatter.

I've added the style file as `.stylua.toml`.

PR #257, opened 2024-05-07
Material Property API

zhardy-lanl:
- Modifies the functionality of `mat.SetProperty`.
    - If the specified property type lives on the material already, the property is overwritten.
    - If the specified property type does not live on the material, it is created.
- Eliminated the use of `mat.AddProperty` throughout the tests. 
- Deprecates `mat.AddProperty` to avoid breaking existing scripts.

Refs #252

### DIFF
--- a/lua/framework/materials/lua_material.cc
+++ b/lua/framework/materials/lua_material.cc
@@ -86,6 +86,9 @@ MatAddMaterial(lua_State* L)
 int
 MatAddProperty(lua_State* L)
 {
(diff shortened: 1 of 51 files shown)


andrsd:
There is some weird reformatting going on in the lua files - makes this PR much harder to review. IDK if that was intentional, probably not. IDK how difficult it would be to undo it, but if #256 goes in after this, it would take care of it...

ragusa (approved):

andrsd (approved):

wdhawkins (approved):

PR #258, opened 2024-05-08
IsotropicMultiGrpSource -> IsotropicMultiGroupSource

zhardy-lanl:
This simple PR just spells out "Group" in `IsotropicMultiGrpSource` and renames the header file.

### DIFF
--- a/framework/materials/scalar_value.h
+++ b/framework/materials/scalar_value.h
@@ -8,7 +8,7 @@
 namespace opensn
 {
 
-/**Simple scalar material property.*/
+/** Simple scalar-valued material property. */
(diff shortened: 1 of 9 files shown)


framework/materials/scalar_value.h
```diff
@@ -8,7 +8,7 @@
 namespace opensn
 {
 
-/**Simple scalar material property.*/
+/** Simple scalar valued material property. */
```
ragusa, line 11: scalar-valued

Don't do the change if you think I am excessively picky.

ragusa (approved):
see my comment, just in case

andrsd (approved):

PR #261, opened 2024-05-10
Updating field function interpolations

wdhawkins:
This commit updates the field function interpolation routines and includes a rewrite of the line interpolator. The line interpolator now implements the collective MAX, SUM, and AVG operations that are also defined for the volume interpolator. The ExportPython() routine has been replaced with an export routine that generates a CSV file. Other refactoring for style and coding standards. 

The slice field function is still present in this PR, but has not been updated for style, etc. It is scheduled for removal shortly.

### DIFF
--- a/external/mpicpp-lite/impl/Communicator.h
+++ b/external/mpicpp-lite/impl/Communicator.h
@@ -264,6 +264,24 @@ public:
     template <typename T>
     void gather(const T * in_values, int n, std::vector<T> & out_values, int root) const;
 
+    /// Gather together values from a group of processes
+    ///
+    /// @tparam T C++ datatype
+    /// @param in_values[in] Values to send
+    /// @param out_values[out] Buffer to receive the data
+    /// @param out_counts[in] Integer array (of length group size) containing the number of elements
+    ///        that are to be received from each process
+    /// @param out_offsets[in] Integer array (of length group size). Entry `i` specifies the
+    ///        displacement (relative to out_values) at which to place the incoming data from
+    ///        process `i`
+    /// @param root Rank of receiving process
+    template <typename T>
+    void gather(const std::vector<T> & in_values,
+                std::vector<T> & out_values,
+                const std::vector<int> & out_counts,
+                const std::vector<int> & out_offsets,
+                int root) const;
+
     /// Gathers data from all tasks and distribute the combined data to all tasks
     ///
     /// @tparam T C++ type of the data
@@ -809,6 +827,31 @@ Communicator::gather(const T * in_values, int n, std::vector<T> & out_values, in
     if (rank() == root)
         out_values.resize(size() * (std::size_t) n);
     gather(in_values, n, out_values.data(), root);
+}
+
+template <typename T>
+inline void
+Communicator::gather(const std::vector<T> & in_values,
+                     std::vector<T> & out_values,
+                     const std::vector<int> & out_counts,
+                     const std::vector<int> & out_offsets,
+                     int root) const
+{
+    assert(out_counts.size() == size());
+    assert(out_offsets.size() == size());
+    int n_out_vals = 0;
+    for (std::size_t i = 0; i < out_counts.size(); i++)
+        n_out_vals += out_counts[i];
+    out_values.resize(n_out_vals);
+    MPI_CHECK_SELF(MPI_Gatherv(in_values.data(),
+                               in_values.size(),
+                               get_mpi_datatype<T>(),
+                               out_values.data(),
+                               out_counts.data(),
+                               out_offsets.data(),
+                               get_mpi_datatype<T>(),
+                               root,
+                               this->comm));
 }
 
 template <typename T>
--- a/framework/field_functions/interpolation/ffinter_line.cc
+++ b/framework/field_functions/interpolation/ffinter_line.cc
@@ -125,7 +125,6 @@ FieldFunctionInterpolationLine::Export(std::string base_name)
   std::vector<int> local_data_sizes(opensn::mpi_comm.size(), 0);
   int local_size = local_data.size();
   opensn::mpi_comm.gather(local_size, local_data_sizes, 0);
-  //MPI_Gather(&local_size, 1, MPI_INT, local_data_sizes.data(), 1, MPI_INT, 0, MPI_COMM_WORLD);
 
   // Compute global size of coordinate and interpolation data and each location's offset
   std::vector<double> global_data;
@@ -179,8 +178,8 @@ FieldFunctionInterpolationLine::Export(std::string base_name)
     ofile << "x,y,z," << ref_ff_->TextName() << "\n";
     for (auto point_data : values)
     {
-      auto [x, y, z, fv ] = point_data;
-      ofile << std::setprecision(5) << x << "," << y << "," << z << "," << fv << "\n";
+      auto [x, y, z, fv] = point_data;
+      ofile << std::setprecision(7) << x << "," << y << "," << z << "," << fv << "\n";
     }
     ofile.close();
 
--- a/framework/field_functions/interpolation/ffinter_line.cc
+++ b/framework/field_functions/interpolation/ffinter_line.cc
@@ -3,12 +3,12 @@
 
 #include "framework/field_functions/interpolation/ffinter_line.h"
 #include "framework/math/spatial_discretization/spatial_discretization.h"
-#include "framework/math/vector_ghost_communicator//vector_ghost_communicator.h"
+#include "framework/math/vector_ghost_communicator/vector_ghost_communicator.h"
 #include "framework/field_functions/field_function_grid_based.h"
-#include "framework/mesh/cell/cell.h"
 #include "framework/mesh/mesh_continuum/mesh_continuum.h"
-#include "framework/runtime.h"
+#include "framework/mesh/cell/cell.h"
 #include "framework/logging/log.h"
+#include "framework/runtime.h"
 #include <fstream>
 
 namespace opensn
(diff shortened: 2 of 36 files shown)


framework/field_functions/interpolation/ffinter_line.cc
```diff
@@ -69,196 +60,132 @@ void
...
+    ofile.open(filename);
+    ofile << "x,y,z," << ref_ff_->TextName() << "\n";
+    for (auto point_data : values)
     {
-      int ff = ca + field_functions_.size();
-      ofile << "plt.plot(data" << ff << "[:,3],data" << ff << "[:,4]"
-            << ",label=\"CustomArray" << ca << "\""
-            << ")\n";
+      auto [x, y, z, fv] = point_data;
+      ofile << std::setprecision(5) << x << "," << y << "," << z << "," << fv << "\n";
```
ragusa, line 182:
```suggestion
      ofile << std::setprecision(5) << x << "," << y << "," << z << "," << std::setprecision(7) << fv << "\n";
```

ragusa (approved):

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -954,13 +954,8 @@ MeshContinuum::CheckPointInsideCell(const Cell& cell, const Vector3& point) cons
...
 
-    const auto v01 = v1 - v0;
-    const auto v0p = point - v0;
-
-    const double v0p_dot_v01 = v0p.Dot(v01);
-
-    if (not(v0p_dot_v01 >= 0 and v0p_dot_v01 < v01.Norm()))
-      inside = false;
+    if (((v0.z - point.z) * (v1.z - point.z)) > 0.0)
+      return false;
```
zhardy-lanl, line 958: This routine is a quite messy... `inside` is set to `true` at first. For slabs, if a point is found to not be within a cell, `false` is returned within the `if` statement. For polygons, `inside` is set to `false` when that condition is encountered, and for polyhedra, `inside` is set to `false` at first and then a check of whether this a point is inside the cell is done to potentially change it to `true`. 

Can this be changed to be uniform across cell types? The variable `inside` doesn't seem to be necessary. For each cell type, check either if the cell is inside(outside) the cell (same for all cell types), return true(false) if the condition is met, and the have the last line of the routine be `return false(true)`?

zhardy-lanl (changes requested):

framework/mesh/mesh_continuum/mesh_continuum.cc
zhardy-lanl: Whatever you would prefer is fine with me. I think we will all agree it should be rewritten, so the question is whether it gets rewritten now or later. If you feel like it isn't a big endeavor and want to rewrite it here, I'm not opposed. If not, an issue would be best before closing this PR.

wdhawkins: I rewrite it for consistency, and then open a PR to review the actual methods.

wdhawkins: Done.

external/mpicpp-lite/impl/Communicator.h
```diff
@@ -264,6 +264,24 @@ class Communicator {
...
+    ///        displacement (relative to out_values) at which to place the incoming data from
+    ///        process `i`
+    /// @param root Rank of receiving process
+    template <typename T>
+    void gather(const std::vector<T> & in_values,
+                std::vector<T> & out_values,
+                const std::vector<int> & out_counts,
+                const std::vector<int> & out_offsets,
+                int root) const;
+
```
andrsd, line 284: If you are updating `mpicpp_lite`, please send a PR into [upstream](https://github.com/andrsd/mpicpp-lite) first.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,69 @@ bool
 MeshContinuum::CheckPointInsideCell(const Cell& cell, const Vector3& point) const
 {
   const auto& grid_ref = *this;
-  typedef Vector3 Vec3;
```
andrsd, line 934: Nice :+1:

framework/field_functions/interpolation/ffinter_line.cc
```diff
@@ -69,196 +60,132 @@ void
...
+    mpi_comm.all_reduce(local_sum, global_sum, mpi::op::sum<double>());
+    op_value_ = global_sum / static_cast<double>(global_size);
+  }
+  else if (op_type_ == FieldFunctionInterpolationOperation::OP_MAX)
+    mpi_comm.all_reduce(local_max, op_value_, mpi::op::max<double>());
 }
 
 void
-FieldFunctionInterpolationLine::ExportPython(std::string base_name)
+FieldFunctionInterpolationLine::Export(std::string base_name)
```
andrsd, line 110: Should this be called `ExportToCSV`?

lua/framework/mesh/field_function_interpolation/ffinterpol_export.cc
```diff
@@ -2,32 +2,28 @@
...
+#include "framework/field_functions/interpolation/ffinterpolation.h"
 #include "framework/console/console.h"
+#include "framework/logging/log.h"
+#include "framework/runtime.h"
 
 namespace opensnlua
 {
 
-RegisterLuaFunctionInNamespace(FFInterpolationExportPython, fieldfunc, ExportPython);
+RegisterLuaFunctionInNamespace(FFInterpolationExport, fieldfunc, Export);
```
andrsd, line 14: Should this be `ExportToCSV`? So that it matches with the export API?

andrsd (changes requested):

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,69 @@ bool
...
-
     const auto pc = point - c;
 
     if (pc.Dot(n) > 0.0)
       return true;
-    else
-      return false;
+
+    return false;
   };
```
zhardy-lanl, line 947: Just something to consider: This could be moved to be located with the check for polyhedra.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,69 @@ bool
...
-    const double v0p_dot_v01 = v0p.Dot(v01);
-
-    if (not(v0p_dot_v01 >= 0 and v0p_dot_v01 < v01.Norm()))
-      inside = false;
-  } // slab
-
+    if (((v0.z - point.z) * (v1.z - point.z)) > 0.0)
+      return false;
+  }
   else if (cell.Type() == CellType::POLYGON)
```
zhardy-lanl, line 957: A comment on the logic in this check would be helpful as that it is not immediately obvious.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,69 @@ bool
...
     if (pc.Dot(n) > 0.0)
       return true;
-    else
-      return false;
+
+    return false;
   };
 
-  bool inside = true;
   if (cell.Type() == CellType::SLAB)
```
zhardy-lanl, line 949: For consistency with the other branches of the if statement, a comment here on the logic of the check would be good.

zhardy-lanl (changes requested):
Overall, much better! Just a few small requests for clarification for better readability.

framework/field_functions/interpolation/ffinter_line.cc
wdhawkins: I debated that. What I was trying to prevent was having an `ExportToCSV`, `ExportToHDF`, `ExportToNNN` -  some of which may not applicable (or even do anything) for every field function. For example, you probably wouldn't want to try to define `ExportToCSV` for volume field functions. It turns out that volume doesn't have export functionality, point probably doesn't really need it, and that just leaves slice and line. Slice is likely going away, so that's just going to leave line, and I went with just plain `Export`. Happy to change it though.

lua/framework/mesh/field_function_interpolation/ffinterpol_export.cc
wdhawkins: I do think this is the way to go for the user-facing interface... so ignore my comment above, and I'll fix this. :smile:

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,69 @@ bool
...
-    const auto v01 = v1 - v0;
-    const auto v0p = point - v0;
-
-    const double v0p_dot_v01 = v0p.Dot(v01);
-
-    if (not(v0p_dot_v01 >= 0 and v0p_dot_v01 < v01.Norm()))
-      inside = false;
-  } // slab
-
+    if (((v0.z - point.z) * (v1.z - point.z)) > 0.0)
```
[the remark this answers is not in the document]
wdhawkins: This reason I went for the first formulation is that it really avoids any need to talk about fuzzy comparisons.

external/mpicpp-lite/impl/Communicator.h
wdhawkins: Will do. In other codes I work in, the workflow is to patch locally, push upstream, and then pull the new upstream version when it's available. This way we don't block development. Guess that really doesn't apply in this case. 😆

andrsd: If you are at the mercy of the package maintainer, that makes sense. They may be slow in merging and possibly releasing new versions. But, it can create conflicts because your local copy can diverge from the upstream, so next time you are updating the local copy, you may need to resolve such conflicts. Updating upstream first would prevent that... That's what I was thinking. But, we can keep this in the PR, there was no development in `mpicpp-lite` lately, so there should be no conflicts...

Note, that for your PR into upstream, I would ask you for a little test case, so we know this works.

lua/framework/mesh/field_function_interpolation/ffinterpol_export.cc
andrsd: Yeah no problem. I think we talked about something like `FieldFunctionIO`, so the file format in the name will be a thing. Similarly, like we have it in the `MeshIO` module...

external/mpicpp-lite/impl/Communicator.h
wdhawkins: I'll submit a PR with a test case.

framework/field_functions/interpolation/ffinter_line.cc
wdhawkins: Fixed.

lua/framework/mesh/field_function_interpolation/ffinterpol_export.cc
wdhawkins: Fixed.

framework/mesh/mesh_continuum/mesh_continuum.cc
wdhawkins: Fixed.

framework/mesh/mesh_continuum/mesh_continuum.cc
wdhawkins: Fixed.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,69 @@ bool
...
-        inside = false;
-        break;
-      }
-    } // for face
-  }   // polygon
-
+        return false;
+    }
+  }
   else if (cell.Type() == CellType::POLYHEDRON)
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

framework/mesh/mesh_continuum/mesh_continuum.cc
wdhawkins: Fixed.

.clang-tidy
zhardy-lanl, line 1: Should this file be here?

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -931,103 +931,77 @@ bool
...
-    const auto v0p = point - v0;
-
-    const double v0p_dot_v01 = v0p.Dot(v01);
-
-    if (not(v0p_dot_v01 >= 0 and v0p_dot_v01 < v01.Norm()))
-      inside = false;
-  } // slab
-
+    // Check each cell edge. A point inside the cell will return a negative value. A point on either
+    // edge will return a zero value, and a point outside the cell will return a postiive value.
```
zhardy-lanl, line 941:
```suggestion
    // edge will return a zero value, and a point outside the cell will return a postive value.
```

zhardy-lanl (approved):
Small typo to fix, but other than that, good to go!

.clang-tidy
wdhawkins: Nope. Fixed.

framework/mesh/mesh_continuum/mesh_continuum.cc
wdhawkins: Good catch! Fixed.

andrsd (approved):

PR #264, opened 2024-05-13
Generate JSON compile commands database

wdhawkins:
This PR adds functionality to CMakeLists.txt to always generate a JSON compile commands database for use with various tools.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ project(opensn
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
 include(CMakePackageConfigHelpers)


andrsd:
Just out of curiosity, why is this not enabled from command line when you need it via `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`?

ragusa:
good question @andrsd . maybe what we should do instead is add a note in the install instructions so that people can turn it on if they want without forcing the export each time for everybody. what do you think @wdhawkins ?

wdhawkins:
Adding `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` will also work. However, since this is a completely innocuous change and enables so much additional code quality functionality, I think it's worthwhile to enable it by default. It adds one additional file to the `build` directory and doesn't slow down compilation. 

In the next week or so, I will have a PR with an initial version of a `clang-tidy` config that requires this functionality. The first set of fixes numbers in the few hundreds, but we will be "tidy" clean after that (at least with our config). This is something that we need to incorporate into our development workflow, IMO, and it'll be easier if we don't have to remember to compile with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` every time.

ragusa:
Ok. I am good with adding this to the repo then!! tag for info: @wdhawkins @andrsd

andrsd:
I see. If this is a prep for `clang-tidy` then this is fine. I saw something like this used for rtags, which was an editor-specific thing, so that's why I asked...

andrsd (approved):

zhardy-lanl (approved):

PR #265, opened 2024-05-13
Removing unused field-function operations

wdhawkins:
None of these operations are used in our regression tests. I'm struggling to think of a use for them, but I haven't used the code as much as @zachhardy and @ragusa. I definitely think `partitioner_predicate` can be removed, but I'd like to get everyone's thoughts on the others.


ragusa:
regarding the`partitioner_predicate`. Don't we need to to partition the mesh in serial and then reload it in parallel? What was the issue we had with high-core count KBA again?

Regarding the multi field functions: I am not sure, we seem to be exporting with fieldfunc.ExportToVTKMulti. I am not sure sure what the Milti in that name stands for (it's a new name, maybe @andrsd knows). But more generally, what is a multi field function? I see we can export multigroup fluxes, so that's not an issue.

Operations on field functions: I suppose operation will not be used to compute reaction rates (e.g., dose rate, where the KERMA XS is given in the material). So, if RR are never computed though FF, I am a bit lost at what the multiFF are for.

wdhawkins:
> regarding the`partitioner_predicate`. Don't we need to to partition the mesh in serial and then reload it in parallel? What was the issue we had with high-core count KBA again?
> 
> Regarding the multi field functions: I am not sure, we seem to be exporting with fieldfunc.ExportToVTKMulti. I am not sure sure what the Milti in that name stands for (it's a new name, maybe @andrsd knows). But more generally, what is a multi field function? I see we can export multigroup fluxes, so that's not an issue.
> 
> Operations on field functions: I suppose operation will not be used to compute reaction rates (e.g., dose rate, where the KERMA XS is given in the material). So, if RR are never computed though FF, I am a bit lost at what the multiFF are for.

Based on your comments and what little I gleaned from the code, I believe multi-field operations allow the user to define a ff that is a function of other ff's. So, it would fit perfectly with your dose calculation example. It looks to me like we should keep copy and multi-field. I think the partitioner-predicate was an attempt to visualize the partitioning of high-core count runs in serial. I don't think it ever worked, and it can likely go.

andrsd:
I don't think `fieldfunc.ExportToVTKMulti` is new. I remember seeing it from the beginning of `opensn`.

And if I am reading the code correctly, it just stores multiple field functions into one file (which is what the doco says as well).

framework/field_functions/operations/field_copy.cc
zhardy-lanl, line 1: As an example for how this would be used, consider a world where we passed information around via field functions. In this scenario, this type of operation would be quite useful for acceleration schemes. You would copy the scalar flux components of the flux moments vector into a scalar flux vector for a diffusion solver and vice verse. Since we don't operate this way and won't any time soon, this is fine to get rid of.

zhardy-lanl (approved):
I don't see an issue in removing these. If we move back towards using field functions after the LBS refactor, we may have to reimplement some things in different forms, but that's okay.

ragusa (approved):

andrsd (approved):

PR #266, opened 2024-05-15
Materials Lua Interface Refactor

### LINKED ISSUE #252: Remove `mat.AddProperty` Lua API and Refactor `mat.SetProperty`

zhardy-lanl:
This PR is a follow-up to #257 that continues the Lua materials interface refactor.

- Removes `mat.AddProperty` from the Lua interface
- Eliminates material property name options within `mat.SetProperty`
- Only allows one property of a given type per material

Closes #252

### DIFF
--- a/test/modules/linear_boltzmann_solvers/transport_steady/reed_balance.lua
+++ b/test/modules/linear_boltzmann_solvers/transport_steady/reed_balance.lua
@@ -39,15 +39,11 @@ for imat = 1, Nmat do materials[imat] = mat.AddMaterial(mat_names[imat]) end
 total = {50., 5., 0., 1., 1.}
 c = {0., 0., 0., 0.9, 0.9}
 for imat = 1, Nmat do
-  mat.AddProperty(materials[imat], TRANSPORT_XSECTIONS)
   mat.SetProperty(materials[imat], TRANSPORT_XSECTIONS, SIMPLE_ONE_GROUP, total[imat], c[imat])
 end
 
 -- Create sources in 1st and 4th materials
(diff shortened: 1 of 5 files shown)


test/modules/linear_boltzmann_solvers/transport_steady/reed_balance.lua
```diff
@@ -1,83 +1,79 @@
...
-  for i = 1, nrefs[imat] do
-    counter = counter + 1
-    nodes[counter] = nodes[counter - 1] + dx
-  end
+    dx = widths[imat] / nrefs[imat]
+    for i = 1, nrefs[imat] do
+        counter = counter + 1
+        nodes[counter] = nodes[counter - 1] + dx
+    end
 end
```
ragusa, line 18: is the StyLua causing this? Just curious. Because I know I pushed the test case before StyLua (when LuaFormatter was the ephemerous option)

ragusa (approved):

test/modules/linear_boltzmann_solvers/transport_steady/reed_balance.lua
andrsd: My guess would be that it is the CLion plugin @zhardy-lanl is using

andrsd (approved):

wdhawkins (approved):

Issue #98, opened 2024-01-31
Uniformization of object creation in lua layer


ragusa:
objects should be created through the object factory using the .create method and modification of those objects should work with a .method.

ragusa:
lua ia gfoing away

PR #270, opened 2024-05-16
Updating `CFEMSolver` to use InputParameters construction

andrsd:
This updated the `CFEMSolver` so it has same programming model as the `LBSSolver`, i.e.:
- construction via InputParameters
- options set via `SetOptions`
- boundary conditions set via `boundary_conditions` parameters in an `options` block.

Refs #98


test/modules/cfem_diffusion/c_diffusion_2d_1a_linear.lua
```diff
@@ -35,27 +35,45 @@ w_vol = logvol.RPPLogicalVolume.Create({xmin=-1000.0,xmax=-0.99999, infy=true, i
...
 s_vol = logvol.RPPLogicalVolume.Create({ymin=-1000.0,ymax=-0.99999, infx=true, infz=true})
 
-e_bndry = 0
-w_bndry = 1
-n_bndry = 2
-s_bndry = 3
+e_bndry = "0"
+w_bndry = "1"
+n_bndry = "2"
+s_bndry = "3"
```
andrsd, line 41: These are supposed to be strings, the C++ code requires `std::string`. This worked since the conversion from `1` to `"1"` was done by the lua layer, but it started to fail when I switched to `InputParameters`.

lua/modules/cfem_diffusion/lua_cfem_diff_solver.cc
zhardy-lanl, line 1: We should consider removing the `lua` prefixes from files in the `lua` directories.

zhardy-lanl (approved):

lua/modules/cfem_diffusion/lua_cfem_diff_solver.cc
ragusa: I agree with @zhardy-lanl . @andrsd , should either one of us create an issue for this and tag it as V1.0?

andrsd: Opened issue #275 for this

wdhawkins (approved):

andrsd:
I just noticed I forgot to remove `lua_cfem_diff_solver.h` which is no longer needed with these changes. I will update his PR momentarily.

ragusa (commented):
I was wondering why the namespace was changed from cfemdiffusion to diffusion. Will dfemdiffusion also be changed to the namespace diffusion and can this create conflict, now that the solver names also not longer have CFEM or DFEM in their names?

zhardy-lanl:
> I was wondering why the namespace was changed from cfemdiffusion to diffusion. Will dfemdiffusion also be changed to the namespace diffusion and can this create conflict, now that the solver names also not longer have CFEM or DFEM in their names?

I believe we decided to go to `diffusion::CFEMSolver` instead of `cfemdiffusion::Solver` awhile back, no? An analog should be done with DFEM diffusion. I think this is a step towards flattening the namespaces out and reducing name clashes.

ragusa:
Thanks. It was not easy to read everything on a phone screen. My previous comment was wrong. Thanks for the clarification.

ragusa (approved):

andrsd:
> Will `dfem_diffusion` also be changed to the namespace `diffusion`

Yes. I already have code for that. And this will be done for the other diffusion solvers as well to match the lua module/namespace.

> I believe we decided to go to `diffusion::CFEMSolver` instead of `cfemdiffusion::Solver` awhile back, no?

Yes, this is the plan. Have a `diffusion` C++ namespace with `{CFEM|DFEM|XYZ}Solver` and then matching syntax in the lua layer, i.e. `diffusion.CFEMSolver.Create()`, etc.

PR #272, opened 2024-05-17
Updating C5G7 tests. Adding weekly test workflow.

wdhawkins:
This PR fixes an issue with the test script for problems that support additional lua arguments. It also adds a github actions workflow for long-running/moderately-sized weekly tests.


test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_jfnk.lua
```diff
@@ -0,0 +1,98 @@
...
+        power_normalization = 1.0,
+        save_angular_flux = true
+    },
+    sweep_type = "AAH"
+})
+
+-- Execute Solver
+if (k_method == "pi") then
+    k_solver = lbs.XXPowerIterationKEigen.Create
+    ({
```
ragusa, line 47: still XX in name?

ragusa (approved):

ragusa:
@wdhawkins I am curious on the compute time between CBC and AAH. Can you tell me the difference, when running on the same machine? Thanks

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_jfnk.lua
wdhawkins: Yes, the other PR hasn't been merged yet. When it is, I'll rebase and update this one.

wdhawkins:
> @wdhawkins I am curious on the compute time between CBC and AAH. Can you tell me the difference, when running on the same machine? Thanks

For PI, AAH is ~30% faster than CBC (200 ranks, Cascade Lake, OmniPath). CBC takes 132 its. AAH takes 173 its with 2% lagged unknowns. For JFNK, CBC is ~20% faster than AAH. CBC takes 7 outers  vs 19 for AAH.

andrsd (approved):

zhardy-lanl (changes requested):
It looks like each of the test files are identical and that the only difference is that in the .json file you specify the `k_method` argument. Do we need a file for each if this is the case?

wdhawkins:
> It looks like each of the test files are identical and that the only difference is that in the .json file you specify the `k_method` argument. Do we need a file for each if this is the case?

Yeah, not a fan of this either. Unfortunately, it looks like the test script requires a unique file name for each test... even if the Lua args differ. I'll open an issue for this problem. In the meantime, I wanted to get these tests running, so I just created a separate input for each case.

zhardy-lanl (approved):

PR #273, opened 2024-05-21
Adding Python linting

andrsd:
- Fixing exception handling
- Adding workflow for python linting
- Python files are flake8 clean

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,7 +11,7 @@ max_line_length = 100
 
 [*.py]
 indent_size = 4
-max_line_length = 79
+max_line_length = 100
 
 [*.rst]
 indent_size = 3
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+[flake8]
+ignore = E203, E266, W503, C901, E741
+max-line-length = 100
+max-complexity = 18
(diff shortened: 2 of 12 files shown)


andrsd:
This is adding lining for python. I am using `flake8` with some general settings. On user side all one has to do is type `flake8` in the source dir. The linter will pickup settings from `.flake8` file and if problems are found you will see something like this:
```
./test/src/checks.py:532:80: W504 line break after binary operator
```

The maximum line length is set to `100` to match the C++ setting. This can be changed, originally it was set to 79, which seems a little too narrow to me.

You will see a lot of changes like this:
```
some long stuff +
more long stuff
```
changed into 
```
some long stuff
+ more long stuff
```
There is a [SO](https://stackoverflow.com/questions/57074300/what-is-the-recommended-way-to-break-a-long-if-statement-w504-line-break-after) talking about this becoming a best practice. so I used that. If there is a strong opinion on using the original syntax, W504 can be ignored.

Feel free to suggest improvements!

zhardy-lanl (approved):

ragusa (approved):

wdhawkins (approved):

PR #277, opened 2024-05-26
Groupset solver doesn't honor GMRES restart parameter

wdhawkins:
The GMRES restart parameter specified in groupset options was never assigned to the corresponding member variable, leading to the solver always using the default value of 30.


andrsd (approved):

ragusa (approved):
A long time ago, I was playing with solver options and I felt that some choices I made were not affecting the number of iterations. Good catch!!!

zhardy-lanl (approved):

PR #279, opened 2024-05-26
Fixing weekly tests to run only one day a week and not for PRs

wdhawkins:


.github/workflows/weekly.yaml
```diff
@@ -2,13 +2,9 @@ name: Weekly Tests
 
 on:
   schedule:
-    - cron: "0 6 * * *"
+    - cron: "0 6 * * 0"
   push:
     branches: [main]
-  workflow_dispatch:
```
andrsd, line 8: Do you want to keep this, so you can trigger the workflow manually?

ragusa (approved):

.github/workflows/weekly.yaml
wdhawkins: I thought it might be handy for testing/debugging purposes.

andrsd: The `workflow_dispatch:` line enables the manual trigger , see [doco](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch)

zhardy-lanl (approved):

.github/workflows/weekly.yaml
wdhawkins: Fixed.

andrsd (approved):

PR #280, opened 2024-05-26
Adding basic restart capability

wdhawkins:
The original restart capability was only available for classic Richardson iteration and was lost when that iterative algorithm was removed. This PR refactors and extends the original restart capability... with a few caveats. 

Restart dumps (when enabled) are now triggered from the AGS algorithm. Only the previous phi iterate is written to restart files. This works well for problems with no lagged fluxes. It also works when using the CBC sweeper (+1 for CBC). In all other cases, the restart provides a good starting guess for phi, but the code will need to iterate to converge lagged fluxes.  Restart files are now hdf5-based. Restart dumps (when enabled) are now time-based, but always triggered upon convergence. It is possible to disable the AGS restart capability so that a solver can implement its own custom restart read/write routines (see PowerIterationKEigen() for an example).


modules/linear_boltzmann_solvers/executors/pi_keigen.cc
```diff
@@ -182,5 +191,78 @@ PowerIterationKEigen::SetLBSScatterSource(const std::vector<double>& input,
...
+  // Check if enough time has elapsed since the last restart write
+  auto& last_restart_write_time = lbs_solver_.GetLastRestartWriteTime();
+  auto now = std::chrono::system_clock::now().time_since_epoch();
+  size_t now_secs = std::chrono::duration_cast<std::chrono::seconds>(now).count();
+  if (((now_secs - last_restart_write_time) < write_restart_time_interval) and not force)
+    return;
+
+  std::string fbase = lbs_solver_.Options().write_restart_directory_name + "/" +
+                      lbs_solver_.Options().write_restart_file_stem;
+  std::string fname = fbase + std::to_string(opensn::mpi_comm.rank()) + ".r";
```
andrsd, line 211: One thing I like to do when using hdf5, is the use of double extension, so here you could do `.r.h5`, or `.restart.h5`.
`.r` seems to be reserved for [R programming language](https://en.wikipedia.org/wiki/R_(programming_language)).

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
```diff
@@ -17,11 +17,8 @@ namespace lbs
 typedef std::vector<size_t> DirIDs; ///< Direction-IDs
 typedef std::vector<DirIDs> UniqueSOGroupings;
 typedef std::map<size_t, size_t> DirIDToSOMap;
-
-typedef std::vector<double> VecDbl;
-typedef std::vector<VecDbl> MatDbl;
-typedef std::vector<Vector3> VecVec3;
-typedef std::vector<VecVec3> MatVec3;
+// typedef std::vector<Vector3> VecVec3;
+typedef std::vector<std::vector<Vector3>> MatVec3;
```
zhardy-lanl, line 21: Not technically within scope, but a good opportunity to switch to `using` or to stop using these custom types.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -552,25 +531,19 @@ LBSSolver::BoundaryOptionsBlock()
...
-                                           "Required only if `type` is `\"isotropic\"`. "
-                                           "An array of isotropic strength per group");
-
-  params.AddOptionalParameter(
-    "function_name",
-    "",
-    "Text name of the lua function to be called for this boundary condition."
-    "For more on this boundary condition type.");
-
+                                           "Required only if `type` is `\"isotropic\"`. An array "
```
andrsd, line 539: I would suggest to use either \` or `"` for the 'isotropic' type.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -797,17 +789,17 @@ LBSSolver::Initialize()
 {
   CALI_CXX_MARK_SCOPE("LBSSolver::Initialize");
 
-  PerformInputChecks(); // a assigns num_groups and grid
-  PrintSimHeader();     // b
+  PerformInputChecks(); // assigns num_groups and grid
```
andrsd, line 792: If `PerformInputChecks` really assigns num_groups and grid then it is either a) misnomer or b) bad design.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -1996,140 +1985,72 @@ LBSSolver::DisAssembleTGDSADeltaPhiVector(const LBSGroupset& groupset,
...
-      {
-        log.Log0Warning() << "Failed to create restart directory: " << folder_name;
-        return;
-      }
-  }
+  // Check if enough time has elapsed since the last restart write
+  auto now = std::chrono::system_clock::now().time_since_epoch();
+  size_t now_secs = std::chrono::duration_cast<std::chrono::seconds>(now).count();
+  if (((now_secs - last_restart_write_time_) < options_.write_restart_time_interval) and not force)
+    return;
```
zhardy-lanl, line 2000: See previous comment.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -1996,140 +1985,72 @@ LBSSolver::DisAssembleTGDSADeltaPhiVector(const LBSGroupset& groupset,
...
 
-  opensn::mpi_comm.barrier();
+  std::string fbase =
+    options_.write_restart_directory_name + "/" + options_.write_restart_file_stem;
+  std::string fname = fbase + std::to_string(opensn::mpi_comm.rank()) + ".r";
```
zhardy-lanl, line 2004: See previous comment.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -1996,140 +1985,72 @@ LBSSolver::DisAssembleTGDSADeltaPhiVector(const LBSGroupset& groupset,
...
 
 void
-LBSSolver::ReadRestartData(const std::string& folder_name, const std::string& file_base)
+LBSSolver::ReadRestartData()
 {
   CALI_CXX_MARK_SCOPE("LBSSolver::ReadRestartData");
 
-  opensn::mpi_comm.barrier();
+  std::string fbase = options_.read_restart_directory_name + "/" + options_.read_restart_file_stem;
+  std::string fname = fbase + std::to_string(opensn::mpi_comm.rank()) + ".r";
```
zhardy-lanl, line 2034: See previous comment.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -1996,140 +1985,72 @@ LBSSolver::DisAssembleTGDSADeltaPhiVector(const LBSGroupset& groupset,
...
+  std::string fbase =
+    options_.write_restart_directory_name + "/" + options_.write_restart_file_stem;
+  std::string fname = fbase + std::to_string(opensn::mpi_comm.rank()) + ".r";
 
-  // Create files
-  // This step might fail for specific locations and
-  // can create quite a messy output if we print it all.
-  // We also need to consolidate the error to determine if
-  // the process as whole succeeded.
+  // Wfrite data
```
andrsd, line 2006: Typo "Wfrite" => "Write"

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
```diff
@@ -17,11 +17,8 @@ namespace lbs
 typedef std::vector<size_t> DirIDs; ///< Direction-IDs
 typedef std::vector<DirIDs> UniqueSOGroupings;
 typedef std::map<size_t, size_t> DirIDToSOMap;
-
-typedef std::vector<double> VecDbl;
-typedef std::vector<VecDbl> MatDbl;
-typedef std::vector<Vector3> VecVec3;
-typedef std::vector<VecVec3> MatVec3;
+// typedef std::vector<Vector3> VecVec3;
```
andrsd, line 20: Delete this line

test/modules/linear_boltzmann_solvers/transport_steady/simple_scatter_sgt1000_sigs999.9.xs
```diff
@@ -0,0 +1,13 @@
+# Exported cross section from ChiTech
```
andrsd, line 1: ChiTech => OpenSn.

It this still in the code somewhere or just produced with and older version of the code?

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -442,39 +432,38 @@ LBSSolver::OptionsBlock()
...
   params.AddOptionalParameter(
     "read_restart_data", false, "Flag indicating whether restart data is to be read.");
+  params.AddOptionalParameter("read_restart_directory_name",
+                              opensn::input_path.stem().string() + "_restart",
+                              "Directory name to use when reading restart data.");
+  params.AddOptionalParameter("read_restart_file_stem",
+                              opensn::input_path.stem().string(),
+                              "File stem to use when reading restart data.");
+  params.AddOptionalParameter("write_restart_directory_name",
+                              opensn::input_path.stem().string() + "_restart",
```
zhardy-lanl, line 453: Does this directory get created if it does not exist?

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_2_unstructured_restart.lua
```diff
@@ -0,0 +1,153 @@
...
+--    slices[k] = fieldfunc.FFInterpolationCreate(SLICE)
+--    fieldfunc.SetProperty(slices[k],SLICE_POINT,{x = 0.0, y = 0.0, z = 0.8001})
+--    fieldfunc.SetProperty(slices[k],ADD_FIELDFUNCTION,fflist[k])
+--    --fieldfunc.SetProperty(slices[k],SLICE_TANGENT,{x = 0.393, y = 1.0-0.393, z = 0})
+--    --fieldfunc.SetProperty(slices[k],SLICE_NORMAL,{x = -(1.0-0.393), y = -0.393, z = 0.0})
+--    --fieldfunc.SetProperty(slices[k],SLICE_BINORM,{x = 0.0, y = 0.0, y = 1.0})
+--    fieldfunc.Initialize(slices[k])
+--    fieldfunc.Execute(slices[k])
+--    fieldfunc.ExportToPython(slices[k])
+--end
```
andrsd, line 124: Maybe remove this commented out section altogether

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_2_unstructured_restart.lua
```diff
@@ -0,0 +1,153 @@
...
+if (master_export == nil) then
+  fieldfunc.ExportToVTKMulti(fflist,"ZPhi3D")
+end
+
+--############################################### Plots
+if (location_id == 0 and master_export == nil) then
+
+  --os.execute("python ZPFFI00.py")
+  ----os.execute("python ZPFFI11.py")
+  --local handle = io.popen("python ZPFFI00.py")
```
andrsd, line 151: These 3 lines should be removed

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -442,39 +432,38 @@ LBSSolver::OptionsBlock()
...
+                              32768,
                               "The maximum MPI message size used during sweep initialization.");
   params.AddOptionalParameter(
     "read_restart_data", false, "Flag indicating whether restart data is to be read.");
+  params.AddOptionalParameter("read_restart_directory_name",
+                              opensn::input_path.stem().string() + "_restart",
+                              "Directory name to use when reading restart data.");
+  params.AddOptionalParameter("read_restart_file_stem",
+                              opensn::input_path.stem().string(),
+                              "File stem to use when reading restart data.");
```
zhardy-lanl, line 451: This could really all be encapsulated in a single attribute called `read_restart_dump` which could be defaulted to an empty string (no restart read), or specified as `<dir>/<stem>`.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -442,39 +432,38 @@ LBSSolver::OptionsBlock()
...
-    "write_restart_data", false, "Flag indicating whether restart data is to be written.");
-  params.AddOptionalParameter(
-    "write_restart_folder_name", "YRestart", "Folder name to use when writing restart data.");
-  params.AddOptionalParameter(
-    "write_restart_file_base", "restart", "File base name to use when writing restart data.");
-  params.AddOptionalParameter(
-    "write_restart_interval",
-    30.0,
-    "Interval at which restart data is to be written. Currently not implemented.");
+    "write_restart_file_stem", "restart", "File stem to use when writing restart data.");
```
zhardy-lanl, line 456: This could be encapsulated into a single variable called something like `restart_dump_path` which could have the same form as `read_restart_dump` above.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -442,39 +432,38 @@ LBSSolver::OptionsBlock()
...
+  params.AddOptionalParameter("write_restart_time_interval",
+                              60,
+                              "Time interval in seconds at which restart data is to be written.");
+  params.ConstrainParameterRange("write_restart_time_interval", AllowableRangeLowLimit::New(60));
```
zhardy-lanl, line 460:
```suggestion
  params.AddOptionalParameter("restart_dump_frequency",
                              60,
                              "Frequency in seconds to write restart dumps.");
  params.ConstrainParameterRange("write_restart_time_interval", AllowableRangeLowLimit::New(60));
```
Should we default to writing restart dumps at the smallest allowable interval?

zhardy-lanl (changes requested):
A few suggestions on input parameters and separating some pieces of code. Let me know your thoughts. This looks like a step in the right direction.

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
zhardy-lanl: I would lean towards `.restart.h5`.

wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
wdhawkins: Fixed.

framework/utils/hdf_utils.h
```diff
@@ -5,109 +5,165 @@
...
 }
 
+inline bool
+H5CreateGroup(hid_t id, const std::string& name)
+{
+  return (H5Gcreate2(id, name.c_str(), H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT) >= 0);
+}
+
+template <typename T>
+bool
```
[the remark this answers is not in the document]
andrsd: If a templated function is instantiated 2 or more times, then we hit the duplicate symbol error when linking. Specifying the function with `inline` fixes the problem. This function is most likely instantiated only once, thus it does not produce the duplicate symbol error, but technically it should be also marked with `inline`.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: I think it 's a bit of both. It looks like it does initialization and checks for various quantities. I'd like to invest more time into refactoring this entire class, so I think I'll defer this issue until after 1.0.

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_2_unstructured_restart.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_2_unstructured_restart.lua
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
andrsd: Ok, no problem :+1:

framework/utils/hdf_utils.h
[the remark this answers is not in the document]
wdhawkins: To start with all of these methods should have been inline to begin with... I just got lucky that this header file was only included once in my initial commit. The rule is that explicit template instantiations are definitions and are subject to the one-definition rule (basically, they can only appear once in a program). So, that means that they must either be in-line as I've done here, or the prototype declared in the header and the definition in a source file. The H5WriteDataset1D method is not a specialization, so it's treated like any other template and should appear in the header.

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
```diff
@@ -77,18 +81,19 @@ PowerIterationKEigen::Initialize()
 
   OpenSnLogicalErrorIf(not front_wgs_context_, ": Casting failed");
 
-  if (reset_phi0_)
+  if (reset_phi0_ && (not lbs_solver_.Options().read_restart_data))
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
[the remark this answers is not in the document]
wdhawkins: I originally implemented something similar to this, but ultimately decided that having the individual pieces of the path available made for better error messages.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Yes, it's created in `lbsSolver::SetOptions()` if `write_restart_time_interval` is set.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Good point. Maybe something like every 10 minutes?

zhardy-lanl: That depends... Are restarts always on? If so, 10 minutes seems okay. If not, then I would default to 0 (get rid of the minimum) and check that it is non-zero whenever restarts are enabled (however that is done).

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -646,23 +628,33 @@ LBSSolver::SetOptions(const InputParameters& params)
...
+        if (opensn::mpi_comm.rank() == 0)
+          std::filesystem::create_directory(options_.write_restart_directory_name);
+        opensn::mpi_comm.barrier();
+        if (not std::filesystem::is_directory(options_.write_restart_directory_name))
+        {
+          throw std::runtime_error("Failed to create restart directory " +
+                                   options_.write_restart_directory_name);
+        }
+      }
+    }
```
zhardy-lanl, line 657: This seems to be more appropriate with `write_restart_directory_name`, FYI.

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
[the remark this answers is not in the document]
zhardy-lanl: Could you point me to one of those error messages? If we use something like `std::filesystem::path` it should be trivial to get the info you need for any error message.

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
```diff
@@ -182,5 +191,78 @@ PowerIterationKEigen::SetLBSScatterSource(const std::vector<double>& input,
...
+  auto write_restart_time_interval = lbs_solver_.Options().write_restart_time_interval;
+  if (write_restart_time_interval == 0)
+    return;
+
+  // Check if enough time has elapsed since the last restart write
+  auto& last_restart_write_time = lbs_solver_.GetLastRestartWriteTime();
+  auto now = std::chrono::system_clock::now().time_since_epoch();
+  size_t now_secs = std::chrono::duration_cast<std::chrono::seconds>(now).count();
+  if (((now_secs - last_restart_write_time) < write_restart_time_interval) and not force)
+    return;
```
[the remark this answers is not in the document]
wdhawkins: I've created a few small helper methods that abstract some of the boilerplate code.  I agree that it makes things much cleaner and easier to understand.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: I defaulted to every 10 minutes. That seems like a reasonable value. Presumably restart dumps will be enabled for mostly long-running problems.

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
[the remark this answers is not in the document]
wdhawkins: That's a very good point. Let me see about fixing this.

ragusa: Same as Zach

wdhawkins: Agreed. I have a new version coming up shortly. All the comments have been great, and I like the new version much better.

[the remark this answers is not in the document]
wdhawkins: I've removed most of the restart parameters for files and paths in favor of `read_restart_path` and `write_restart_path` similar to suggestions by @zhardy-lanl below. I didn't create a `GetRestartDumpPath()` method so that individual solvers can tailor the path for their needs, but it's still a much cleaner interface.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed.

andrsd (approved):

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -742,6 +702,18 @@ LBSSolver::SetOptions(const InputParameters& params)
       }
     }
   } // for p
+
+  if (options_.write_restart_time_interval > 0)
```
zhardy-lanl, line 706: If `write_restart_path` is not specified, where would restarts be written to?

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -742,6 +702,18 @@ LBSSolver::SetOptions(const InputParameters& params)
...
+  {
+    auto dir = options_.write_restart_path.parent_path();
+    {
+      if (opensn::mpi_comm.rank() == 0)
+        std::filesystem::create_directory(dir);
+      opensn::mpi_comm.barrier();
+      if (not std::filesystem::is_directory(dir))
+        throw std::runtime_error("Failed to create restart directory " + dir.string());
+    }
```
zhardy-lanl, line 715: Does this block need to be in a separate scope?

zhardy-lanl (changes requested):
Looks way better. Just a few clarification questions and a tiny change suggestions.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: By separate scope, do you mean moved it's own method, or just moving it to a separate scope in-place? I could see moving it to it's own method, but it's fairly small and likely only needed here.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Nice catch. I forgot to go back and set the default value for `write_restart_path`. I'll fix that.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
zhardy-lanl: Lines 710-714 are within curly braces that are not part of an if statement or anything of the sorts. That's what I am referring to. I forget what the correct terminology for that is.

wdhawkins: That is the correct terminology, and I just missed it looking at the excerpt above. Fixed!  Thanks!

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed!

zhardy-lanl (approved):

ragusa (approved):

modules/linear_boltzmann_solvers/executors/pi_keigen.cc
```diff
@@ -182,5 +191,78 @@ PowerIterationKEigen::SetLBSScatterSource(const std::vector<double>& input,
...
+
+  bool global_succeeded = true;
+  mpi_comm.all_reduce(location_succeeded, global_succeeded, mpi::op::logical_and<bool>());
+  if (global_succeeded)
+  {
+    log.Log() << "Successfully wrote restart data to " << fbase << "X.r";
+    last_restart_write_time = now_secs;
+  }
+  else
+    log.Log0Error() << "Failed to write restart data to " << fbase << "X.r";
```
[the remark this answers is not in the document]
wdhawkins: I considered that, but I decided against terminating the run if we couldn't write a restart.

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
wdhawkins: I switched these to `using`. I'll have another PR that addresses `typedef` in general.

PR #281, opened 2024-05-30
Removing lua prefix/postfix from files in the lua layer

### LINKED ISSUE #275: Remove `lua` prefix/postfix from file names in the lua layer
Do it!

andrsd:
Closes #275


lua/framework/math/functions/lua_dim_a_to_dim_b.cc
zhardy-lanl, line 1: Should this be `function_dim_a_to_dim_b`?

lua/framework/math/functions/lua_dim_a_to_dim_b.h
zhardy-lanl, line 1: Should this be `function_dim_a_to_dim_b`?

zhardy-lanl (changes requested):

wdhawkins (approved):

lua/framework/math/functions/lua_dim_a_to_dim_b.h
andrsd: No, because the class name is `LuaDimAToDimB`. The parent class is `FunctionDimAToDimB`.

zhardy-lanl (approved):

lua/framework/logging/log.cc
```diff
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
 // SPDX-License-Identifier: MIT
 
-#include "lua_log.h"
+#include "log.h"
 #include "framework/logging/log.h"
```
ragusa, line 5: could it be confusing to have 2 files named log.h in different folders? wouldn't it be better to have different filenames for clarity??

andrsd: This should actually use the full path "lua/framework/logging/log.h", that would avoid the potential confusion.
Our search path for header files is set to the repo root, so compiler will always pick up the right file here, i.e. the one from the active directory.
And yes, situations like this were probably the driver for the lua prefix/postfix.

ragusa: @zhardy-lanl @wdhawkins @andrsd : so what do you think about filenames (.cc or .h) being the same in the lua/ and in the rest of the src code?

zhardy-lanl: I agree with @andrsd that the full path should be used to avoid confusion and am in full support of removing `_lua` or `lua_` from file names. Things in the Lua folder should be assumed to be related to Lua.

ragusa: Ok. So is this PR making sure that the full path is used or should there be a new issue?

andrsd: This PR is about removing the lua prefix from file names. You just happen to spot another problem that needs to be fixed...

ragusa: created isse here: #283 
will resolve this conversation and approve this PR now

ragusa (approved):

PR #255, opened 2024-05-07
Changing boundary ID to boundary name mapping in orthogonal mesh generator

andrsd:
Closes #229

PR #284, opened 2024-06-02
Fix boundary id mapping

wdhawkins:
This PR fixes an issue with boundary id mapping in lbs_solver. The global id mapping was changed in #255, but a routine in lbs_solver was still using hardcoded values corresponding to the pre-#255 mapping. In fixing this issue, I've tried to bring some consistency to the boundary id mapping (even in places where it's not technically necessary). 

With this fix, all of the regression inputs that were modified in #255 needed to be reverted to their pre-#255 state.


framework/mesh/mesh.h
```diff
@@ -61,6 +61,16 @@ enum MeshAttributes : int
   UNSTRUCTURED = (1 << 5)
```
ragusa, line 61: @wdhawkins @andrsd @zhardy-lanl  : I thought we had a discussion about bitwise operator for mesh enum and that we wanted to get rid of them (but we would keep the bitwise ops for source specifications). I just saw this here by chance. Should be create an issue?

ragusa (approved):

framework/mesh/mesh.h
wdhawkins: Done. See #286.

zhardy-lanl (approved):
Thank you for tracking this down! So am I correct in assuming that all boundary conditions were backwards? I don't know how I missed that on #255, if so. The boundary ID changes should not have touched the input language at any point. It should have been assigning different IDs to boundary names under the hood only.

andrsd (approved):

wdhawkins:
> Thank you for tracking this down! So am I correct in assuming that all boundary conditions were backwards? I don't know how I missed that on #255, if so. The boundary ID changes should not have touched the input language at any point. It should have been assigning different IDs to boundary names under the hood only.

Yes, that's correct. Everything was backwards. I dug through the code pretty carefully, so I think I got everything this time.  🤞

PR #285, opened 2024-06-03
Use full path in #includes

### LINKED ISSUE #283: use full path for header files
This PR is about removing the lua prefix from file names. You just happen to spot another problem that needs to be fixed...

_Originally posted by @andrsd in https://github.com/Open-Sn/opensn/pull/281#discussion_r1623398987_

andrsd:
- Use full relative path from the repo root in #includes
- Don't include `lua` dir into include dir search

Closes #283

### DIFF
--- a/lua/CMakeLists.txt
+++ b/lua/CMakeLists.txt
@@ -7,7 +7,7 @@ target_include_directories(libopensnlua
     $<INSTALL_INTERFACE:include/opensn>
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
-    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}
     ${PROJECT_SOURCE_DIR}/external
     ${LUA_INCLUDE_DIR}
 )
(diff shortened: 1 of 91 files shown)


lua/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ target_include_directories(libopensnlua
     $<INSTALL_INTERFACE:include/opensn>
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
-    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}
```
andrsd, line 10: The reason for this change is that this added `<repo_root>/lua` into include path, so people could omit the `lua` prefix in the `#include`s. After this change they have to use the path from the repo_root, i.e. include the `lua` subdirectory.

zhardy-lanl (approved):
The only comment I have here is that in general I would be okay with using relative paths for *only* the associated .h file.

wdhawkins (approved):

andrsd:
> The only comment I have here is that in general I would be okay with using relative paths for _only_ the associated .h file.

This was exactly the case that put all this into motion :-)

image

And the [discussion](https://github.com/Open-Sn/opensn/pull/281#pullrequestreview-2089526484) that triggered this.

ragusa (approved):
I know it makes it feel pedantic but I think it avoids any confusion.

PR #287, opened 2024-06-04
Second Moment Method

zhardy-lanl:
This PR adds the second moment method as k-eigenvalue acceleration routine and includes a C5G7 test. The format is comparable to that of the source corrected DSA routine.

### DIFF
--- a/modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
+++ b/modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
@@ -129,8 +129,7 @@ PowerIterationKEigenSMM::Initialize()
       throw std::logic_error("Only vacuum and reflective boundaries are valid for "
                              "k-eigenvalue problems.");
 
-  const auto bcs = TranslateBCs(lbs_solver_.SweepBoundaries(),
-                                /*vacuum_bcs_are_dirichlet=*/false);
+  const auto bcs = TranslateBCs(lbs_solver_.SweepBoundaries(), false);
 
   // Create the diffusion materials
   const auto xs_map = PackGroupsetXS(
@@ -196,7 +195,7 @@ PowerIterationKEigenSMM::Execute()
   while (nit < max_iters_)
   {
     // Compute the transport 1/k fission source
-    SetLBSFissionSource(phi_ell, /*additive=*/false);
+    SetLBSFissionSource(phi_ell, false);
     Scale(q_moments_local_, 1.0 / k_eff_);
 
     // Solve some transport inners
@@ -238,7 +237,7 @@ PowerIterationKEigenSMM::Execute()
         // add in the second moment correction, and solve the system
         auto b = AssembleDiffusionRHS(Sf + Ss);
         diffusion_solver_->AddToRHS(b + correction);
-        diffusion_solver_->Solve(phi0, /*use_initial_guess=*/true);
+        diffusion_solver_->Solve(phi0, true);
 
         // Bump the new solution to old
         phi0_old = phi0;
--- a/modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
+++ b/modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
@@ -300,7 +300,7 @@ PowerIterationKEigenSMM::Execute()
   log.Log() << "\n";
   log.Log() << "        Final k-eigenvalue    :        " << std::setprecision(7) << k_eff_;
   log.Log() << "        Final change          :        " << std::setprecision(6) << k_eff_change
-            << " (num_TrOps:" << front_wgs_context_->counter_applications_of_inv_op_ << ")"
+            << " (Number of Sweeps:" << front_wgs_context_->counter_applications_of_inv_op_ << ")"
             << "\n";
 
   if (lbs_solver_.Options().use_precursors)
--- a/modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
+++ b/modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
@@ -98,12 +98,12 @@ PowerIterationKEigenSMM::Initialize()
   const auto num_groups = lbs_solver_.NumGroups();
   const auto num_gs_groups = front_gs_.groups_.size();
 
-  // Specialized unit cell matrices
+  // Specialized SMM data
   dimension_ = lbs_solver_.Grid().Dimension();
   ComputeAuxiliaryUnitCellMatrices();
+  ComputeBoundaryFactors();
 
   // Create PWLC structure, if needed
-
   if (diffusion_sdm_name_ == "pwlc")
   {
     pwlc_ptr_ = PieceWiseLinearContinuous::New(grid);
@@ -146,6 +146,7 @@ PowerIterationKEigenSMM::Initialize()
                                                        bcs,
                                                        xs_map,
                                                        lbs_solver_.GetUnitCellMatrices(),
+                                                       true,
                                                        diffusion_verbose_);
   }
   else
@@ -156,6 +157,7 @@ PowerIterationKEigenSMM::Initialize()
                                                         bcs,
                                                         xs_map,
                                                         lbs_solver_.GetUnitCellMatrices(),
+                                                        true,
                                                         diffusion_verbose_);
   }
 
@@ -178,6 +180,7 @@ PowerIterationKEigenSMM::Initialize()
     tmp.assign(pwld.GetNumLocalDOFs(diff_uk_man), 0.0);
   else
     tmp.assign(pwlc_ptr_->GetNumLocalAndGhostDOFs(diff_uk_man), 0.0);
+  AssembleDiffusionBCs();
   diff_solver->AssembleAand_b(tmp);
   log.Log() << "Done assembling diffusion system.";
 }
@@ -318,7 +321,6 @@ PowerIterationKEigenSMM::ComputeClosures(const std::vector<std::vector<double>>&
   const auto& grid = lbs_solver_.Grid();
   const auto& pwld = lbs_solver_.SpatialDiscretization();
   const auto& transport_views = lbs_solver_.GetCellTransportViews();
-  const auto num_groupsets = lbs_solver_.Groupsets().size();
 
   // Create a local tensor vector, set it to zero
   auto local_tensors = tensors_->MakeLocalVector();
@@ -329,10 +331,8 @@ PowerIterationKEigenSMM::ComputeClosures(const std::vector<std::vector<double>>&
   for (const auto& groupset : lbs_solver_.Groupsets())
   {
     const auto& psi_uk_man = groupset.psi_uk_man_;
-    const auto& quadrature = groupset.quadrature_;
-    const auto num_gs_angles = quadrature->omegas_.size();
-    const auto wt_sum =
-      std::accumulate(quadrature->weights_.begin(), quadrature->weights_.end(), 0.0);
+    const auto& quad = groupset.quadrature_;
+    const auto num_gs_dirs = quad->omegas_.size();
 
     const auto first_grp = groupset.groups_.front().id_;
     const auto num_gs_groups = groupset.groups_.size();
@@ -354,13 +354,13 @@ PowerIterationKEigenSMM::ComputeClosures(const std::vector<std::vector<double>>&
           const auto dof_map = pwld.MapDOFLocal(cell, i, tensor_uk_man_, g, 0);
           double* T = &local_tensors[dof_map];
 
-          // Perform the quadrature integration
-          for (int n = 0; n < num_gs_angles; ++n)
+          // Perform the quad integration
+          for (int d = 0; d < num_gs_dirs; ++d)
           {
-            const auto& omega = quadrature->omegas_[n];
-            const auto& wt = quadrature->weights_[n];
+            const auto& omega = quad->omegas_[d];
+            const auto& wt = quad->weights_[d];
 
-            const auto psi_dof = pwld.MapDOFLocal(cell, i, psi_uk_man, n, gsg);
+            const auto psi_dof = pwld.MapDOFLocal(cell, i, psi_uk_man, d, gsg);
             const auto coeff = wt * psi[gs][psi_dof];
 
             for (int k = 0; k < dimension_; ++k)
@@ -374,7 +374,7 @@ PowerIterationKEigenSMM::ComputeClosures(const std::vector<std::vector<double>>&
                 T[k * dimension_ + l] += coeff * omega[dim_idx_k] * omega[dim_idx_l];
               }
             }
-          } // for angle n
+          } // for direction d
         }   // for groupset group gsg
       }     // for node i
 
@@ -389,56 +389,34 @@ PowerIterationKEigenSMM::ComputeClosures(const std::vector<std::vector<double>>&
           {
             const auto i = cell_mapping.MapFaceNode(f, fi);
             const auto imap = pwld.MapDOFLocal(cell, i);
+            const auto bfac = bndry_factors_[imap][gs];
 
-            // Compute the diffusion boundary factor. This is simply the
-            // computation of the quadrature of \Omega \cdot \hat{n} divided
-            // by the sum of the quadrature weights. In most cases, this should
-            // be equivalent to half.
-            if (bndry_factors_.count(imap) == 0)
-            {
-              bndry_factors_[imap].resize(num_groupsets, 0.0);
-
-              double val = 0.0;
-              for (int n = 0; n < num_gs_angles; ++n)
-                val +=
-                  quadrature->weights_[n] * std::fabs(quadrature->omegas_[n].Dot(face.normal_));
-              bndry_factors_[imap][gs] = val / wt_sum;
-            }
-            //            const auto& bfac = bndry_factors_[imap][gs];
-
-            // Initialize corrections, if not done already
+            // Reset boundary closures
             betas_[imap].assign(lbs_solver_.NumGroups(), 0.0);
             auto& beta = betas_[imap];
 
-            // Compute the boundary corrections
+            // Compute the boundary closure
             for (int gsg = 0; gsg < num_gs_groups; ++gsg)
             {
               const auto g = first_grp + gsg;
 
               // Perform the quadrature integration
-              for (int n = 0; n < num_gs_angles; ++n)
+              for (int d = 0; d < num_gs_dirs; ++d)
               {
-                const auto& wt = quadrature->weights_[n];
-                const auto& omega = quadrature->omegas_[n];
+                const auto& wt = quad->weights_[d];
+                const auto& omega = quad->omegas_[d];
                 const auto mu = std::fabs(omega.Dot(face.normal_));
 
-                const auto psi_dof = pwld.MapDOFLocal(cell, i, psi_uk_man, n, gsg);
-
-                //                beta[g] += wt * (mu - bfac) * psi[gs][psi_dof];
-                beta[g] += wt * (mu - 0.5) * psi[gs][psi_dof];
-
-              } // for angle n
+                const auto psi_dof = pwld.MapDOFLocal(cell, i, psi_uk_man, d, gsg);
+                beta[g] += wt * (mu - bfac) * psi[gs][psi_dof];
+              } // for direction n
             }   // for groupset group gsg
           }     // for face node fi
         }
-
         ++f;
-
       } // for face
     }   // for cell
-
     ++gs;
-
   } // for groupset
 
   // Set the ghosted tensor vector with the local tensors, then
@@ -636,7 +614,9 @@ PowerIterationKEigenSMM::ComputeSourceCorrection() const
       // Boundary face terms
       if (not face.has_neighbor_)
       {
-        auto bc = diffusion_solver_->BCS().at(face.neighbor_id_);
+        BoundaryCondition bc;
+        if (diffusion_solver_->BCS().count(face.neighbor_id_) == 1)
+          bc = diffusion_solver_->BCS().at(face.neighbor_id_);
 
         // Contribute the boundary closure term to the diffusion source term.
         // This term is the sum of the integrated incident and outgoing
@@ -661,9 +641,7 @@ PowerIterationKEigenSMM::ComputeSourceCorrection() const
               {
                 const auto j = cell_mapping.MapFaceNode(f, fj);
                 const auto jmap = pwld.MapDOFLocal(cell, j);
-
                 val += betas_.at(jmap)[g] * face_M[i][j];
-
               } // for face node fj
 
               output.SetValue(imap, -val, VecOpType::ADD_VALUE);
@@ -677,6 +655,78 @@ PowerIterationKEigenSMM::ComputeSourceCorrection() const
 
   output.Assemble();
   return output.MakeLocalVector();
+}
+
+void
+PowerIterationKEigenSMM::AssembleDiffusionBCs() const
+{
+  const auto& grid = lbs_solver_.Grid();
+  const auto& pwld = lbs_solver_.SpatialDiscretization();
+  const auto& unit_cell_matrices = lbs_solver_.GetUnitCellMatrices();
+
+  const auto& diff_sd = diffusion_solver_->SpatialDiscretization();
+  const auto& diff_uk_man = diffusion_solver_->UnknownStructure();
+
+  const auto num_gs_groups = front_gs_.groups_.size();
+
+  // Loop over cells
+  std::vector<int64_t> rows;
+  std::vector<int64_t> cols;
+  std::vector<double> vals;
+  for (const auto& cell : grid.local_cells)
+  {
+    const auto& cell_mapping = pwld.GetCellMapping(cell);
+    const auto& fe_values = unit_cell_matrices[cell.local_id_];
+
+    // Loop over faces
+    int f = 0;
+    for (const auto& face : cell.faces_)
+    {
+      if (not face.has_neighbor_)
+      {
+        BoundaryCondition bc;
+        if (diffusion_solver_->BCS().count(face.neighbor_id_))
+          bc = diffusion_solver_->BCS().at(face.neighbor_id_);
+
+        if (bc.type == BCType::ROBIN)
+        {
+          // skip reflective bcs
+          if (std::fabs(bc.values[0]) < 1.0e-12)
+            continue;
+
+          const auto& face_M = fe_values.intS_shapeI_shapeJ[f];
+
+          const auto num_face_nodes = cell_mapping.NumFaceNodes(f);
+          for (int fi = 0; fi < num_face_nodes; ++fi)
+          {
+            const auto i = cell_mapping.MapFaceNode(f, fi);
+            const auto imap = diff_sd.MapDOF(cell, i, diff_uk_man, 0, 0);
+
+            for (int fj = 0; fj < num_face_nodes; ++fj)
+            {
+              const auto j = cell_mapping.MapFaceNode(f, fj);
+              const auto jmap = diff_sd.MapDOF(cell, j, diff_uk_man, 0, 0);
+              const auto bfac = bndry_factors_.at(pwld.MapDOFLocal(cell, j))[0];
+
+              for (int gsg = 0; gsg < num_gs_groups; ++gsg)
+              {
+                // This will not work if extended to multiple groupsets because in
+                // the current paradigm, one diffusion solver per groupset would be
+                // required.
+                rows.push_back(imap + gsg);
+                cols.push_back(jmap + gsg);
+                vals.push_back(bfac * face_M[i][j]);
+              }
+            } // for face node fj
+          }   // for face node fi
+        }     // if Robin boundary
+      }       // if boundary face
+      ++f;
+    } // for face
+  }   // for cell
+
+  // Add the contributions to the diffusion solver matrix
+  diffusion_solver_->AddToMatrix(rows, cols, vals);
 }
 
 std::vector<double>
@@ -839,6 +889,53 @@ PowerIterationKEigenSMM::ComputeAuxiliaryUnitCellMatrices()
   opensn::mpi_comm.barrier();
   if (accel_pi_verbose_)
     log.Log() << "Second moment method cell matrices computed.";
+}
+
+void
+PowerIterationKEigenSMM::ComputeBoundaryFactors()
+{
+  const auto& grid = lbs_solver_.Grid();
+  const auto& pwld = lbs_solver_.SpatialDiscretization();
+  const auto num_groupsets = lbs_solver_.Groupsets().size();
+
+  // Loop over groupsets
+  int gs = 0;
+  for (const auto& groupset : lbs_solver_.Groupsets())
+  {
+    const auto& quad = groupset.quadrature_;
+    const auto num_gs_dirs = quad->omegas_.size();
+    const auto wt_sum = std::accumulate(quad->weights_.begin(), quad->weights_.end(), 0.0);
+
+    // Loop over cells
+    for (const auto& cell : grid.local_cells)
+    {
+      const auto& cell_mapping = pwld.GetCellMapping(cell);
+
+      // Loop over faces
+      int f = 0;
+      for (const auto& face : cell.faces_)
+      {
+        const auto num_face_nodes = cell_mapping.NumFaceNodes(f);
+        for (int fi = 0; fi < num_face_nodes; ++fi)
+        {
+          const auto i = cell_mapping.MapFaceNode(f, fi);
+          const auto imap = pwld.MapDOFLocal(cell, i);
+
+          if (bndry_factors_.count(imap) == 0)
+            bndry_factors_[imap].resize(num_groupsets, 0.0);
+
+          // Compute the diffusion boundary factor, or the quadrature integral of
+          // |\Omega \cdot \hat{n}| divided by the sum of the quadrature weights.
+          double val = 0.0;
+          for (int d = 0; d < num_gs_dirs; ++d)
+            val += quad->weights_[d] * std::fabs(quad->omegas_[d].Dot(face.normal_));
+          bndry_factors_[imap][gs] = val / wt_sum;
+        }
+        ++f;
+      } // for face
+    }   // for cell
+    ++gs;
+  } // for groupset
 }
 
 void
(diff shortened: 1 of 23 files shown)


test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7.lua
```diff
@@ -74,6 +83,30 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwld"
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 109: same as above. there could be other occurences

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7.lua
```diff
@@ -74,6 +83,30 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwlc",
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 97: is it possible to bring these two lines outside of the if statements?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_jfnk.lua
```diff
@@ -74,6 +83,32 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwld"
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 111: same as before. in addition, why are the instructions repeated here?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi.lua
```diff
@@ -74,6 +83,30 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwlc",
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 97: same as before

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi.lua
```diff
@@ -74,6 +83,30 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwld"
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 109: same as before

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_scdsa.lua
```diff
@@ -74,6 +83,30 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwlc",
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 97: same as before

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_scdsa.lua
```diff
@@ -74,6 +83,30 @@ elseif (k_method == "pi_scdsa_pwlc") then
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwld"
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 109: same as before

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+        k_solver = lbs.PowerIterationKEigenSMM.Create
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwld"
+        })
+        solver.Initialize(k_solver)
```
ragusa, line 108: why is this indented?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwlc",
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 97: move outside of if statement?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+
+-- Execute Solver
+if (k_method == "pi") then
+    k_solver = lbs.PowerIterationKEigen.Create
+    ({
+        lbs_solver_handle = phys1,
+        k_tol = 1.0e-8
+    })
+    solver.Initialize(k_solver)
+    solver.Execute(k_solver)
```
ragusa, line 61: move outside of if statement?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+    ({
+        lbs_solver_handle = phys1,
+        diff_accel_sdm = "pwld",
+        accel_pi_verbose = true,
+        k_tol = 1.0e-8,
+        accel_pi_k_tol = 1.0e-8,
+        accel_pi_max_its = 50,
+    })
+    solver.Initialize(k_solver)
+    solver.Execute(k_solver)
```
ragusa, line 73: move outside of if statement?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+    ({
+        lbs_solver_handle = phys1,
+        diff_accel_sdm = "pwlc",
+        accel_pi_verbose = true,
+        k_tol = 1.0e-8,
+        accel_pi_k_tol = 1.0e-8,
+        accel_pi_max_its = 50,
+    })
+    solver.Initialize(k_solver)
+    solver.Execute(k_solver)
```
ragusa, line 85: move outside of if statement?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+        ({
+            lbs_solver_handle = phys1,
+            accel_pi_verbose = true,
+            k_tol = 1.0e-8,
+            accel_pi_k_tol = 1.0e-8,
+            accel_pi_max_its = 30,
+            diff_sdm = "pwld"
+        })
+        solver.Initialize(k_solver)
+        solver.Execute(k_solver)
```
ragusa, line 109: move outside of if statement?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
```diff
@@ -0,0 +1,131 @@
...
+    ({
+        lbs_solver_handle = phys1,
+        nl_max_its = 50,
+        nl_abs_tol = 1.0e-10,
+        nl_rel_tol = 1.0e-10,
+        l_max_its = 20,
+        num_initial_power_iterations = 2,
+    })
+    solver.Initialize(k_solver)
+    solver.Execute(k_solver)
```
ragusa, line 121: move outside of if statement?

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+        diffusion_solver_->Solve(phi0, /*use_initial_guess=*/true);
+
+        // Bump the new solution to old
+        phi0_old = phi0;
+      }
+
+      // Compute a new diffusion eigenvalue. This requires mapping the
+      // diffusion solution back to the transport discretization.
+      std::vector<double> phi;
+      TransferDiffusionToTransport(phi0, phi);
```
ragusa, line 254: is that operation needed even if the SMM was discretized using PWLD ?

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+            {
+              bndry_factors_[imap].resize(num_groupsets, 0.0);
+
+              double val = 0.0;
+              for (int n = 0; n < num_gs_angles; ++n)
+                val +=
+                  quadrature->weights_[n] * std::fabs(quadrature->omegas_[n].Dot(face.normal_));
+              bndry_factors_[imap][gs] = val / wt_sum;
+            }
+            //            const auto& bfac = bndry_factors_[imap][gs];
```
ragusa, line 408: is this supposed to be commetned out and left here?

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+              // Perform the quadrature integration
+              for (int n = 0; n < num_gs_angles; ++n)
+              {
+                const auto& wt = quadrature->weights_[n];
+                const auto& omega = quadrature->omegas_[n];
+                const auto mu = std::fabs(omega.Dot(face.normal_));
+
+                const auto psi_dof = pwld.MapDOFLocal(cell, i, psi_uk_man, n, gsg);
+
+                //                beta[g] += wt * (mu - bfac) * psi[gs][psi_dof];
```
ragusa, line 428: same as before I suspect. However, I thought that the ang quadrature was not exact for half-range integrals, so why are you using 0.5 here, and not the approximate values fromt he ang quadrature? Or am I missing the mark completely here?

ragusa (commented):
lots of comments in the code. thank you!

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+
+  // Create PWLC structure, if needed
+
+  if (diffusion_sdm_name_ == "pwlc")
+  {
+    pwlc_ptr_ = PieceWiseLinearContinuous::New(grid);
+    ghost_info_ = MakePWLDGhostInfo(pwld, phi_uk_man);
+  }
+
+  //======================================== Create tensor structure
```
andrsd, line 113: Remove them `=`s here and in the rest of the file.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+  diff_uk_man.AddUnknown(UnknownType::VECTOR_N, num_gs_groups);
+
+  // Create the diffusion boundaries
+  for (const auto& [bid, bc] : lbs_solver_.BoundaryPreferences())
+    if ((bc.type == BoundaryType::ISOTROPIC) or (bc.type == BoundaryType::ARBITRARY))
+      throw std::logic_error("Only vacuum and reflective boundaries are valid for "
+                             "k-eigenvalue problems.");
+
+  const auto bcs = TranslateBCs(lbs_solver_.SweepBoundaries(),
+                                /*vacuum_bcs_are_dirichlet=*/false);
```
andrsd, line 137: Remove `/*vacuum_bcs_are_dirichlet=*/`.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+  double k_eff_ell = k_eff_;
+  double k_eff_change = 1.0;
+
+  auto phi_ell = phi_old_local_;
+
+  int nit = 0;
+  while (nit < max_iters_)
+  {
+    // Compute the transport 1/k fission source
+    SetLBSFissionSource(phi_ell, /*additive=*/false);
```
andrsd, line 203: Remove `/*additive=*/`

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+        // Define the nodal diffusion scattering source.
+        // This returns a local PWLD vector which must still be integrated
+        // against the diffusion test functions.
+        auto Ss = SetNodalDiffusionScatterSource(phi0_old);
+
+        // Assemble the normal terms of the RHS of the diffusion solver,
+        // add in the second moment correction, and solve the system
+        auto b = AssembleDiffusionRHS(Sf + Ss);
+        diffusion_solver_->AddToRHS(b + correction);
+        diffusion_solver_->Solve(phi0, /*use_initial_guess=*/true);
```
andrsd, line 245: Remove `/*use_initial_guess=*/`

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+
+      // Bump variables for next iteration
+      lambda_old = lambda;
+      F0_old = F0;
+      phi0_m = phi0;
+
+      if (lambda_change < accel_pi_k_tol_)
+        break;
+
+    } // acceleration
```
andrsd, line 272: If this is to tell the reader about the acceleration loop, then this should be before the `for` statement.
If this is just copy-paste, maybe even delete this. 

Also, these comments _were_ helpful when for loops were long and people used editors that were not able to go to the matching opening curly brackets. These days we don't really need these, IMO. Maybe, it is time to have discussion about these comments. I can see them a lot in the code. What does @ragusa and @wdhawkins think about those? Useful? Not useful?

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+
+    if (k_eff_change < k_tolerance_)
+      break;
+  }
+
+  // Print summary
+  log.Log() << "\n";
+  log.Log() << "        Final k-eigenvalue    :        " << std::setprecision(7) << k_eff_;
+  log.Log() << "        Final change          :        " << std::setprecision(6) << k_eff_change
+            << " (num_TrOps:" << front_wgs_context_->counter_applications_of_inv_op_ << ")"
```
andrsd, line 308: Should `num_TrOps` be something a user would understand better? It seems like it was a variable name, but it is not any more?

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+  // Compute integrals
+  const auto num_local_cells = lbs_solver_.Grid().local_cells.size();
+  K_tensor_matrices_.resize(num_local_cells);
+  for (const auto& cell : lbs_solver_.Grid().local_cells)
+  {
+    const auto& cell_mapping = discretization.GetCellMapping(cell);
+    const auto num_cell_nodes = cell_mapping.NumNodes();
+    const auto qp_data = cell_mapping.MakeVolumetricFiniteElementData();
+
+    //    TensorMatrix K(num_cell_nodes, std::vector<MatDbl>(num_cell_nodes));
```
andrsd, line 813: This line can be most likely removed

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+        output[output_dof] = input[input_dof];
+      }
+  }
+}
+
+double
+PowerIterationKEigenSMM::CheckScalarFluxConvergence(const std::vector<double>& phi_new,
+                                                    const std::vector<double>& phi_old)
+{
+
```
andrsd, line 934: Empty line can be removed

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1122 @@
...
+  for (const int64_t ghost_id : ghost_ids)
+    ghost_global_to_local_map[ghost_id] = static_cast<int64_t>(num_local_dofs + k++);
+
+  return {vec_ghost_comm, ghost_global_to_local_map};
+}
+
+int
+PowerIterationKEigenSMM::MapAssociatedFaceNode(const Vector3& node,
+                                               const std::vector<Vector3>& nbr_nodes,
+                                               const double epsilon /* = 1.0e-12 */)
```
andrsd, line 1112: Remove `/* = 1.0e-12 */`, it is in the header file. 

The reason why one does not want to do this is that sometime in future, somebody is gonna change the value in the header file and they won't change it here and it would create inconsistency.

andrsd (changes requested):
Also, commit no. 3 should be squashed into commit no. 1

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
ragusa: I actually thought it was a clever way of commenting something related to the specifics of the method. I would very much prefer to keep such comments.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
ragusa: I actually thought it was a clever way of commenting something related to the specifics of the method. I would very much prefer to keep such comments.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
ragusa: I actually thought it was a clever way of commenting something related to the specifics of the method. I would very much prefer to keep such comments.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
ragusa: I agree it is not very clear for an average user. I think it stands for number of transport operations (or action of the transport operator).  Myself, I am not 100% sure if a unit of such `TrOps` is a sweep per group and per direction or if it is some large unit of work.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
ragusa: this is an example of a comment that, I agree, can be removed

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: The logic of both PWLC->PWLD and PWLD->PWLD are within this routine. A separate vector is used for the diffusion step, so it needs to be transferred back into the flux moments vector even for the same discretization.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: For right now, yes. There is a bit of a story behind this that I will explain in the next comment.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: You are not missing the mark at all. You just caught a very small cheat that I have that I intended to fix but never got around to. Because we are recycling the PWLC diffusion code, the boundary conditions in the matrix are set with 0.5. For the sake of consistency, I use 0.5 here as well. The lines that are commented out for this reason, but should remain for whenever the diffusion boundary condition is changed in the matrix.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: It is the number of sweeps (transport operations) done for a given groupset. This is recycled from other existing outputs in the code. I can change this to something else, just let me know what you suggest.

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7.lua
zhardy-lanl: I can do that. I did this because it is the way the file existed before I touched it. I will do this for each c5g7 test script.

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_jfnk.lua
zhardy-lanl: Typo. Good catch.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: I have no preference. My last commit removes these. I can revert later if necessary.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: I am in favor of keeping them. They don't hurt anything IMO and help when the modern IDEs are not available.

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7_pi_smm.lua
zhardy-lanl: Shouldn't be. Fixing now.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
andrsd: So `Number of sweeps:` ?

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
ragusa: I prefer to get these.

zhardy-lanl: @andrsd How firm are you on removing these?

andrsd: They are one-off thing. How do you determine when to use them and when not? You certainly don't call other methods using this style. It reminds me of keyword params from Python, i.e. this does not belong into C++, IMO. And there should be doco comments at the method declarations to explain these things, so user code should not need to do this at all.
In my book, these are good reasons against doing this.

zhardy-lanl: I think those are fair points. I will leave them out and not revert the commit removing them.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: Changing to `Sweep Count:`

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: @ragusa The latest commit implements the correct boundary conditions. I can discuss it more in depth tomorrow at our meeting as that it is a little convoluted. Nonetheless, I'm sure you'd be interested to know that the final eigenvalue did not change through at least the 6th digit.

ragusa (approved):

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1212 @@
...
+        break;
+
+    } // acceleration
+
+    // Project back to transport and check convergence
+    k_eff_ = lambda;
+    k_eff_change = std::fabs(k_eff_ / k_eff_ell - 1.0);
+    k_eff_ell = k_eff_;
+
+    TransferDiffusionToTransport(phi0, phi_new_local_);
```
wdhawkins, line 277: Any reason this method doesn't return std::vector? I think it would result in cleaner code.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1212 @@
...
+          // discretization is used to obtain the tensor DoF.
+          const auto jmap = pwld.MapDOFLocal(cell, j, tensor_uk_man_, g, 0);
+          const double* T = &tensors[jmap];
+
+          // This adds the rank-2 tensor contraction (double dot product)
+          // of the closure tensor and the tensor stiffness matrix for nodes
+          // i and j.
+          for (int k = 0; k < dimension_; ++k)
+            for (int l = 0; l < dimension_; ++l)
+              val += *T++ * K[i][j][k][l];
```
wdhawkins, line 489: We shouldn't do this: `*T++`. I'll just leave the one comment, but it applies to several places in this file.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1212 @@
...
+          {
+            const auto& omega = quad->omegas_[d];
+            const auto& wt = quad->weights_[d];
+
+            const auto psi_dof = pwld.MapDOFLocal(cell, i, psi_uk_man, d, gsg);
+            const auto coeff = wt * psi[gs][psi_dof];
+
+            for (int k = 0; k < dimension_; ++k)
+            {
+              for (int l = 0; l < dimension_; ++l)
```
wdhawkins, line 368: Although it's not in our style guide, I think it's good practice to always use braces for loops and conditionals unless the bodies of all nested statements are a single line.  For example:
```
for (int i = 0; i < a; ++i)
  for(int j = 0; j < b; ++j)
   c = matrix[j][j];
```
requires no braces, but:
```
for (int i = 0; i < a; ++i)
  for(int j = 0; j < b; ++j)
     if [i != 0] 
       c = matrix[j][j];
```
would need to be written as:
```
for (int i = 0; i < a; ++i)
{
  for(int j = 0; j < b; ++j)
  {
     if [i != 0] 
       c = matrix[j][j];
  }
}
```

This applies in several places, but it's strictly a style decision.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
```diff
@@ -0,0 +1,1212 @@
...
+    const auto& cell_mapping = pwld.GetCellMapping(cell);
+
+    for (int i = 0; i < cell_mapping.NumNodes(); ++i)
+    {
+      // Get the diffusion and transport DoFs
+      const auto input_dof_map = pwld.MapDOFLocal(cell, i, phi_uk_man, 0, first_grp);
+      const auto output_dof_map = diff_sd.MapDOFLocal(cell, i, diff_uk_man, 0, 0);
+
+      // Now copy the data
+      double* output_ptr = &output[output_dof_map];
```
wdhawkins, line 971: I think:
```
std::copy_n(std::advance(mapped_phi.begin() + input_dof_map), 
            num_gs_groups,
            std::advance(output.begin() + output_dof_map));
```
will work here without having to devolve everything to pointers.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
wdhawkins: I agree with @andrsd. Unless it's a very long loop that spans several hundred lines, I don't find the end-of-loop comments very useful and usually remove them during a refactor.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: Yes there is a reason for it. We are copying a scalar flux into a flux moments vector which may have higher moments that are needed for the subsequent sweep. In reality, `TransferTransportToDiffusion` could have an `std::vector<double>` return, but I wanted to keep the call signatures the same since they are complimentary functions to one another.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: Just out of curiosity, why should we not do this? Readability? I have always liked doing this for contiguous data since it I don't like having to index like `T[k * dimension_ + l]`, or `psi[cell_id * num_cell_nodes * num_dirs * num_grps + i * num_dirs * num_grps + d * num_grps + g]` when taken to more of an extreme.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: I'm a little confused by your example because you say "unless the bodies of all nested statements are a single line".  Isn't the body of all loops and conditionals in the example a single line? The code you tagged here fits your description of not having a single line body. I will make the change for similar loops throughout, but will leave anything with a single line body as is.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
wdhawkins: I think we should avoid pointer arithmetic wherever possible. It may be more concise, but it's also error prone. Even though it looks messier, I think it's easier to see what's happening when you use index notation - especially for new developers just learning the code and for developers having to maintain the code.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: How about something like: 
```
std::copy_n(&mapped_phi[input_dof_map], num_gs_groups, &output[output_dof_map]);
```
I think `std::advance` does the same thing as `+` in your suggestion, so `+` should be a `,` instead, or `std::advance` not used.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
wdhawkins: Single statement, yes, but the `if` statement is multi-line.  Hopefully I understood your comment and that makes sense.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: Yea, that's fair. It is self documenting, but very gross, haha! Making the change. We should probably discuss this on Wednesday and create an issue for removing pointer arithmetic if we decide that's the route we want to go. I've always thought that `*T++` was faster than `T[k * dimension_ + l]`. Is that not the case? Or is it inconsequential? I'll make the change.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
wdhawkins: You are right, the plus should be a comma in my code above. I think I originally was going for `mapped_phi.begin() + input_dof_map` which would also work. The `std::advance` just made it more general.  These are all vectors and probably won't change, so your `copy_n` solution is just fine.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
zhardy-lanl: This is a topic that we should likely discuss on Wednesday. I prefer to have the end of loop comments, *especially* for deeply nested loops. I am not a fan of things like:
```
            }
          }
        }
      }
    }
  }
}
```
Having comments helps me keep my place when developing, especially when a counter needs to be incremented or something like that. Otherwise, I find myself scrolling up and down to make sure I am in the correct level of nesting for the line of code.

wdhawkins: Valid point... in cases like that, I can see where end-of-loop comment could be very useful.

modules/linear_boltzmann_solvers/executors/pi_keigen_smm.cc
wdhawkins: I strongly suspect that there is no difference in performance between the two. The index notation is going to end up being pointer arithmetic after the compiler is finished with it.

wdhawkins (approved):

PR #290, opened 2024-06-08
Fixing balance calculation with reflecting boundary conditions

wdhawkins:
This PR is dependent on #306 for new regression tests to successfully pass.


test/modules/linear_boltzmann_solvers/transport_steady_cbc/hdpe_balance.lua
```diff
@@ -0,0 +1,79 @@
...
+num_groups=1
+
+-- Add cross sections to materials
+if num_groups == 172 then
+  mat.SetProperty(materials[1], TRANSPORT_XSECTIONS, OPENMC_XSLIB, "../../MGXS_Generation/detector/mgxs_det/mgxs_detector.h5", 294.0, "hdpe")
+else
+  total = {1.}
+  c     = {0.9}
+  mat.SetProperty(materials[1], TRANSPORT_XSECTIONS, SIMPLE_ONE_GROUP, total[1], c[1])
+end
```
ragusa, line 31: currently, this input file is only using the 1-g manufactured data. 
- If you prefer that the test be run with the real HDPE XS, let me know. I can email them to you. 
- If you prefer to keep the 1-g test case, then you should delete the unused lines and remove the if statement and also, maybe, change the name of the lua file so that it does not refer to HDPE.

test/modules/linear_boltzmann_solvers/transport_steady/hdpe_balance.lua
```diff
@@ -0,0 +1,78 @@
...
+num_groups=1
+
+-- Add cross sections to materials
+if num_groups == 172 then
+  mat.SetProperty(materials[1], TRANSPORT_XSECTIONS, OPENMC_XSLIB, "../../MGXS_Generation/detector/mgxs_det/mgxs_detector.h5", 294.0, "hdpe")
+else
+  total = {1.}
+  c     = {0.9}
+  mat.SetProperty(materials[1], TRANSPORT_XSECTIONS, SIMPLE_ONE_GROUP, total[1], c[1])
+end
```
ragusa, line 31: same comment

ragusa:
@wdhawkins  also, let us **NOT** close the balance issue yet. I still need to check what is happening i the balance statement for (1) k-eigenvalue problems and for (2) subcritical source-driven problems!

ragusa:
@wdhawkins : test failed. do you have a tab in this line in the json file?
"[0]   Balance (Gain - Loss)       ="

andrsd (approved):

test/modules/linear_boltzmann_solvers/transport_steady_cbc/hdpe_balance.lua
wdhawkins: Yes, I was going to ask you to send me your HDPE XS file. If you'll email it to me, I'll fix these tests.

zhardy-lanl (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_solver/lbs_discrete_ordinates_solver.cc
```diff
@@ -654,55 +653,66 @@ DiscreteOrdinatesSolver::ComputeBalance()
...
     const auto& IntS_shapeI = fe_intgrl_values.intS_shapeI;
 
-    // Inflow
-    // This is essentially an integration over
-    // all faces, all angles, and all groups.
-    // Only the cosines that are negative are
-    // added to the integral.
+    // Inflow: This is essentially an integration over all faces, all angles, and all groups. For
+    // non-reflective boundaries, only the cosines that are negative are added to the inflow
+    // integral. For reflective boundaries, inflow = outflow.
```
ragusa, line 658: This is ONLY true in the math world. In practice, there are iterations, etc., so that equality implies that there are the same but it is NOT true in a code. Please amend the comment. Maybe state, For reflective boundaries, it is expected that,upon convergence, inflow = outflow (within numerical tolerances set by the user).

ragusa (commented):
I bet we do not check convergence on reflected angular flux, correct? It would be good to see what a 1-g pure absorber problem does for a fully reflected cube and a volumetric source. I bet we will convergence in 1 iteration (no scattering) but the net leakage will be non-zero

wdhawkins:
> I bet we do not check convergence on reflected angular flux, correct? It would be good to see what a 1-g pure absorber problem does for a fully reflected cube and a volumetric source. I bet we will convergence in 1 iteration (no scattering) but the net leakage will be non-zero

I do think we converge the reflected flux. This was always an issue in PDT, but Jan and I had a discussion about fixing this. I'll run a 1g test problem and see.

ragusa:
@wdhawkins  can you do a 1-g pure absorber in a cube with a volumetric source and all reflective BC. This should converge in one iteration because there is no scattering. But the inflow will not be equal to the outflow if we do not have a loop to converge the reflective BC angular fluxes. Unless there is some trickery with the TDG and cycle detection.

ragusa:
> > I bet we do not check convergence on reflected angular flux, correct? It would be good to see what a 1-g pure absorber problem does for a fully reflected cube and a volumetric source. I bet we will convergence in 1 iteration (no scattering) but the net leakage will be non-zero
> 
> I do think we converge the reflected flux. This was always an issue in PDT, but Jan and I had a discussion about fixing this. I'll run a 1g test problem and see.
> 
> 1g, infinite, volumetric source, pure absorber:
> 
> ```
> [0]  00:00:00.1 Computing b
> [0]  00:00:00.1 WGS groups [0-0] Iteration     0 Residual         1
> [0]  00:00:00.1 WGS groups [0-0] Iteration     1 Residual  0.141115
> [0]  00:00:00.1 WGS groups [0-0] Iteration     2 Residual 0.0142328
> [0]  00:00:00.1 WGS groups [0-0] Iteration     3 Residual 0.000883628
> [0]  00:00:00.1 WGS groups [0-0] Iteration     4 Residual 2.82161e-08
> [0]  00:00:00.1 WGS groups [0-0] Iteration     5 Residual 2.42963e-12 CONVERGED
> ...
> [0]  Balance table:
> [0]   Absorption rate             = 1.000000e+03
> [0]   Production rate             = 1.000000e+03
> [0]   In-flow rate                = 1.553633e+02
> [0]   Out-flow rate               = 1.553633e+02
> [0]   Gain (In-flow + Production) = 1.155363e+03
> [0]   Balance (Gain - Loss)       = -1.881517e-11
> [0]   Balance/Gain, in %          = -1.628507e-12
> ```

Great. Does this work equally well for GMES and Classic Richardson ??

wdhawkins:
It does work for both GMRES and Richardson.

ragusa:
> It does work for both GMRES and Richardson.

Can we add this 1-g test to the regression suite and grab the `In-flow rate` from the console as the test value? Thanks!! I'll approve after that!

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
```diff
@@ -515,5 +515,43 @@
         "abs_tol": 1.0e-6
       }
     ]
+  },
+  {
+    "file": "1g_infinite_pure_absorber_balance_gmres.lua",
+    "comment": "Infinite, 1g, pure absoreber, with balance using GMRES",
```
ragusa, line 521: typo -- > absorber

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
```diff
@@ -515,5 +515,43 @@
...
+        "type": "KeyValuePair",
+        "key": "[0]   Out-flow rate               =",
+        "goldvalue": 1.553633e+02,
+        "abs_tol": 1.0e-6
+      }
+    ]
+  },
+  {
+    "file": "1g_infinite_pure_absorber_balance_richardson.lua",
+    "comment": "Infinite, 1g, pure absoreber, with balance using Richardson",
```
ragusa, line 540: typo -- > absorber

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
```diff
@@ -515,5 +515,43 @@
...
+      {
+        "type": "KeyValuePair",
+        "key": "[0]   In-flow rate                =",
+        "goldvalue": 1.965343e+02,
+        "abs_tol": 1.0e-6
+      },
+      {
+        "type": "KeyValuePair",
+        "key": "[0]   Out-flow rate               =",
+        "goldvalue": 1.965343e+02,
```
ragusa, line 552: I do not get it: why is the golden value for richardson 1.96e2 while it is 1.55e2 for GMRes. Are you not solving the same problem??

ragusa (changes requested):

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_gmres.lua
```diff
@@ -0,0 +1,71 @@
...
+end
+
+meshgen = mesh.OrthogonalMeshGenerator.Create({ node_sets = { nodes, nodes, nodes } })
+mesh.MeshGenerator.Execute(meshgen)
+
+-- Set Material IDs
+mesh.SetUniformMaterialID(0)
+
+materials = {}
+materials[1] = mat.AddMaterial("HDPE")
```
ragusa, line 20: change name or simply remove name

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_richardson.lua
```diff
@@ -0,0 +1,70 @@
...
+end
+
+meshgen = mesh.OrthogonalMeshGenerator.Create({ node_sets = { nodes, nodes, nodes } })
+mesh.MeshGenerator.Execute(meshgen)
+
+-- Set Material IDs
+mesh.SetUniformMaterialID(0)
+
+materials = {}
+materials[1] = mat.AddMaterial("HDPE")
```
ragusa, line 20: remove name

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
wdhawkins: It is the same problem. I chalked this up to differences in the iterative process. No?

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_gmres.lua
```diff
@@ -0,0 +1,71 @@
...
+      inner_linear_method = "gmres",
+      l_abs_tol = 1.0e-9,
+      l_max_its = 300,
+      gmres_restart_interval = 30,
+    },
+  },
+  options = {
+    scattering_order = 0,
+    spatial_discretization = "pwld",
+    save_angular_flux = true,
```
ragusa, line 50: I do not think this is needed so please remove

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_richardson.lua
```diff
@@ -0,0 +1,70 @@
...
+      angular_quadrature_handle = pquad,
+      inner_linear_method = "richardson",
+      l_abs_tol = 1.0e-9,
+      l_max_its = 300,
+    },
+  },
+  options = {
+    scattering_order = 0,
+    spatial_discretization = "pwld",
+    save_angular_flux = true,
```
ragusa, line 49: I do not think this is needed so please remove

test/modules/linear_boltzmann_solvers/transport_steady/hdpe_balance.lua
```diff
@@ -0,0 +1,69 @@
...
+      inner_linear_method = "gmres",
+      l_abs_tol = 1.0e-9,
+      l_max_its = 300,
+      gmres_restart_interval = 30
+    }
+  },
+  options = {
+    scattering_order = 0,
+    spatial_discretization = "pwld",
+    save_angular_flux = true,
```
ragusa, line 53: I do not think this is needed so please remove

ragusa (changes requested):

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
ragusa: @wdhawkins Not possible! The solution is uniform and isotropic, Phi = Q/sigma_a, and Psi = Phi/4pi. So, the leakage are the half-range integrals of a constant, isotropic value on each face, so both methods need to give the same leakage! 196 versus 155 is a HUGE difference.

test/modules/linear_boltzmann_solvers/transport_steady/hdpe_balance.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_richardson.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_gmres.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_richardson.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/1g_infinite_pure_absorber_balance_gmres.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
wdhawkins: And, I've noticed that richardson gives a different answer depending on the number of processors.

ragusa: @wdarylhawkins : the exact answer is, for your problem: 150 / true_pi * half_range_integral
where the half range integral is Omega dot n integrated over the half range. This value gets closer and closer to true_pi as you add more directions. So, I would trust more the value from GMRES, which is 155,... with 32 directions.

Two things:
1. you need to fix Richardson
2. please verify that, as the # of directions increases, the lekage tends towards 150.00000

wdhawkins: #2 certainly looks to be true with GMRES. At 32k angles, leakage is 1.500403e+02. At 64k angles, it's 1.500102e+02.

wdhawkins: Not sure what's up with Richardson. Even without reflecting bcs, something seems off. I'll start looking at it.

wdhawkins: Small update. This bug is not something that was introduced in OpenSn. It looks like it's a chi-tech bug.

wdhawkins: I believe I have it fixed. Let me do a bit of testing, but we should discuss.

wdhawkins: #306 fixes the issue with Richardson. I've also verified that leakages tends towards 150 as we refine the number of angles for the Richardson test.

ragusa (approved):

PR #291, opened 2024-06-10
DFEMSolver can be constructed using InputParameters

andrsd:
Refs #98


lua/modules/cfem_diffusion/cfem_diff_solver.cc
zhardy-lanl, line 1: While we are modifying these files, it may be good to rename files to `c(d)fem_diffusion_solver.(h)cc`

lua/modules/cfem_diffusion/cfem_diff_solver.h
zhardy-lanl, line 1: Could this header file go away?

lua/modules/dfem_diffusion/dfem_diff_solver.h
zhardy-lanl, line 1: Could this header file go away?

modules/common/diffusion_bndry.h
zhardy-lanl, line 1: Not sure I am a huge fan of having a `common` directory. I think it would make more sense to have a `diffusion` directory that houses `cfem` and `dfem` subdirectories (for now)

modules/dfem_diffusion/dfem_diffusion_solver.h
```diff
@@ -81,16 +70,38 @@ class Solver : public opensn::Solver
...
+  size_t num_globl_dofs_ = 0;
+
+  std::vector<double> field_;
+
+  /// approx solution
+  Vec x_ = nullptr;
+  /// RHS
+  Vec b_ = nullptr;
+  /// linear system matrix
+  Mat A_ = nullptr;
```
zhardy-lanl, line 90: For short member declarations, does it make more sense to put descriptions to the right?

zhardy-lanl (changes requested):

modules/common/diffusion_bndry.h
andrsd: I was thinking the same for all the header files, etc. that you commented on above. I think I would build the `diffusion` module after I convert the FV and MG diffusion solvers. Right now, I am just converting the constructors to use the input parameters.

modules/dfem_diffusion/dfem_diffusion_solver.h
andrsd: I usually do not do that. One has to remember to use `///<` otherwise the doco ends up with the wrong member variable. And if the text gets longer (for different reasons), clang starts to format into a column over multiple lines while when I use the comment before, I get the full width.  So, I personally do not mix the `///` and `///<` comments.

ragusa (approved):

modules/common/diffusion_bndry.h
zhardy-lanl: Okay, so `common` is temporary then? I don't particularly like adding a directory into the source tree for a temporary purpose, but if this will go away... I will approve.

zhardy-lanl (approved):

wdhawkins (approved):

PR #292, opened 2024-06-10
No more local Vec3

andrsd:
Refs #86


ragusa (approved):

zhardy-lanl (approved):
Agreed that a typedef to save 3 characters is a bit excessive... Thanks

PR #296, opened 2024-06-17
MeshAttributes are no longer bit-wise flags

### LINKED ISSUE #286: Remove bitwise ops for MeshAttribute enum
The MeshAttribute enum uses bitwise operations to provide support for combinations of mesh types. Unfortunately, this is overly complicated and buggy. We need to convert this to a standard enum and explicitly handle the (limited) mesh combinations.

andrsd:
- MeshAttributes are not bit-wise flags
- Removing trailing underscores from public member variables in SplitMeshInfo
- Creating setters according to code standards in `UnpartitionedMesh`

Closes #286

### DIFF
--- a/framework/mesh/mesh_generator/split_file_mesh_generator.cc
+++ b/framework/mesh/mesh_generator/split_file_mesh_generator.cc
@@ -300,8 +300,8 @@ SplitFileMeshGenerator::ReadSplitMesh()
     dir_path.string() + "/" + file_prefix_ + "_" + std::to_string(pid) + ".cmesh";
 
   SplitMeshInfo info_block;
-  auto& cells = info_block.cells_;
-  auto& vertices = info_block.vertices_;
+  auto& cells = info_block.cells;
+  auto& vertices = info_block.vertices;
   std::ifstream ifile(file_path, std::ios_base::binary | std::ios_base::in);
 
   OpenSnLogicalErrorIf(not ifile.is_open(), "Failed to open " + file_path.string());
@@ -321,7 +321,7 @@ SplitFileMeshGenerator::ReadSplitMesh()
   info_block.ortho_attributes.Ny = ReadBinaryValue<size_t>(ifile);
   info_block.ortho_attributes.Nz = ReadBinaryValue<size_t>(ifile);
 
-  info_block.num_global_vertices_ = ReadBinaryValue<size_t>(ifile);
+  info_block.num_global_vertices = ReadBinaryValue<size_t>(ifile);
 
   // Read boundary map
   const size_t num_bndries = ReadBinaryValue<size_t>(ifile);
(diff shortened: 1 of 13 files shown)


zhardy-lanl (commented):
Some more work needs to be done here to finalize this. First, `MeshAttributes` should probably be renamed to something like `MeshType`. Attributes was used as a catchall term. Second, a mesh can be unstructured or orthogonal **and** extruded, so this structure doesn't quite work I don't think.

Here is my proposal:
```
enum MeshType : int
{
  UNSTRUCTURED,
  ORTHOGONAL
}
```
Meshes should be defaulted to `UNSTRUCTURED` since that is the most general.

Additionally, a bool attribute should be added to `MeshContinuum` called `extruded_`, defaulted to false, with a getters and setters for now.

wdhawkins (approved):

andrsd:
I can do this work in this PR. It is just that #286 did not specify this work - just undo the bit-wise flags. But, I do agree with your comment @zhardy-lanl .

ragusa:
> I can do this work in this PR. It is just that #286 did not specify this work - just undo the bit-wise flags. But, I do agree with your comment @zhardy-lanl .

Sorry the specs were not clear in the issue. I agree with Zach and thanks for updating this PR.

framework/mesh/mesh_generator/split_file_mesh_generator.cc
```diff
@@ -316,12 +317,13 @@ SplitFileMeshGenerator::ReadSplitMesh()
                          std::to_string(opensn::mpi_comm.size()) + " processes.");
 
   info_block.dimension = ReadBinaryValue<unsigned int>(ifile);
-  info_block.mesh_attributes_ = ReadBinaryValue<int>(ifile);
+  info_block.mesh_type = static_cast<MeshType>(ReadBinaryValue<int>(ifile));
```
ragusa, line 320: just curious what the static cast does here. it wasn't there before. what is an mistake?

ragusa (approved):

framework/mesh/mesh_generator/split_file_mesh_generator.cc
andrsd: Originally `mesh_attributes_` were declared as `int` the `SplitMeshInfo` and the static cast was done `SplitFileMeshGenerator::SetupLocalMesh `. I changed the declaration of `SplitMeshInfo::mesh_attributes_` to be `MeshAttributes`, so the static cast was moved inside the `SplitFileMeshGenerator::ReadSplitMesh `.

Then, `mesh_attributes_` became `mesh_type`...

framework/mesh/unpartitioned_mesh/unpartitioned_mesh.h
```diff
@@ -64,11 +64,14 @@ class UnpartitionedMesh
...
-  Options& MeshOptions() { return mesh_options_; }
+  void SetMeshOptions(const Options& opts) { mesh_options_ = opts; }
   const Options& MeshOptions() const { return mesh_options_; }
 
-  MeshAttributes& Attributes() { return attributes_; }
-  const MeshAttributes& Attributes() const { return attributes_; }
+  void SetType(MeshType type) { mesh_type_ = type; }
+  const MeshType& Type() const { return mesh_type_; }
+
+  void SetExtruded(bool extruded = true) { extruded_ = extruded; }
```
zhardy-lanl, line 73: While it makes sense to have this default to `true`, this routine appearing without an argument looks a little funny. Would it make more sense to just not have a default argument?

zhardy-lanl (approved):

PR #307, opened 2024-06-21
Diffusion module refactoring

andrsd:
- FVSolver can be constructed using InputParameters
- MGSolver can be constructed using InputParameters
- Grouping diffusion solvers into a diffusion module

Refs #98


andrsd:
There are no regression tests for the FV and MG solver. They will probably uncover some mistakes. But, these tests could be done as a separate task...

lua/modules/diffusion/diffusion.cc
```diff
@@ -2,19 +2,162 @@
...
+  auto& solver =
+    opensn::GetStackItem<opensn::diffusion::CFEMSolver>(opensn::object_stack, handle, fname);
+  solver.SetDCoefFunction(d_coef_function);
+  solver.SetQExtFunction(q_ext_function);
+  solver.SetSigmaAFunction(sigma_a_function);
+
+  auto options_params = opensn::diffusion::CFEMSolver::OptionsBlock();
+  options_params.AssignParameters(params.GetParam("arg1"));
+
+  solver.SetOptions(options_params);
```
zhardy-lanl, line 61: Is there a base diffusion solver that could house this since it seems to be the same for all diffusion solvers?

lua/modules/diffusion/diffusion.cc
```diff
@@ -2,19 +2,162 @@
...
+  auto& solver =
+    opensn::GetStackItem<opensn::diffusion::MGSolver>(opensn::object_stack, handle, fname);
+
+  auto options_params = opensn::diffusion::MGSolver::OptionsBlock();
+  options_params.AssignParameters(params.GetParam("arg1"));
+
+  solver.SetOptions(options_params);
+
+  return opensn::ParameterBlock();
+}
```
zhardy-lanl, line 154: Is CFEM the only option in the MG solver at present? If so, that should be changed fixed later on.

modules/diffusion/boundary.h
```diff
@@ -29,6 +30,15 @@ class Boundary
...
 };
 
+/** Multigroup diffusion boundary */
+class MGBoundary
+{
+public:
+  BoundaryType type_ = BoundaryType::Vacuum;
+
+  std::array<std::vector<double>, 3> mg_values_;
+};
```
zhardy-lanl, line 40: Ultimately, it would be nice to have a group agnostic diffusion solver. We should consider redoing the diffusion solvers in the future.

lua/modules/diffusion/diffusion.cc
ragusa: is the diffusion solvers are merged here, then CFEM and DFEM should be available options

andrsd: The MG solver only has CFEM option as far as I can tell..

lua/modules/diffusion/diffusion.cc
andrsd: There is not yet. I think I would like to see the tests for MG and FV solvers first. 

It is obvious that some base class for the diffusion solver is needed. There is a lot of overlap ATM. But, one step at a time :-)

lua/modules/diffusion/diffusion.cc
ragusa: ok, so the description "Grouping diffusion solvers into a diffusion module" confused me. I see now that it is more a grouping of the files than a synthesis/merge of all of the flavors of diffusion solvers.

modules/diffusion/boundary.h
andrsd: I think this data structure is only used to store the user input from the lua file. Then, it is internally converted into what ever the solver needs. Am I right @ragusa or did I miss something?

lua/modules/diffusion/diffusion.cc
andrsd: That would be really helpful 👍 Thanks!

modules/diffusion/boundary.h
ragusa: this is the (a,b,c) coefficients of the Robin BC in multigroup

lua/modules/diffusion/diffusion.cc
zhardy-lanl: Yep, not asking for anything now! At some point, there was a single diffusion solver that had discretization options (i.e. if you used CFEM it called a different assembly routine than DFEM). Something like that could be worth thinking about in the future

modules/diffusion/boundary.h
zhardy-lanl: Right, this is the general form of a Robin BC for multigroup. It could be used for one-group instead of having a separate boundary condition. This is not something to change now. Just jotting down thoughts so I am less likely to lose them in the future.

lua/modules/diffusion/diffusion.cc
andrsd: In this sense, module is a collection of classes. Now, all solvers are in the diffusion namespace, both in the C++ and lua space.  The work that would introduce the common diffusion class, etc. would be called refactoring.

lua/modules/diffusion/diffusion.cc
ragusa: I suppose this is not what you are looking for: mip_diffusion_3d_3a_dsa_ortho.lua

andrsd: > I suppose this is not what you are looking for: mip_diffusion_3d_3a_dsa_ortho.lua

No. That's using the DFEM solver which was already converted. As far as I could tell, there were no tests for the MG or FV solver.

zhardy-lanl (approved):

ragusa (approved):

wdhawkins (approved):

PR #319, opened 2024-06-25
adding chi to xs

ragusa:
adding chi so that it can be printed out in the console (why was it not there before?).

added the first test under `framework/materials` (I think we could use more in that folder)

@zhardy-lanl  : check the line where chi is multiplied by `ff_i`

### DIFF
--- a/framework/materials/multi_group_xs/multi_group_xs.h
+++ b/framework/materials/multi_group_xs/multi_group_xs.h
@@ -107,6 +107,8 @@ public:
     return adjoint_ ? transposed_transfer_matrices_.at(ell) : transfer_matrices_.at(ell);
   }
 
+  const std::vector<double>& Chi() const { return chi_; }
+
   const std::vector<double>& SigmaFission() const { return sigma_f_; }
 
   const std::vector<double>& NuSigmaF() const { return nu_sigma_f_; }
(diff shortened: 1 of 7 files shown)


framework/materials/multi_group_xs/openmc_import.cc
```diff
@@ -137,17 +137,17 @@ MultiGroupXS::Initialize(const std::string& file_name,
     }
 
     // Chi
-    auto chi = H5ReadDataset1D<double>(file, path + "chi");
-    OpenSnLogicalErrorIf(chi.empty(), "\"chi\" data block not found in " + file_name + ".");
-    if (not chi.empty())
+    chi_ = H5ReadDataset1D<double>(file, path + "chi");
+    OpenSnLogicalErrorIf(chi_.empty(), "\"chi\" data block not found in " + file_name + ".");
+    if (not chi_.empty())
```
wdhawkins, line 142: No need for `if (not chi_.empty())` since the code won't get past the `OpenSnLogicalErrorIf` if chi is empty.

andrsd (changes requested):

ragusa:
@wdhawkins @andrsd thanks for the early comments. I should have put this one as a draft as I still need to add a test. I'll invite people to review later this AM

framework/materials/multi_group_xs/openmc_import.cc
ragusa: fix. I suppose this was there before :-)

framework/materials/multi_group_xs/multi_group_xs.h
```diff
@@ -147,6 +149,7 @@ class MultiGroupXS : public MaterialProperty
   std::vector<double> sigma_a_;            ///< Absorption cross section
   std::vector<double> sigma_f_;            ///< Fission cross section
   std::vector<double> nu_sigma_f_;         ///< Neutron production due to fission
+  std::vector<double> chi_;                ///< Neutron production due to fission
```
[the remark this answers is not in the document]
ragusa: marinara sauce added.

andrsd (approved):

zhardy-lanl (approved):

wdhawkins (approved):

PR #320, opened 2024-06-26
Changing global variable `name`

zhardy-lanl:
To avoid clashes, renaming the global variable `name` to a static variable named `program`. Please suggest alternatives if you have other ideas.

### DIFF
--- a/framework/runtime.cc
+++ b/framework/runtime.cc
@@ -52,7 +52,7 @@ Initialize()
     cali_mgr.start();
   }
 
-  CALI_MARK_BEGIN(opensn::name.c_str());
+  CALI_MARK_BEGIN(opensn::program.c_str());
 
   SystemWideEventPublisher::GetInstance().PublishEvent(Event("ProgramStart"));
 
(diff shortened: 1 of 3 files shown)


ragusa (approved):
I love `program`. It reminds me of the good old Fortran days !

andrsd:
I am not sure what name clashes this avoid. An identifier must be unique within a namespace, this is why we created the `opensn` namespace and put the whole code inside it. If you have another variable called `name`, then you avoid the clash by using the namespace :-) That is the purpose behind the namespace. It allows you do specify the variable you want without the need of renaming.

zhardy-lanl:
> I am not sure what name clashes this avoid. An identifier must be unique within a namespace, this is why we created the `opensn` namespace and put the whole code inside it. If you have another variable called `name`, then you avoid the clash by using the namespace :-) That is the purpose behind the namespace. It allows you do specify the variable you want without the need of renaming.

Not clashes exactly, but when `name` is used as an argument, clang-tidy complains because of the `name` variable that is defined. I think in general `name` is too general to be used at that level.

andrsd (approved):

wdhawkins (approved):

PR #322, opened 2024-07-01
Improved volumetric source

zhardy-lanl:
This PR is improves the `DistributedSource` (now `VolumetricSource`) to allow for block ID-based sources. This is a first step towards removing `IsotropicMultiGroupSource` from `Materials` to avoid issues like defining multiple material zones with identical xs sets to turn on a source. Instead of integrating this into tests now, IMO it would be better to integrate this once the materials interface has changed. Otherwise, there will be two very large bulk changes. This has been tested on a few examples (remove `IsotropicMultiGroupSource` and replace with `VolumetricSource`), and tests still pass.

- A volumetric source can be assigned to a logical volume, a set of block IDs, or both. For the latter, the block IDs would serve as a mask within a logical volume where only cells of a given block ID within a logical volume contain the source.
- A volumetric source can be defined using *either* a `SpatialMaterialFunction` or a group-wise source strength vector. If both are specified, an error is thrown.
- The preexisting default behavior of a unit source in all groups when no function or group-wise source strength vector is provided is kept.
- The inner machinery of the class functions the same as before, where upon the solver's initialization, where the volumetric source gets populated with the cell IDs that it belongs to.


lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_sources.cc
```diff
@@ -55,16 +55,16 @@ RegisterLuaFunctionInNamespace(AddDistributedSource, lbs, AddDistributedSource);
 int
 AddDistributedSource(lua_State* L)
 {
-  const std::string fname = "lbs.AddDistributedSource";
+  const std::string fname = "lbs.AddVolumetricSource";
```
ragusa, line 58: Volumetric implies space-dependent. The source is also already group-dependent. Eventually, it would be good to have the source be direction-dependent. So, given this, wasn't distributed better than volumetric? maybe there is something better than distributed.

modules/linear_boltzmann_solvers/lbs_solver/source_functions/source_function.cc
```diff
@@ -220,10 +220,10 @@ SourceFunction::AddPointSources(const LBSGroupset& groupset,
 }
 
 void
-SourceFunction::AddDistributedSources(const LBSGroupset& groupset,
-                                      std::vector<double>& q,
-                                      const std::vector<double>&,
-                                      const SourceFlags source_flags)
+SourceFunction::AddVolumetricSources(const LBSGroupset& groupset,
+                                     std::vector<double>& q,
+                                     const std::vector<double>& phi,
```
ragusa, line 225: just curious, phi is not used below, is it?

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
```diff
@@ -0,0 +1,125 @@
...
+InputParameters
+VolumetricSource::GetInputParameters()
+{
+  InputParameters params = Object::GetInputParameters();
+
+  params.SetGeneralDescription("General implementation of an arbitrary multi-group "
+                               "distributed source. A distributed source is given by a "
+                               "logical volume and a Lua function handle. Currently, only "
+                               "isotropic distributed sources are allowed.");
+  params.SetClassName("Distributed Source");
```
ragusa, line 27: distributed was forgotten in the replace all

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
```diff
@@ -0,0 +1,125 @@
...
+                                   "An array of multi-group source strength values. Note that this "
+                                   "is only used when a function is not provided.");
+  params.AddOptionalParameter(
+    "logical_volume_handle",
+    SIZE_T_INVALID,
+    "Handle to the logical volume the volumetric source is defined within.");
+  params.AddOptionalParameter(
+    "function_handle",
+    SIZE_T_INVALID,
+    "Handle to a ResponseFunction object to be used to define the source.");
```
ragusa, line 45: why response ?

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
```diff
@@ -0,0 +1,125 @@
...
+  log.Log() << "Volumetric source has " << num_global_subscribers_ << " total subscribing cells.";
+}
+
+std::vector<double>
+lbs::VolumetricSource::operator()(const Cell& cell, const Vector3& xyz, const int num_groups) const
+{
+  if (std::count(subscribers_.begin(), subscribers_.end(), cell.local_id_) == 0)
+    return std::vector<double>(num_groups, 0.0);
+  else if (not function_)
+    return strength_.empty() ? std::vector<double>(num_groups, 1.0) : strength_;
```
ragusa, line 120: I do not think we should allow default options. If no functions are given and strength is empty, then complaint. I think the user must enter a value/array.

test/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_3.lua
```diff
@@ -162,15 +162,15 @@ function ResponseFunction(xyz, mat_id)
 end
 response_func = opensn.LuaSpatialMaterialFunction.Create({ lua_function_name = "ResponseFunction" })
```
ragusa, line 163: is there some sort of default ot be a ResponseFunction? If so, should be just be a bit more generic?

ragusa, line 163: [repeat from another comment] also the name SpatialMaterialFunction feels weird now. When doing the adjoint, the distributed/volumetric source was often associated with a material (the "detector cross section"). But now that distributed sources can be used in forward mode, I do not think that associating a source with a material is what we wanted to keep. It is now a bit awkward to me to read SpatialMaterialFunction here now.

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
```diff
@@ -0,0 +1,125 @@
...
+
+VolumetricSource::VolumetricSource(const InputParameters& params)
+  : block_ids_(params.GetParamVectorValue<int>("block_ids")),
+    logvol_(params.ParametersAtAssignment().Has("logical_volume_handle")
+              ? GetStackItemPtrAsType<LogicalVolume>(
+                  object_stack, params.GetParamValue<size_t>("logical_volume_handle"))
+              : nullptr),
+    strength_(params.GetParamVectorValue<double>("group_strength")),
+    function_(params.ParametersAtAssignment().Has("function_handle")
+                ? GetStackItemPtrAsType<SpatialMaterialFunction>(
```
ragusa, line 58: also the name SpatialMaterialFunction feels weird now. When doing the adjoint, the distributed/volumetric source was often associated with a material (the "detector cross section"). But now that distributed sources can be used in forward mode, I do not think that associating a source with a material is what we wanted to keep. It is now a bit awkward to me to read SpatialMaterialFunction here now.

ragusa (changes requested):

ragusa:
Should an example be added tot he tests where source assignment is done using blockID and not LV ?

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_sources.cc
zhardy-lanl: I axed distributed because calling something "distributed" in a parallel code to avoid any confusion between physics and computer science language. Volumetric does imply space-dependent, and this is a space-dependent source. I agree that it would be good to have a direction dependent source. I am open to proposals for names, but I think in most cases if you say volumetric source, it will be understood as the right-hand side `Q` in the transport equation.

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
zhardy-lanl: Good catch!

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
zhardy-lanl: That was the old name of the function type. Changing to `SpatialMaterialFunction`. The original implementation of this class was tied to the adjoint solver, which is where that came from originally.

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
zhardy-lanl: This was the default behavior before this PR, which was used in some of the tests. I will make the change.

test/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_3.lua
zhardy-lanl: I'm not quite sure what you mean.

ragusa: IT's about this part: `lua_function_name = "ResponseFunction"`

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
zhardy-lanl: > also the name SpatialMaterialFunction feels weird now. 

`SpatialMaterialFunction` just describes the arguments to an `Evaluate` call in the function system. It takes a cell, a node, a material(block) ID, and a number of components. The `SpatialMaterialFunction` has been an attribute of the `Distributed(Volumetric)Source` since the function system was put into the code awhile back. 

> When doing the adjoint, the distributed/volumetric source was often associated with a material (the "detector cross section"). But now that distributed sources can be used in forward mode, I do not think that associating a source with a material is what we wanted to keep. 

The volumetric source in its previous state was applied to cells within a logical volume. The user could pass a function that performs operations based on the arguments stated above that could 1) provide spatial and group-dependence within the logical volume, and 2) filter for particular material/block IDs. If was not function is passed, then the default behavior of a unit source in each group within the logical volume was used. This is where the tie to materials came into play. It has nothing to do with the `Material` object, like `IsotropicMultiGroupSource` does.

I'm not sure exactly what you mean by "I do not think that associating a source with a material is what we wanted to keep" since the `VolumetricSource` is associated with block IDs from the mesh and/or logical volumes not `Material` objects. Anything other than this would be a departure from how sources are currently specified in the code. As I state in the description, this PR is meant to move the functionality of `IsotropicMultiGroupSource` out of something that lives on `Material`, not to redo the source specification system.

> It is now a bit awkward to me to read SpatialMaterialFunction here now.

Do the above comments make this less awkward? Otherwise, I'm not sure how to address this.

zhardy-lanl:
> Should an example be added tot he tests where source assignment is done using blockID and not LV ?

Since the stated purpose of this PR is for `VolumetricSource` to take the functionality of `IsotropicMultiGroupSource`, I could create a test identical to another that uses `VolumetricSource` with block IDs instead. What are your thoughts on this? 

What I didn't want to do was to go change every single test to this when there is still a considerable amount of work to do before we have a finalized `Material` independent source specification system (i.e. it would be a huge change that would not last long)

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
ragusa: We are trying to undo the connection of source->material object (big scheme of things). In that process, I do not think we should have the connection between a volumetric source and either an LV or a set of blockIDs (using the future name here) be called SpatialMaterialFunction. The source is in a volume, which is identified by an LV or block ID. It is the Material portion of the name that bugs me.

zhardy-lanl: I had typed up a response and thought it submitted but it was marked as "pending" in a review I must have accidentally started. I'm not entirely sure what the particular change you are asking for is. I understand your concern, but how would you like me to address it?  Change the name of `SpatialMaterialFunction`, or implement a new function in the function system that only takes a spatial coordinate called something like `SpatialFunction` (along with a `LuaSpatialFunction` counterpart)? If the prior, what would you suggest changing the name to that contains the same information? `SpatialMaterialIDFunction`? `SpatialBlock(ID)Function`?

test/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_3.lua
zhardy-lanl: This comment was again typed up yesterday but left in a "pending" state in a review.

The argument `lua_function_name` must correspond to the name of a defined Lua function in the script. In this case, `ResponseFunction` is the name of the function defined directly above this line. If it was called something else, the argument would have to be changed. By default `lua_function_name` is empty. In the previous version, this resulted in the default behavior of a unit source within the logical volume (left over from the original adjoint implementation to default to a volumetric QoI). In the current implementation, if `lua_function_name` is empty, a `group_strength` vector must be supplied. When a `group_strength` vector is specified, this acts as a constant source within the logical volume and/or block IDs.

zhardy-lanl:
The adjoint tests have now been modified to include the new `VolumetricSource` functionality. In `response_2d_1.lua`, the `IsotropicMultiGroupSource` is swapped with a `VolumetricSource` set based off of block IDs. The other two tests are modified to account for the default unit source being removed.

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_sources.cc
ragusa: Comment. Ok, volumetric is clear. I think it is already group dependent. Do you think MG_volumetric could make it even more obvious? (Any version that includes mg or multigroup may have my slight preference)

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
ragusa: Request (I think). It is the Material portion of the function name that feels strange to me. The source will be turned on based on a LV or a set of (soon to be named) block IDs. So the Material portion of the name is what I was referring to

test/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_3.lua
zhardy-lanl: Oh, I see. The function is only used to define the source for the adjoint problem. The function is defined after the forward solve is done, so it is not used there.

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
zhardy-lanl: We discussed this last Wednesday and I will be modifying the function system class `SpatialMaterialFunction` to remove the material ID function input and rename it to something like `SpatialFunction` or `VectorSpatialFunction` to mirror `ScalarSpatialFunction` which outputs a scalar value.

I should reiterate, however, that `SpatialMaterialFunction` is not something defined in this PR or within the `VolumetricSource` class. It is a generic function in the function system. Its use here was due to the fact that it was the only available function in the function system that had the desired input/output behavior. While I don't think that we should be renaming generic objects in `framework` based on individual use cases, because there are no other use cases for `SpatialMaterialFunction` in the code and we do not have a more general function the returns a vector from  a spatial coordinate argument, this modification is worthwhile here.

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_sources.cc
zhardy-lanl: At this point, I think that it is okay to leave off multigroup. The old source `IsotropicMultiGroupSource` was housed in `framework`, making the extra clarification more necessary. Because `VolumetricSource`, like `PointSource`, is housed within the LBS module, in my mind multigroup is essentially implied. For example, `PointSource` is not `MultiGroupPointSource`. Eventually, if we would like to extend this to be arbitrary in angle as well, it may be worthwhile to rename `VolumetricSource` -> `IsotropicVolumetricSource` and `PointSource` -> `IsotropicPointSource` to make room for the generic `VolumetricSource` and `PointSource` classes that are arbitrary in angle.

andrsd, line 58: Should `AddDistributedSource` (i.e. the function name) be changed to `AddVolumetricSource`. Also, the same name change for the lua function? This seems to be changing the `fname` which is what the user would see in case of an error.

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_sources.cc
```diff
@@ -74,15 +74,15 @@ RegisterLuaFunctionInNamespace(ClearDistributedSources, lbs, ClearDistributedSou
 int
 ClearDistributedSources(lua_State* L)
 {
-  const std::string fname = "lbs.ClearDistributedSources";
+  const std::string fname = "lbs.ClearVolumetricSources";
```
andrsd, line 77: The same thing here: `fname` is changed but not the function names...

andrsd (changes requested):

andrsd (approved):

wdhawkins (approved):

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_sources.cc
ragusa: I agree on leaving out MG since this is housed in `lbs`. I like the addition of Isotropic but because this needs to happen for both volumetic and point sources, it could be preferable to do that renaming in a new PR

modules/linear_boltzmann_solvers/lbs_solver/volumetric_source/volumetric_source.cc
ragusa: I believe I knew that `SpatialMaterialFunction` was not new to the PR. It was just that the name did not make sense (or no longer made sense).

ragusa (approved):

Issue #321, opened 2024-06-29
groupsets issue with k-eigenvalue simulations


ragusa:
I run a problem with 172g XS obtained from OpenMC. I do not get the same answers when using a single group set versus more group sets.
GS-n: n groupsets
MPI-r: r ranks
```
GS-1, MPI-1  : keff = 1.179888
GS-1, MPI-12: keff = 1.179893

GS-3, MPI-1. : keff = 1.022954
GS-3, MPI-12: keff = 1.714134
```

With more mesh cells and more directions, the symptoms are identical. I am attaching the smaller mesh size problem. It runs quickly.
[gs.tar.gz](https://github.com/user-attachments/files/16041650/gs.tar.gz)

ragusa:
I have looked in the regression suite, we do not have a single test case that uses more than 1 group set.

wdhawkins:
Ugh... this strikes me as a difficult one.  To start with, I can easily get a segfault with different groupset structures. I'll get that fixed first. Hopefully it's the root cause of all of our problems. We'll see.

PR #323, opened 2024-07-05
Bug fixes for multi-groupset, k-eigenvalue problems with reflecting boundaries

wdhawkins:
This PR fixes multiple issues with multi-groupset k-eigenvalue problems with reflecting boundaries. The default k-eigenvalue solver was incorrectly setting active sources for only the first groupset. In addition, angular flux storage for reflecting boundaries was not sized properly for the current implementation (the implementation can be improved, but that is beyond the scope of this PR).  Resolves issue #321.

### DIFF
--- a/modules/linear_boltzmann_solvers/executors/pi_keigen.cc
+++ b/modules/linear_boltzmann_solvers/executors/pi_keigen.cc
@@ -148,12 +148,19 @@ PowerIterationKEigen::Execute()
     WriteRestartData();
 
   // Print summary
+  int total_num_sweeps = 0;
+  for (auto& wgs_solver : lbs_solver_.GetWGSSolvers())
+  {
+    auto context = wgs_solver->GetContext();
+    auto wgs_context = std::dynamic_pointer_cast<lbs::WGSContext>(context);
+    total_num_sweeps += wgs_context->counter_applications_of_inv_op_;
+  }
+
   log.Log() << "\n";
   log.Log() << "        Final k-eigenvalue    :        " << std::setprecision(7) << k_eff_;
   log.Log() << "        Final change          :        " << std::setprecision(6) << k_eff_change
-            << " (Number of Sweeps:" << front_wgs_context_->counter_applications_of_inv_op_ << ")"
-            << "\n";
-  log.Log() << "\n";
+            << " (Total number of sweeps:" << total_num_sweeps << ")"
+            << "\n\n";
 
   if (lbs_solver_.Options().use_precursors)
   {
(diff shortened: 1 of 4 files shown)


modules/linear_boltzmann_solvers/executors/pi_keigen.cc
```diff
@@ -172,11 +172,14 @@ PowerIterationKEigen::SetLBSFissionSource(const std::vector<double>& input, cons
   if (not additive)
     Set(q_moments_local_, 0.0);
 
-  active_set_source_function_(front_gs_,
```
zhardy-lanl, line 175: I don't see a reason for any of the members named `front_<...>`. Should we just get rid of them?

wdhawkins: `front_wgs_context` is used to print the number of sweeps (which I'm not sure we even need to do). For this type of problem, I'm not sure it's a interesting statistic.

zhardy-lanl: Either way, that could be done without `front_wgs_context` as a member though. Also, shouldn't we sum the number of sweeps in each groupset instead? Why would the first one contain information cumulative for all of them?

wdhawkins: Yes, it should be the sum. I'm going to remove that anyway... as I said, I don't think it's a useful stat here. However, `pi_keign_smm` and `pi_keigen_scdsa` depend on those member variables, so they can't be removed. I think `smm` checks for a single groupset, but I don't know about `scdsa`. If it doesn't, I'll add that.

zhardy-lanl: Awesome. Yep, the others do (somewhat unnecessarily) because they are only implemented for a single groupset. That's a future issue to tackle though.

zhardy-lanl: I hate to say this after you removed it... But I forgot that I used print statement about sweeps that was removed in my SMM paper. Is there any way that could be added back in to print the sum of sweep counts over groupsets? It was useful to have to compare how many fewer sweeps you need using acceleration schemes. Sorry I did not think of that before

wdhawkins: No problem. Fixed.

zhardy-lanl (approved):

ragusa (approved):

PR #325, opened 2024-07-07
Refactor AGS solver

wdhawkins:
This PR refactors the AGS solver and grew out of fixes for mult-groupset problems. The existing AGS solver implementation seemed only half-finished to me and was hard-coded to use PETSc Krylov-Richardson iteration with a maximum iteration count of 1. I didn't see any way for the user to adjust the maximum iteration count or set an iteration tolerance. It also seems that an effort was made to support some type of multi-AGS solver scheme - but I don't believe it was ever finished (and I don't see why we would want to do this). 

The PR implements the following fixes:

1. Removes the multi-AGS solver infrastructure
2. Adds the ability for the user to set AGS tolerance and maximum number of iterations
3. Simplifies the AGS solver by using classic Richardson instead of a PETSc solver (this simplifies and reduces code and is leaner memory-wise)

This PR is currently a draft to get everyone's thoughts on the implementation. If there is a desire to support an iterative method other than Richardson (say GMRES), I can revert to using PETSc for the AGS solver.


modules/linear_boltzmann_solvers/discrete_ordinates_adjoint_solver/lbs_adj_solver.cc
```diff
@@ -76,8 +76,7 @@ DiscreteOrdinatesAdjointSolver::Initialize()
 void
 DiscreteOrdinatesAdjointSolver::Execute()
 {
-  primary_ags_solver_->Setup();
-  primary_ags_solver_->Solve();
+  ags_solver_->Solve();
```
zhardy-lanl, line 79: This applies to multiple places, but why is `ags_solver_->Setup();` not needed?

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
```diff
@@ -12,55 +12,38 @@ namespace lbs
 {
 
 /**Linear Solver specialization for Within GroupSet (WGS) solves.*/
```
zhardy-lanl, line 14: This comment should be changed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
```diff
@@ -12,55 +12,38 @@ namespace lbs
...
-class AGSLinearSolver : public LinearSolver
+class AGSLinearSolver
 {
+private:
+  std::shared_ptr<LinearSolverContext> context_ptr_;
+  int maximum_iterations_;
+  double tolerance_;
+  bool verbose_;
```
zhardy-lanl, line 21: I forgot whether we decided on any format for class layout, but attributes are at the bottom of class definitions in most cases.

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
```diff
@@ -230,6 +229,8 @@ struct Options
   bool verbose_inner_iterations = true;
   bool verbose_ags_iterations = false;
   bool verbose_outer_iterations = true;
+  int maximum_ags_iterations = 1;
```
zhardy-lanl, line 232: Maybe `max` instead of `maximum`?

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -462,6 +462,9 @@ LBSSolver::OptionsBlock()
     "verbose_inner_iterations", true, "Flag to control verbosity of inner iterations.");
   params.AddOptionalParameter(
     "verbose_outer_iterations", true, "Flag to control verbosity of across-groupset iterations.");
+  params.AddOptionalParameter(
```
zhardy-lanl, line 465: Maybe `max` instead of `maximum` for smaller input parameter names?

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
```diff
@@ -12,55 +12,38 @@ namespace lbs
...
+  void SetVerbosity(bool verbose) { verbose_ = verbose; }
 
-  void SetSystem() override;
+  double GetTolerance() { return tolerance_; }
 
-  void SetPreconditioner() override;
+  void SetTolerance(double tolerance) { tolerance_ = tolerance; }
 
-  void SetRHS() override;
+  int GetMaximumIterations() { return maximum_iterations_; }
```
zhardy-lanl, line 44: Maybe `Max` instead of `Maximum` to be in line with other comments.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
```diff
@@ -12,55 +12,38 @@ namespace lbs
...
-  void SetVerbosity(bool verbose_y_n) { verbose_ = verbose_y_n; }
-
-  void Solve() override;
-
-protected:
-  void SetSystemSize() override;
+  void SetVerbosity(bool verbose) { verbose_ = verbose; }
 
-  void SetSystem() override;
+  double GetTolerance() { return tolerance_; }
```
zhardy-lanl, line 40: Curious to know everyone's thoughts on `GetTolerance` vs. `Tolerance` for getters broadly.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
```diff
@@ -12,55 +12,38 @@ namespace lbs
 {
 
 /**Linear Solver specialization for Within GroupSet (WGS) solves.*/
-class AGSLinearSolver : public LinearSolver
+class AGSLinearSolver
 {
+private:
+  std::shared_ptr<LinearSolverContext> context_ptr_;
```
zhardy-lanl, line 18: This would be a great opportunity to get rid of an unnecessary context. There is no reason we need a context to store a list of `WGSSolvers`. This can be stored in this class.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
```diff
@@ -12,55 +12,38 @@ namespace lbs
 {
 
 /**Linear Solver specialization for Within GroupSet (WGS) solves.*/
-class AGSLinearSolver : public LinearSolver
+class AGSLinearSolver
```
zhardy-lanl, line 15: Unless there is a notion of a nonlinear AGS solver, I would probably go with `AGSSolver`.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.cc
```diff
@@ -16,107 +17,77 @@ namespace opensn
...
-  // This is necessary for multiple ags iterations to function
-  // and for keigen-value problems
+  auto& wgs_solvers = ags_context_ptr->wgs_solvers_;
+  auto& cell_transport_views = lbs_solver.GetCellTransportViews();
+  auto& groupsets = lbs_solver.Groupsets();
+  auto& phi_new = lbs_solver.PhiNewLocal();
+  auto num_moments = lbs_solver.NumMoments();
+  std::vector<double> phi_old(lbs_solver.PhiOldLocal().size(), 0.0);
+
+  auto piecewise_change =
```
zhardy-lanl, line 35: This feels like it should live somewhere else to me

modules/linear_boltzmann_solvers/discrete_ordinates_adjoint_solver/lbs_adj_solver.cc
wdhawkins: `Setup` is really specific to the PETSc solvers, doing a lot of PETSc things that aren't needed for this solver... at least if we're using Classic Richardson.

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
wdhawkins: Works for me. I'll change that.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
wdhawkins: Good point. That's all it does here. Let me see about removing it completely.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.cc
wdhawkins: I could make it a separate method in `AGSLinearSolver`, but I end up having to do a bunch of variable assignments and things that had already been done in `Solve`. In Chi-Tech, a very similar method lived in `LBSSolver` and was used for the Classic Richardson WGS routine. I could move it there. To be honest, I'm not a fan of any of the solutions. Maybe a separate routine in `AGSLinearSolver` is best?

zhardy-lanl: Hmm... Well here's my first question: Where will something like this ultimately be used? It feels like it should be available to the `WGSLinearSolver` as well, doesn't it? If yes, I'd say that `LBSSolver` makes the most sense.

As a side note, the loop over groupsets is not entirely necessary here. Flux moments are stored for all groups so this could be a single loop over `g < lbs_solver.NumGroups()` instead.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
wdhawkins: I've removed AGSContext.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.h
wdhawkins: Personally, I've always preferred them without Get/Set. In this case, I was just following what was already there.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_linear_solver.cc
wdhawkins: Fixed. I made the pointwise check a private member of the AGSSolver class rather than move it into LBSSolver. It's unique in the fact that it computes the pointwise change between LSBSolver::PhiNewLocal() and a private AGSSolver member variable, so that seemed the appropriate place.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
```diff
@@ -0,0 +1,57 @@
...
+{
+
+class LBSSolver;
+
+/**Solver for Across-Groupset (AGS) solves.*/
+class AGSSolver
+{
+public:
+  /**Constructor.
+   * \param ags_context_ptr Pointer Pointer to the context to use.
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
```diff
@@ -0,0 +1,57 @@
...
+   * \param verbose bool Flag to enable verbose output.*/
+  AGSSolver(LBSSolver& lbs_solver, std::vector<std::shared_ptr<LinearSolver>> wgs_solvers)
```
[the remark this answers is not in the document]
wdhawkins: The vector of `wgs_solvers` contains the wgs solver for each groupset. An ags solve is really just a loop over the groupsets that executes the wgs solver for that groupset.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -0,0 +1,121 @@
...
+              double delta_phi = std::fabs(phi_new[mng_idx] - phi_old[mng_idx]);
+              if (max_phi >= std::numeric_limits<double>::min())
+                pw_change = std::max(delta_phi / max_phi, pw_change);
+              else
+                pw_change = std::max(delta_phi, pw_change);
+            } // for g
+          }   // for m
+        }     // for groupset
+      }       // for i
+    }         // for c
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
```diff
@@ -0,0 +1,57 @@
...
+
+  void SetVerbosity(bool verbose) { verbose_ = verbose; }
+
+  double GetTolerance() { return tolerance_; }
+
+  void SetTolerance(double tolerance) { tolerance_ = tolerance; }
+
+  int GetMaxIterations() { return max_iterations_; }
+
+  void SetMaxIterations(int max_iterations) { max_iterations_ = max_iterations; }
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -0,0 +1,121 @@
...
+    double global_pw_change = 0.0;
+    mpi_comm.all_reduce<double>(pw_change, global_pw_change, mpi::op::max<double>());
+    return global_pw_change;
+  };
+
+  // Save qmoms to be restored after each iteration. This is necessary for multiple ags iterations
+  // to function and for keigen-value problems
+  const auto saved_qmoms = lbs_solver_.QMomentsLocal();
```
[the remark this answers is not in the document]
wdhawkins: This is a holdover from the previous implementation. I haven't actually tried to run without it. I'll have to dig a little deeper to see what's going on here.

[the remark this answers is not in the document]
ragusa: @wdhawkins I will let you resolve this conversation when you are ready. In the meantim, I will approve the changes

ragusa:
@wdhawkins should we had a test for AGS iteration (before I go ahead and approve)?

wdhawkins:
> @wdhawkins should we had a test for AGS iteration (before I go ahead and approve)?

Yes, we should. I haven't had a chance to create one. Do you know if there's an existing test that we could easily adapt for this?

andrsd (approved):

wdhawkins:
I've added a 3g/3-groupset test with manufactured cross-section data that requires iterating on the groupset-to-groupset scattering to get the single groupset answer.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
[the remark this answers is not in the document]
wdhawkins: I've looked into this, and, yes, it is necessary. The core of the AGS solver is essentially a loop over groupets that invokes the WGS solver for each groupset. In a nutshell, the WGS solver overwrites `q_moments_local`, and it needs to be restored after the loop over WGS solvers and before the next AGS iteration. Since the core of PI is the invocation of the AGS solver, the same logic applies.

ragusa (approved):

modules/linear_boltzmann_solvers/lbs_solver/lbs_structs.h
```diff
@@ -230,6 +229,8 @@ struct Options
   bool verbose_inner_iterations = true;
   bool verbose_ags_iterations = false;
   bool verbose_outer_iterations = true;
+  int max_ags_iterations = 1;
```
zhardy-lanl, line 232: Were we going to change this default value after the discussion last Wednesday?

wdhawkins: Let me look at this a bit. I can default it to something like 10 which may be a better option.

wdhawkins: I'm going to change the default to 10 iterations. This is going to change the answers for some of the regression tests. I'll make sure that each test that this affects converges the upscattering to a decent value.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "framework/logging/log.h"
+#include "framework/runtime.h"
+#include "caliper/cali.h"
+#include <petscksp.h>
+#include <iomanip>
+
+namespace opensn
+{
+namespace lbs
+{
```
andrsd, line 18: This needs to be removed now that #328 is in the `main`

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
```diff
@@ -0,0 +1,57 @@
...
+#pragma once
+
+#include "modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.h"
+#include "framework/utils/timer.h"
+#include <vector>
+
+namespace opensn
+{
+namespace lbs
+{
```
andrsd, line 13: Remove this as well.

andrsd (changes requested):

andrsd (approved):

PR #327, opened 2024-07-08
Updating configure_dependencies.py to PETSc 3.20.6

wdhawkins:
This PR updates configure_dependencies.py to use PETSc 3.20.6 and includes other minor changes to successfully build on macOS 14.5+.

### DIFF
--- a/resources/configure_dependencies.py
+++ b/resources/configure_dependencies.py
@@ -85,9 +85,14 @@ package_info = {
         "5.4.6",
         "https://www.lua.org/ftp/lua-5.4.6.tar.gz"
     ],
+    "hdf5": [
+        "1.14.3",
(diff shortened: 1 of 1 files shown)


ragusa (approved):
Why was fopenmp removed (or why could it have been there before)?

andrsd:
Apple does not support openmp at all. We usually get it if we are using homebrew, conda, etc. So, this could be it. Moreover, we are not using openmp anywhere and PETSc internally does not do threading either. They leave it to the application to deal with that. This is probably why we are able to remove this flag with no harm.

andrsd (approved):

wdhawkins:
David is correct. Apple clang doesn't support openmp, and it's not used in OpenSn or PETSc. With these changes, OpenSn and dependencies will compile with the Apple toolchain and a suitable MPI implementation.

zhardy-lanl (approved):
Looks like there is a Python formatting issue, FYI

Issue #57, opened 2023-12-13
Time dependent transient capability


ragusa:
This needs to be restored. and then the test in transport_transient should be made golden and added to the regression suite.

- [ ] rename `DiscOrdTransientSolver` to `DiscreteOrdinatesTransientSolver`
- [ ] get rid of `lbts_lua_utils` namespace

Issue #41, opened 2023-11-15
folder organization in module/LBS


ragusa:
there are diffusion solvers in 

- a_lbs_solver/acceleration 
- b_diffusion_solver

can this be harmonized?

zhardy-lanl:
This is a big can of worms that does need to be addressed sooner rather than later. The current code design is that `a_lbs_solver` is the broad and encompassing data structure that defines a transport or diffusion problem. All of the `b`, `c`, and `d` variants are specializations. The `LBSSolver` gets attached to an executor which then solves the problem.

There are several issues with the current design in my view:

1. `a_lbs_solver` is meant to be inherited by transport *and* diffusion specific data structures. There is a lot of data in `LBSSolver` and other classes within `a_lbs_solver`, however, that is transport specific. For example, DSA is tied to transport and not diffusion, yet `a_lbs_solver` contains all of the DSA related routines and not `b_discrete_ordinates_solver`. There are numerous other examples in this regard.
2. Our standalone diffusion solvers are essentially wrappers around the DSA functionality, which is the opposite of how it should be. Our DSA routines should create diffusion solvers instead.

This will ultimately require a meeting where we go through the `linear_boltzmann_solvers` directory in great detail and decide on a new strategy that makes more sense.

zhardy-lanl:
There are many different paths we can take on this topic... I'll propose a number of thoughts we should discuss:

1. Should we have a different diffusion solver for each discretization, or house the different discretizations in a single solver?
2. Should there be separate diffusion solvers for multi-group and single group?
3. Should we have standalone diffusion modules that get created by `DiscreteOrdinatesSolver` for DSA, or should `LBSSolver` define diffusion solvers for DSA that then get used by standalone modules?

I think if we can really nail down on these questions (and potentially others that I am not considering) it will  really clear up what needs to be done for this issue.

carlo-fiorina:
This seems the best place for this, instead of opening  a new issue.

_lbs_solver.h_ contains
_#include "modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/boundary/sweep_boundary.h"
#include "modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/sweep.h"_
However, a base class shouldn't normally need to know the details of derived classes. This violates the Open-Closed Principle (OCP)—the base class should remain open for extension but closed for modification. In other words, you should not need to modify the parent class when modifying the derived classes, or when creating new derived classes.

**Possible solution:**
[...]

PR #328, opened 2024-07-08
Removing nested namespaces

andrsd:
- Removing declaration in no-longer existing namespace
- Removing prk namespace
- Removing diffusion namespace
- Removing lbs namespace
- Removing nested lbs namespace from `opensnlua` namespace

Refs #44


lua/modules/diffusion/diffusion.cc
```diff
@@ -49,13 +49,12 @@ CFEMSetOptions(const opensn::InputParameters& params)
...
   const size_t handle = params.GetParamValue<size_t>("arg0");
-  auto& solver =
-    opensn::GetStackItem<opensn::diffusion::CFEMSolver>(opensn::object_stack, handle, fname);
+  auto& solver = opensn::GetStackItem<opensn::CFEMSolver>(opensn::object_stack, handle, fname);
   solver.SetDCoefFunction(d_coef_function);
   solver.SetQExtFunction(q_ext_function);
   solver.SetSigmaAFunction(sigma_a_function);
 
-  auto options_params = opensn::diffusion::CFEMSolver::OptionsBlock();
+  auto options_params = opensn::CFEMSolver::OptionsBlock();
```
zhardy-lanl, line 57: If we get rid of the `diffusion` namespace, these solvers should be called `CFEMDiffusionSolver`, etc.

lua/modules/linear_bolzmann_solvers/do_transient/lbts_lua_advance_time_data.cc
```diff
@@ -33,12 +33,12 @@ LBTSAdvanceTimeData(lua_State* L)
   const int solver_handle = lua_tointeger(L, 1);
 
   auto& solver =
-    opensn::GetStackItem<lbs::DiscOrdTransientSolver>(opensn::object_stack, solver_handle, fname);
+    opensn::GetStackItem<DiscOrdTransientSolver>(opensn::object_stack, solver_handle, fname);
```
zhardy-lanl, line 36: Ouch, never knew the unused transient solver is `DiscOrdTransientSolver`. We should change this to `DiscreteOrdinatesTransientSolver` when we get around to it

lua/modules/linear_bolzmann_solvers/do_transient/lbts_lua_create_solver.cc
```diff
@@ -5,7 +5,7 @@
 #include "framework/runtime.h"
 
 #if 0
-namespace lbs::lbts_lua_utils
+namespace lbts_lua_utils
```
zhardy-lanl, line 8: We should consider renaming these things like `lbts` at some point since they no longer make a whole lot of sense with the solver being `DiscOrdTransientSolver`. Future work, not here.

zhardy-lanl (changes requested):

lua/modules/diffusion/diffusion.cc
andrsd: I think that would happen as part of #41

lua/modules/linear_bolzmann_solvers/do_transient/lbts_lua_advance_time_data.cc
andrsd: I added this as a task into #57 which is dedicated to transient capability

lua/modules/linear_bolzmann_solvers/do_transient/lbts_lua_create_solver.cc
andrsd: Also added into #57 as a task

lua/modules/diffusion/diffusion.cc
zhardy-lanl: I disagree. That issue has to do with refactoring the diffusion solver implementation, not the naming convention. Here, removing the namespace makes the class name completely meaningless. IMO, the name changes should be done here.

zhardy-lanl: Yikes, both options are bad... @wdarylhawkins @ragusa What are your thoughts on this?

andrsd: Yeah, it is a big can of worms. But, I think it about time we do something about that. It an architectural change, so the sooner that happens the better...

andrsd: Maybe, we can start adding ideas what/how to do this to #41. I have a vague idea what needs to happen, but something more concrete would help...

zhardy-lanl: This will be a good topic for the meeting tomorrow. I will try to take a deeper look and add some thoughts on it before then.

ragusa: @andrsd @zhardy-lanl. 
- Wish list: the code can solve the diffusion approximation, with CG and DG FEM. This could be useful to initialize a transport (k- or source-driven) solve but it could also be the only solve a user does if they are ok with diffusion. Whether this leaves inside the lbs module or not is a design decision that I likely have no preference for. 
- the ability of doing a manufactured solution is very important to me. So having a diffusion solver with the capability is important to me. I found bugs in the MIP-DSA when I coded the standalone DFEM Diffusion solver. There could be a few left. I really need to stress test the DSA and compare the results from my previously published results. I have found the OpenSn DSA to be less effective than what I used to experience in my standalone codes. 
- The multigroup diffusion code already has the two-grid acceleration. This is useful and should not be lost. 
- I contributed a lot to this mess and I apologize. If the design decision is to keep a diffusion solver outside of lbs and diffusion accelerators inside of lbs, then doing manufactured solutions for the accelerators is not required. Some code may be duplicated but I am not sure if this is tolerable or not.

andrsd: I moved this comment to [issue #332](https://github.com/Open-Sn/opensn/issues/332#issuecomment-2222894600)

andrsd: As for this PR we agreed that the originally suggested name change should happen with this PR. Any further changes will be done in individual issues.

ragusa (approved):

lua/modules/diffusion/diffusion.cc
zhardy-lanl: Has this name change been done?

andrsd:
Rebased on the latest `main` to resolve conflicts.

wdhawkins (approved):

lua/modules/diffusion/diffusion.cc
zhardy-lanl: Oh I do now. Sorry!

zhardy-lanl (approved):

Issue #330, opened 2024-07-10
Add coordinate system information to the mesh


zhardy-lanl:
Currently, only the discretization contains coordinate system information.

wdhawkins:
Closed by #583

PR #333, opened 2024-07-10
Removed discretization requirement for `FieldFunctionGridBased`

zhardy-lanl:
This PR adds constructors and input options to make the default behavior of `FieldFunctionGridBased` default to a finite volume representation. The `discretization` input parameter is now defaulted to "FV" and constructors which require a grid instead of a discretization. 

Due to a lack of coordinate system information on the mesh, this currently only works for Cartesian grids. This will be changed with the resolution of #330.

### DIFF
--- a/framework/field_functions/field_function.h
+++ b/framework/field_functions/field_function.h
@@ -15,7 +15,7 @@ class FieldFunction : public Object
 {
 private:
   std::string text_name_;
-  opensn::Unknown unknown_;
+  Unknown unknown_;
   UnknownManager unknown_manager_;
 
 public:
@@ -34,7 +34,7 @@ public:
   /**Returns the text name of the field function.*/
   const std::string& TextName() const { return text_name_; }
   /**Returns a reference to the unknown structure.*/
-  const opensn::Unknown& Unknown() const { return unknown_; }
+  const Unknown& GetUnknown() const { return unknown_; }
   /**Returns a reference to the unknown manager that can be used in
    * spatial discretizations.*/
   const UnknownManager& GetUnknownManager() const { return unknown_manager_; }
--- a/framework/field_functions/field_function.h
+++ b/framework/field_functions/field_function.h
@@ -13,40 +13,34 @@ class Cell;
 
 class FieldFunction : public Object
 {
-private:
-  std::string text_name_;
-  Unknown unknown_;
-  UnknownManager unknown_manager_;
-
 public:
-  /**Returns required input parameters.*/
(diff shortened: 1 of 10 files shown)


framework/field_functions/field_function_grid_based.h
```diff
@@ -18,109 +18,78 @@ namespace opensn
...
 
-  /**Returns required input parameters.*/
   static InputParameters GetInputParameters();
-
-  /**ObjectMaker based constructor.*/
   explicit FieldFunctionGridBased(const InputParameters& params);
 
-  /**Creates a field function, filling it with zeros.*/
-  FieldFunctionGridBased(const std::string& text_name,
+  /// Creates a field function, filling it with zeros.
```
andrsd, line 29: Coding standard says to use `/** */` over `///`. But, I do see significant use of `///` mainly for single line comments. Maybe an update to the coding standard? What do you guys think?

My personal preference is:
```
/**
 * text
 */
method(params);

/// Description
type variable;
```
Because typically API needs more text that the variable. It is rarely the case where variables need more than 1 line and if so, multiline `///` does the trick.

framework/field_functions/field_function.h
```diff
@@ -13,40 +13,34 @@ class Cell;
...
-  FieldFunction(const std::string& text_name, Unknown unknown);
+  FieldFunction(std::string name, Unknown unknown);
 
   virtual ~FieldFunction() = default;
 
-  // Getters
-  /**Returns the text name of the field function.*/
-  const std::string& TextName() const { return text_name_; }
+  /// Returns the text name of the field function.
+  const std::string& Name() const { return name_; }
```
andrsd, line 25: :+1: on the change from `TextName` to `Name`.

andrsd (approved):

framework/field_functions/field_function_grid_based.h
andrsd: > Does `/** */` on one line fit within the coding standards?

It does. And it is used in the code quite a bit as well. If others agree, we can amend the coding standard to allow `///` for single line comments...

wdhawkins (approved):

framework/field_functions/field_function_grid_based.h
ragusa: I am fine with both `/** */~ and `///`

ragusa (approved):
I approve. There is one conversation where I added my 2 cents regarding the one-line comments

PR #334, opened 2024-07-10
Simple auxiliary variable infrastructure

zhardy-lanl:
This PR implements a simple auxiliary variable infrastructure. 

- The base solver class now stores a map called `aux_field_functions_` that maps a field name to a field function.
- Auxiliary variables are added to the solver via the `auxvars` input parameter. This parameter is an array of string names that must correspond to the name of an existing field function. For now, this requires a user to define any auxiliary field functions in the input file before creating the solver.
- For a solver to use an auxiliary variable, there must be some piece of code that 1) checks for a specific field name in `aux_field_functions_`, and 2) implements code to consume and use the auxiliary field. This feature would require specific field names (i.e. a user could not name a field something arbitrary).

I am putting this draft PR up to get feedback on this implementation. What I am ultimately trying to do is to remove the densities vector from `LBSSolver`, but keep the same existing functionality.


framework/physics/solver.cc
```diff
@@ -212,4 +215,19 @@ Solver::SetProperties(const ParameterBlock& params)
...
 }
 
+void
+Solver::SetAuxiliaryFieldFunction()
+{
+  for (const auto& auxvar : auxvars_)
+  {
+    for (auto& ff : field_function_stack)
+    {
+      if (auxvar == ff->TextName())
```
ragusa, line 225: in the FF, we still use TextName(). Should it be changed to Name() later or now?

andrsd:
This looks ok to me...

framework/physics/solver.cc
zhardy-lanl: This may have been changed in my previous PR #333 and need to be updated here.

PR #336, opened 2024-07-13
Refactoring regression test scripts

### LINKED ISSUE #276: Allow for multiple tests with one script
Modify the test system to allow for tests based off the same script. In particular, this is useful for having tests for different methods on the same problem.

Refs #272

wdhawkins:
This PR refactors the regression test script and addresses some bugs and limitations. In addition to cosmetic changes, the following issues are addressed:

1. System capacity is now automatically calculated unless overridden by the user using the `-j` option. This should provide for faster testing out-of-the-box on most systems.
2. Multiple tests using the same input file with different lua arguments is now supported. In this case, lua arguments are included in parentheses after the input filename on the Pass/Fail line.
4. Resolved an issue that could result in hangs or job failures if a particular test generated a large amount of output. In this case, the output could exceed the memory buffers for `stdout` and `stderr` in the call to `Popen`. This happens frequently on weekly tests with C5G7. `Popen` now writes `stdout` and `stderr` directly to files that are then concatenated into a single output file.

Please review this PR carefully! I don't want to introduce any bugs into our regression testing.

Resolves #276.


test/run_tests
```diff
@@ -48,7 +40,7 @@ parser.add_argument(
     "-d", "--directory",
     type=str, required=False,
     default=os.path.abspath(os.path.dirname(__file__)),
-    help="The test directory to process"
+    help="The directory to search for test files"
 )
 
 parser.add_argument(
```
zhardy-lanl, line 46: I've always not liked the `-t` option requiring `-d` and then taking the filename for the `-t` argument. Rather, `-t` accepting a path would be much nicer (tab complete). Is that change too big to add to this PR? I'm not sure how this would work for the C5G7 tests that are distinguished by a command line argument, but that problem pre-dated this PR.

wdhawkins: Let me see what that would take to fix. For something like C5G7, if you specify the test file using `-t`, it will run the file with all of the lua argument options.

wdhawkins: Fixed. `--test/-t` now takes the full path to a test file and doesn't require `--directory/-d`. Tab completion should work as well as the shell will take care of that.

test/run_tests
```diff
@@ -111,59 +107,55 @@ argv.exe = os.path.abspath(argv.exe)
...
 print()
 
-jobmanager.PrintCaughtWarnings(warning_manager1,
-                               "Warnings during test configuration")
+job_manager.PrintCaughtWarnings(warning_manager1, "Warnings during test configuration")
 print()
 
 with warnings.catch_warnings(record=True) as warning_manager2:
-    retval = jobmanager.RunTests(tests, argv)
+    RETVAL = job_manager.RunTests(tests, argv)
```
andrsd, line 157: Any specific reason for all-caps? I don't believe there is a coding standard for python, but this seems off.

test/src/job_manager.py
```diff
@@ -298,7 +296,7 @@ def ConfigureTests(test_hierarchy: dict, argv):
                 else:
                     print("skipping " + obj.filename)
 
-            # if a specific test has dependencies, also add them to the list of executed tests
+            # Ff a specific test has dependencies, also add them to the list of executed tests
```
andrsd, line 299: Typo: `Ff` -> `If`

test/run_tests
wdhawkins: Ugh... that was a pylint thing that I forgot to undo. Fixed.

test/src/job_manager.py
wdhawkins: Fixed!

ragusa (approved):

zhardy-lanl (approved):

andrsd (approved):

PR #337, opened 2024-07-19
Removing `resources` directory and `configure.sh`

wdhawkins:
This PR removes the `resources` directory and relocates useful content to the `tools` directory. It also removes the `configure.sh` script in favor of the more traditional `cmake` build process.


test/framework/math/spatial_discretization/cg/mesh_2d_tri.lua
```diff
@@ -1,5 +1,5 @@
 --############################################### Setup mesh
 meshgen1 = mesh.FromFileMeshGenerator.Create({
-  filename = "../../../../../resources/TestMeshes/TriangleMesh2x2.obj",
+  filename = "../../../../../tools/mesh/test_meshes/TriangleMesh2x2.obj",
```
andrsd, line 3: I was thinking that stuff related to testing, like these meshes, would go into `tests/assets/`. There could be one more level if needed, so for meshes `tests/assets/meshes`.

`tools` has a different meaning for me. I would not look for test meshes there...

wdhawkins: I like that. I'll move them to `tests/assets/meshes`.

wdhawkins: Fixed.

ragusa (approved):

test/framework/math/spatial_discretization/cg/mesh_2d_tri.lua
zhardy-lanl: @andrsd Why `assets` over something like `data`? Just curious since I've never heard the verbiage for this type of use case before.

andrsd: I think that language originates from web-related development, where you need things like CSS, JS along with the source code. JS and CSS would not be considered data - it is "stuff" you need with the code. And I think this terminology just found its way to other areas of the code development.

andrsd (approved):

zhardy-lanl (approved):

PR #340, opened 2024-07-29
Updating coding standard: use of comments

andrsd:


doc/coding_standard.md
```diff
@@ -130,12 +130,16 @@ call();
 ```
 
 For [doxygen](https://www.doxygen.nl/) style comments `/** */` should be used.
+For single-line doxygen-style comment, use `///`.
```
ragusa, line 133: Do you want to mandate `///` or just sugest it as an acceptable alternative for one-liner ?

ragusa: I let you resolve this whichever is best. I will approve right away.

ragusa (approved):

wdhawkins (approved):

doc/coding_standard.md
zhardy-lanl: I think that at this point in time we should allow either `/** text */` or `/// text` for single line comments just based on the sheer number of `/** text */` comments that exist in the code. If we move to only `///` we would then need to undertake the massive effort of modifying those existing comments.

doc/coding_standard.md
```diff
@@ -129,13 +129,20 @@ In-code comments should use `//`.
 call();
 ```
 
-For [doxygen](https://www.doxygen.nl/) style comments `/** */` should be used.
+For [doxygen](https://www.doxygen.nl/)-style comments, use `///` for single-line comments
```
zhardy-lanl, line 132: If we do not allow `/** */` single line comments, are we going to remove these throughout the codebase?

andrsd: I started working on that. I am almost done from with `framework/`...

zhardy-lanl: Separate PR I take it?

andrsd: Yes

zhardy-lanl (approved):

PR #341, opened 2024-08-06
Adding multi-platform support for properly escaping Lua argument strings in `run_tests`

wdhawkins:


ragusa:
Hi @zhardy-lanl , we had a discussion a few weeks about that triple backslash. Maybe this will fix it but I vaguely remember trying different variants, from triple to no backslashes, on my Mac. I am surprised this works. What do you think?

test/src/test_slot.py
```diff
@@ -38,7 +39,10 @@ def _Run(self):
         for arg in test.args:
             if arg.find("\"") >= 0:
                 qarg = arg.replace('"', '\\"')
-                cmd += "--lua " + "\\\"" + qarg + "\\\""
+                if platform == "darwin":
+                    cmd += "--lua " + "\\\"" + qarg + "\\\""
+                else:
+                    cmd += "--lua " + qarg
```
ragusa, line 45: should it be `platform.system()` and not just `platform` ?

wdhawkins: Hmm. `platform` worked for me. I tested on both OSX and Linux. Let me look at `platform.system()`.

wdhawkins:
> > Hi @zhardy-lanl , we had a discussion a few weeks about that triple backslash. Maybe this will fix it but I vaguely remember trying different variants, from triple to no backslashes, on my Mac. I am surprised this works. What do you think?
> 
> I honestly have no clue how the Lua command line string arguments work. What I have traditionally done to get it to work is something like `[["opt"]]`, which I think Jan may have told me about. I don't think I've seen the triple backslashes.

A lot of the extra quotes and backslashes are escaping things to please Python. It really comes down to the following: The arguments in `tests.json` look like: `k_method=\"pi\"`. Python reads this into an argument string as `k_method="pi"`. When appending the argument to the shell command line, Linux wants `k_method=\"pi\"` and OSX apparently wants `\"k_method=\"pi\"\"`.

zhardy-lanl:
> A lot of the extra quotes and backslashes are escaping things to please Python. It really comes down to the following: The arguments in `tests.json` look like: `k_method=\"pi\"`. Python reads this into an argument string as `k_method="pi"`. When appending the argument to the shell command line, Linux wants `k_method=\"pi\"` and OSX apparently wants `\"k_method=\"pi\"\"`.

That is strange. I wonder why `[[k_method="pi"]]` works on OSX

wdhawkins:
> > A lot of the extra quotes and backslashes are escaping things to please Python. It really comes down to the following: The arguments in `tests.json` look like: `k_method=\"pi\"`. Python reads this into an argument string as `k_method="pi"`. When appending the argument to the shell command line, Linux wants `k_method=\"pi\"` and OSX apparently wants `\"k_method=\"pi\"\"`.
> 
> That is strange. I wonder why `k_method=[["pi"]]` works on OSX

Yes, I also expect that to work. We could change the parameter specification in `test.json` to use that syntax. In this case, the single quotes are escaping the entire string from bash, and the double brackets are an optional Lua syntax for string literals... just a replacement for double quotes.

zhardy-lanl (approved):

andrsd (approved):

test/src/test_slot.py
andrsd: [sys.platform](https://docs.python.org/3/library/sys.html#sys.platform)

[sys.platform.platform()](https://docs.python.org/3/library/sys.html#sys.platform)

:-D

andrsd: and [sys.platform.system()](https://docs.python.org/3/library/platform.html#platform.system)

ragusa: I don't think that all options are identical (case of the answer)

ragusa: add `.lower()`

wdhawkins: After looking through the documentation, I believe what's in the code now is the correct way to go about determining the platform.

PR #347, opened 2024-08-11
Forcing some multi-groupset tests to use only 1 AGS iteration

wdhawkins:
The default behavior of the AGS solver has changed, and we now attempt a pointwise convergence of the groupset-to-groupset scattering. The previous behavior was to only do 1 AGS iteration. For a handful of our regression tests, pointwise AGS convergence is not necessary (doesn't change the answer we are testing for) and significantly increases the run time. For those tests, this PR sets the number of AGS iterations to 1.

### DIFF
--- a/test/modules/linear_boltzmann_solvers/transport_steady/transport_1d_1.lua
+++ b/test/modules/linear_boltzmann_solvers/transport_steady/transport_1d_1.lua
@@ -92,6 +92,7 @@ lbs_options = {
     },
   },
   scattering_order = 5,
+  max_ags_iterations = 1,
 }
 
 phys1 = lbs.DiscreteOrdinatesSolver.Create(lbs_block)
--- a/test/modules/linear_boltzmann_solvers/transport_steady/transport_1d_3a_dsa_ortho.lua
+++ b/test/modules/linear_boltzmann_solvers/transport_steady/transport_1d_3a_dsa_ortho.lua
@@ -94,6 +94,7 @@ lbs_block = {
 
 lbs_options = {
   scattering_order = 1,
+  max_ags_iterations = 1,
 }
 
 phys1 = lbs.DiscreteOrdinatesSolver.Create(lbs_block)
--- a/test/modules/linear_boltzmann_solvers/transport_steady/transport_2d_1_poly.lua
+++ b/test/modules/linear_boltzmann_solvers/transport_steady/transport_2d_1_poly.lua
@@ -97,6 +97,7 @@ lbs_options = {
     },
   },
   scattering_order = 1,
+  max_ags_iterations = 1,
 }
 
 phys1 = lbs.DiscreteOrdinatesSolver.Create(lbs_block)
--- a/test/modules/linear_boltzmann_solvers/transport_steady/transport_2d_2_unstructured.lua
+++ b/test/modules/linear_boltzmann_solvers/transport_steady/transport_2d_2_unstructured.lua
@@ -98,6 +98,7 @@ lbs_options = {
     },
   },
   scattering_order = 1,
+  max_ags_iterations = 1,
 }
 
 phys1 = lbs.DiscreteOrdinatesSolver.Create(lbs_block)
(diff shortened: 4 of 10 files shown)


ragusa:
Reading from your description, I get the feeling that `max_ags_iterations` being greater than 1 increases significantly the solution time and therefore, most of the tests will take longer, except the few tests in this PR for which you are forcing  `max_ags_iterations`  to `1`.
Generally speaking, I find that the iterative techniques, if one wants a very accurate answer, are quite expensive. This was already the case for the multigroup eigenvalue problem. Further, what is required for a highly accurate answer without paying too dearly in CPU times sometimes conflicts with what should be done in massively parallel runs (long pipe fills). Maybe what I am trying to say is that there should be optimum iterative strategies for < 100 ranks, < 1000 ranks, < 5000 ranks, and then above 5000. Sometimes, I feel like I am using iterative parameters "recommended" for large runs but am only using 200 ranks.

wdhawkins:
> Reading from your description, I get the feeling that `max_ags_iterations` being greater than 1 increases significantly the solution time and therefore, most of the tests will take longer, except the few tests in this PR for which you are forcing `max_ags_iterations` to `1`. Generally speaking, I find that the iterative techniques, if one wants a very accurate answer, are quite expensive. This was already the case for the multigroup eigenvalue problem. Further, what is required for a highly accurate answer without paying too dearly in CPU times sometimes conflicts with what should be done in massively parallel runs (long pipe fills). Maybe what I am trying to say is that there should be optimum iterative strategies for < 100 ranks, < 1000 ranks, < 5000 ranks, and then above 5000. Sometimes, I feel like I am using iterative parameters "recommended" for large runs but am only using 200 ranks.

It turns out that most of our tests are single groupset. For these problems, the code automatically sets the maximum number of AGS iterations to 1, and everything is controlled by the WGS options. The compute time for these tests should not change. For the handful of multi-groupset tests that we have, AGS is now converging the solution pointwise. This can be expensive, but may not change the answer we are using for comparison purposes. This is the case for all but one or two problems, so I've set the maximum number of AGS iterations to 1 where appropriate. 

Getting the iterative scheme correct for problems where you may have lagged angular fluxes, reflective boundary conditions, multiple groupsets, upscattering, and inner and outer iterations is part skill, part luck, and part black magic. This might be a good project for an undergrad if we can find someone interested. We need to define some representative problems, learn how to run them well, and write this up as a section in the user manual. Ideally, we can build the necessary smarts into the code to allow it to figure some of this out itself.

zhardy-lanl:
> Ideally, we can build the necessary smarts into the code to allow it to figure some of this out itself.

Tangentially on this topic, I remember talking about how the response evaluator tests did many AGS iterations and the answer changed in #325. Was this changed so that only 1 AGS iteration was done in those problems? We discussed building into the code that for single groupset problems no AGS iterations would be done (either max iterations is 1 or only the WGS solver is called). I am I misremembering something? It doesn't look like there is a diff on the response evaluator tests in #325 or the change in the codebase for single groupset problems.

wdhawkins:
> > Ideally, we can build the necessary smarts into the code to allow it to figure some of this out itself.
> 
> Tangentially on this topic, I remember talking about how the response evaluator tests did many AGS iterations and the answer changed in #325. Was this changed so that only 1 AGS iteration was done in those problems? We discussed building into the code that for single groupset problems no AGS iterations would be done (either max iterations is 1 or only the WGS solver is called). I am I misremembering something? It doesn't look like there is a diff on the response evaluator tests in #325 or the change in the codebase for single groupset problems.

The change forcing 1 AGS iteration for single groupset problems was added as part of #325. See https://github.com/Open-Sn/opensn/blob/5a0275b4c6274d813fbe1975ed8284d07e887108/modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc#L1604.

zhardy-lanl:
> The change forcing 1 AGS iteration for single groupset problems was added as part of #325. See
> 
> https://github.com/Open-Sn/opensn/blob/5a0275b4c6274d813fbe1975ed8284d07e887108/modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc#L1604

Okay great. Thanks for pointing that out!

ragusa (approved):

zhardy-lanl (approved):
For posterity, we should revisit this once we have consistent convergence tolerances

PR #355, opened 2024-08-28
Adding L2-convergence check to the AGS iterative solver

wdhawkins:
This PR changes the default convergence check for the AGS solver from a pointwise check on phi to a check on the L2 norm of the error in phi. This mirrors the behavior of the old PETSc AGS solver and brings some consistency to convergence checking across the WGS and AGS solvers. Using a pointwise convergence check for AGS is optional, and will be used by the Classic Richardson WGS solver since it also uses a pointwise convergence check.

### DIFF
--- a/framework/math/math.cc
+++ b/framework/math/math.cc
@@ -681,4 +681,44 @@ operator-(const std::vector<double>& a, const std::vector<double>& b)
   return result;
 }
 
+double
+ComputeL2Change(std::vector<double>& x, std::vector<double>& y)
+{
+  assert(x.size() == y.size());
+
+  double norm = 0.0;
+  for (auto i = 0; i < x.size(); ++i)
+  {
+    double val = x[i] - y[i];
+    norm += val * val;
+  }
+
+  double global_norm = 0.0;
+  mpi_comm.all_reduce<double>(norm, global_norm, mpi::op::sum<double>());
+
+  return std::sqrt(global_norm);
+}
+
+double
+ComputePointwiseChange(std::vector<double>& x, std::vector<double>& y)
+{
+  assert(x.size() == y.size());
+
+  double pw_change = 0.0;
+  for (auto i = 0; i < x.size(); ++i)
+  {
+    double max = std::max(x[i], y[i]);
+    double delta = std::fabs(x[i] - y[i]);
+    if (max >= std::numeric_limits<double>::min())
+      pw_change = std::max(delta / max, pw_change);
+    else
+      pw_change = std::max(delta, pw_change);
+  }
+
+  double global_pw_change = 0.0;
+  mpi_comm.all_reduce<double>(pw_change, global_pw_change, mpi::op::max<double>());
+
+  return global_pw_change;
+}
+
 } // namespace opensn
--- a/modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
+++ b/modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
@@ -1,7 +1,8 @@
 // SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
 // SPDX-License-Identifier: MIT
 
-#include "modules//linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h"
+#include "modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h"
+#include "modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/convergence.h"
 #include "modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.h"
 #include "framework/math/linear_solver/linear_matrix_action_Ax.h"
 #include "framework/mesh/mesh_continuum/mesh_continuum.h"
@@ -41,7 +42,7 @@ AGSSolver::Solve()
 
     if (lbs_solver_.Options().ags_pointwise_convergence)
     {
-      double pw_change = ComputePointwisePhiChange();
+      double pw_change = ComputePointwisePhiChange(lbs_solver_, phi_old_);
       double rho = (iter == 0) ? 0.0 : sqrt(pw_change / pw_change_prev);
       pw_change_prev = pw_change;
 
@@ -57,7 +58,7 @@ AGSSolver::Solve()
     }
     else
     {
-      double norm = ComputeL2PhiChange();
+      double norm = ComputeL2PhiChange(lbs_solver_, phi_old_);
 
       iter_stats << std::left << std::setw(5) << iter << " Error Norm " << std::left
                  << std::setw(14) << norm;
@@ -92,64 +93,6 @@ AGSSolver::Solve()
   // iteration limit
   if (lbs_solver_.RestartsEnabled() && lbs_solver_.Options().enable_ags_restart_write)
     lbs_solver_.WriteRestartData();
-}
-
-double
-AGSSolver::ComputeL2PhiChange() const
-{
-  double norm = 0.0;
-  auto& phi_new = lbs_solver_.PhiNewLocal();
-  for (int i = 0; i < phi_new.size(); ++i)
-  {
-    double val = phi_new[i] - phi_old_[i];
-    norm += val * val;
-  }
-
-  double global_norm = 0.0;
-  mpi_comm.all_reduce<double>(norm, global_norm, mpi::op::sum<double>());
-  global_norm = std::sqrt(global_norm);
-
-  return global_norm;
-}
-
-double
-AGSSolver::ComputePointwisePhiChange() const
-{
-  auto grid_ptr = GetCurrentMesh();
-  auto& cell_transport_views = lbs_solver_.GetCellTransportViews();
-  auto& groupsets = lbs_solver_.Groupsets();
-  auto& phi_new = lbs_solver_.PhiNewLocal();
-  auto num_moments = lbs_solver_.NumMoments();
-
-  double pw_change = 0.0;
-  for (const auto& cell : grid_ptr->local_cells)
-  {
-    auto& transport_view = cell_transport_views[cell.local_id_];
-    for (int i = 0; i < cell.vertex_ids_.size(); ++i)
-    {
-      for (auto groupset : groupsets)
-      {
-        int gsi = groupset.groups_.front().id_;
-        for (int g = 0; g < groupset.groups_.size(); ++g)
-        {
-          size_t m0g_idx = transport_view.MapDOF(i, 0, gsi + g);
-          double max_phi = std::max(fabs(phi_new[m0g_idx]), fabs(phi_old_[m0g_idx]));
-          for (int m = 0; m < num_moments; ++m)
-          {
-            size_t mng_idx = transport_view.MapDOF(i, m, gsi + g);
-            double delta_phi = std::fabs(phi_new[mng_idx] - phi_old_[mng_idx]);
-            if (max_phi >= std::numeric_limits<double>::min())
-              pw_change = std::max(delta_phi / max_phi, pw_change);
-            else
-              pw_change = std::max(delta_phi, pw_change);
-          } // for g
-        }   // for m
-      }     // for groupset
-    }       // for i
-  }         // for c
-  double global_pw_change = 0.0;
-  mpi_comm.all_reduce<double>(pw_change, global_pw_change, mpi::op::max<double>());
-  return global_pw_change;
 }
 
 } // namespace opensn
--- a/modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
+++ b/modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
@@ -36,16 +36,44 @@ AGSSolver::Solve()
       solver->Solve();
     }
 
-    double pw_change = ComputePointwisePhiChange();
-    double rho = (iter == 0) ? 0.0 : sqrt(pw_change / pw_change_prev);
-    pw_change_prev = pw_change;
+    std::stringstream iter_stats;
+    iter_stats << program_timer.GetTimeString() << " AGS Iteration ";
 
-    if (pw_change < std::max(tolerance_ * (1.0 - rho), 1.0e-10))
-      converged = true;
+    if (lbs_solver_.Options().ags_pointwise_convergence)
+    {
+      double pw_change = ComputePointwisePhiChange();
+      double rho = (iter == 0) ? 0.0 : sqrt(pw_change / pw_change_prev);
+      pw_change_prev = pw_change;
+
+      iter_stats << std::left << std::setw(5) << iter << " Point-wise change " << std::left
+                 << std::setw(14) << pw_change << " Spectral-radius estimate " << std::left
+                 << std::setw(10) << rho;
+
+      if (pw_change < std::max(tolerance_ * (1.0 - rho), 1.0e-10))
+      {
+        converged = true;
+        iter_stats << " CONVERGED";
+      }
+    }
     else
-      phi_old_ = lbs_solver_.PhiNewLocal();
+    {
+      double norm = ComputeL2PhiChange();
 
-    lbs_solver_.QMomentsLocal() = saved_qmoms; // Restore qmoms
+      iter_stats << std::left << std::setw(5) << iter << " Error Norm " << std::left
+                 << std::setw(14) << norm;
+
+      if (norm < tolerance_)
+      {
+        converged = true;
+        iter_stats << " CONVERGED";
+      }
+    }
+
+    if (verbose_)
+      log.Log() << iter_stats.str();
+
+    // Restore qmoms
+    lbs_solver_.QMomentsLocal() = saved_qmoms;
 
     // Write restart data
     if (lbs_solver_.RestartsEnabled() and lbs_solver_.TriggerRestartDump() and
@@ -54,25 +82,34 @@ AGSSolver::Solve()
       lbs_solver_.WriteRestartData();
     }
 
-    if (verbose_)
-    {
-      std::stringstream iter_stats;
-      iter_stats << program_timer.GetTimeString() << " AGS Iteration " << std::left << std::setw(5)
-                 << iter << " Point-wise change " << std::left << std::setw(14) << pw_change
-                 << " Spectral-radius estimate " << std::left << std::setw(10) << rho;
-      if (converged)
-        iter_stats << " CONVERGED";
-      log.Log() << iter_stats.str();
-    }
-
     if (converged)
       break;
+    else
+      phi_old_ = lbs_solver_.PhiNewLocal();
   }
 
   // If restarts are enabled, always write a restart dump upon convergence or when we reach the
   // iteration limit
   if (lbs_solver_.RestartsEnabled() && lbs_solver_.Options().enable_ags_restart_write)
     lbs_solver_.WriteRestartData();
+}
+
+double
+AGSSolver::ComputeL2PhiChange() const
+{
+  double norm = 0.0;
+  auto& phi_new = lbs_solver_.PhiNewLocal();
+  for (int i = 0; i < phi_new.size(); ++i)
+  {
+    double val = phi_new[i] - phi_old_[i];
+    norm += val * val;
+  }
+
+  double global_norm = 0.0;
+  mpi_comm.all_reduce<double>(norm, global_norm, mpi::op::sum<double>());
+  global_norm = std::sqrt(global_norm);
+
+  return global_norm;
 }
 
 double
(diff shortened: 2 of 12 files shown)


modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -75,6 +96,24 @@ AGSSolver::Solve()
...
+double
+AGSSolver::ComputePhiErrorNorm() const
+{
+  double norm = 0.0;
+  auto& phi_new = lbs_solver_.PhiNewLocal();
+  for (int i = 0; i < phi_new.size(); ++i)
+  {
+    double val = phi_new[i] - phi_old_[i];
+    norm += val * val;
+  }
```
ragusa, line 108: we do not have a norm function for vectors ?

ragusa (approved):

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: I'm pretty sure we do, but, doing it this way, eliminated the need to store another temporary copy of phi.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
```diff
@@ -463,6 +463,8 @@ LBSSolver::OptionsBlock()
   params.AddOptionalParameter(
     "max_ags_iterations", 100, "Maximum number of across-groupset iterations.");
   params.AddOptionalParameter("ags_tolerance", 1.0e-6, "Across-groupset iterations tolerance.");
+  params.AddOptionalParameter(
+    "ags_pointwise_convergence", false, "Flag that enables pointwise convergence checking.");
```
zhardy-lanl, line 467: It may be better to have this be a string constrained to "ptwise" and "l2" (with default "l2" in this case) to make it clear what the convergence check is when specifying it in the input file.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
```diff
@@ -48,6 +48,8 @@ class AGSSolver
   double tolerance_;
   bool verbose_;
 
+  double ComputePhiErrorNorm() const;
```
zhardy-lanl, line 51: Consider changing this to `ComputePhiL2Change` or something like that if you are sticking to using this routine instead of the existing routines.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -36,13 +36,45 @@ AGSSolver::Solve()
...
+      {
+        std::stringstream iter_stats;
+        iter_stats << program_timer.GetTimeString() << " AGS Iteration " << std::left
+                   << std::setw(5) << iter << " Point-wise change " << std::left << std::setw(14)
+                   << pw_change << " Spectral-radius estimate " << std::left << std::setw(10)
+                   << rho;
+        if (converged)
+          iter_stats << " CONVERGED";
+        log.Log() << iter_stats.str();
+      }
```
zhardy-lanl, line 58: Consider unifying the printers and using a check to write point-wise or l2 instead of duplicating the whole block.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -54,17 +86,6 @@ AGSSolver::Solve()
...
-    if (verbose_)
-    {
-      std::stringstream iter_stats;
-      iter_stats << program_timer.GetTimeString() << " AGS Iteration " << std::left << std::setw(5)
-                 << iter << " Point-wise change " << std::left << std::setw(14) << pw_change
-                 << " Spectral-radius estimate " << std::left << std::setw(10) << rho;
-      if (converged)
-        iter_stats << " CONVERGED";
-      log.Log() << iter_stats.str();
-    }
```
zhardy-lanl, line 66: Use an if check for point wise change here instead. See comment above.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -75,6 +96,24 @@ AGSSolver::Solve()
...
+
+  double global_norm = 0.0;
+  mpi_comm.all_reduce<double>(norm, global_norm, mpi::op::sum<double>());
+  global_norm = std::sqrt(global_norm);
+
+  return global_norm;
+}
+
 double
 AGSSolver::ComputePointwisePhiChange() const
```
zhardy-lanl, line 118: Isn't this implemented in `LBSSolver` in your Classic Richardson PR? Will there be duplicate routines when both PRs are in? Is that okay? Where should this routine live?

out
zhardy-lanl, line 1: What is the purpose of this file here?

zhardy-lanl (changes requested):

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: I'll see if I can clean this up a bit.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: I think so. This is a tiny bit of code, and I think we should be memory conscious - even at the cost of duplicating a bit of code here and there.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: Not quite. This isn't a test between lbs_solver::phi_new_local and lbs_solver::phi_old_local.

test/modules/linear_boltzmann_solvers/transport_steady/tests.json
```diff
@@ -647,7 +647,7 @@
       {
         "type": "KeyValuePair",
         "key": "[0]   Out-flow rate               =",
-        "goldvalue": 1.343586e+01,
+        "goldvalue": 1.343510e+01,
```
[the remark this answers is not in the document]
wdhawkins: Yes. L2 is now the default.

out
wdhawkins: Accidentally committed. I've removed it.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
zhardy-lanl: What about writing a specific L2-error routing in math.h/cc that takes two vectors? That way we don't have to write a new L2-error routine every time we need to access a different solution vector in a class.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
ragusa: semantics: shouldn't this be ComputeL2ChangePhi? just like it the previous one should be ComputeErrorNormPhi? just wondering.

deeper question to @zhardy-lanl : why this  note" if you are sticking to using this routine instead of the existing routines"

zhardy-lanl: Probably `ComputeL2PhiChange` is more consistent IMO. On the deeper question, see my comments on having multiple routines floating around to compute L2 errors/changes. I said that in the event that the routine was abandoned in favor of existing routines to compute L2 norms.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
zhardy-lanl: Okay I am confused by a number of things now... In another comment you said we should be memory conscious and avoid making a copy of phi, but the AGS solver now stores a copy of phi. Where is it appropriate to make copies then? As far as I can tell, the difference between this routine and the one that is introduced in the Classic Richardson PR is that the Classic Richardson version in LBSSolver operates on group sets. I assume you by numerous global synchronizations you mean having to do that for every groupset, but couldn't that routine just be rewritten to handle a single groupset or all groupsets by doing something like passing the groupset ID and if it is -1 the operation is done on all groupsets?

wdhawkins: > Okay I am confused by a number of things now... In another comment you said we should be memory conscious and avoid making a copy of phi, but the AGS solver now stores a copy of phi. Where is it appropriate to make copies then? As far as I can tell, the difference between this routine and the one that is introduced in the Classic Richardson PR is that the Classic Richardson version in LBSSolver operates on group sets. I assume you by numerous global synchronizations you mean having to do that for every groupset, but couldn't that routine just be rewritten to handle a single groupset or all groupsets by doing something like passing the groupset ID and if it is -1 the operation is done on all groupsets?

The AGS solver has to have it's own `phi_old` that spans all groupsets. I don't know any other way to make it work. This was previously a PETSc vector (x_old) and is now an `std::vector` member of the AGS solver. In fact, the AGS solver is quite a bit leaner memory-wise since we previously had both `x` (phi) and `x_old` (phi_old) PETSc vectors (at least... I can't remember if there were others). In this case, the `phi_old` vector is necessary to implement the algorithm correctly. In the case of computing the L2 error, I don't need to make another copy of phi to compute that quantity and, IMHO, memory is at a premium, so it's better to duplicate 10 lines of code than to waste memory to use an existing routine.  I do agree that it's better to move the L2 routine that takes two vectors out of `ags_solver` and into math.h/math.cc. I didn't do this initially since I didn't expect it to get much use, but that is the better solution.

On the topic of Classic Richardson and pointwise computations, I could add functionality to the pointwise routine to handle all groupsets if you pass -1 as the id.  This seems a bit clunky to me, and would probably require duplicating a huge chunk of that routine to prevent having `if` statements in code that needs to be lean and fast and/or to avoid lots of global syncs. I'll look at it and see what can be done.

zhardy-lanl: `phi_new_local_` and `phi_old_local_` in `LBSSolver` span all groupsets. In the classic Richardson PR implementation this is why it is indexed `gsi + g`. You are already pulling in `lbs_solver_.PhiNewLocal()`, why can you not use `lbs_solver_.PhiOldLocal()`?

On the pointwise routine, you are already duplicating the code, it just depends on where you do it. Alternatively, you could have an override that does not take a groupset in `LBSSolver`. A second option would be to loop over groupsets at the outermost level and do the check there. A third option would be to create a vector of groups outside at the top of the function based on the function input. There is already an if statement in the innermost portion of the code.

zhardy-lanl: I fully understand the argument of allowing for some duplicate code to allow for better performance. The issues I am trying to raise here are that 1) we should try to generalize where we can first, and 2) if we do duplicate code, we should think carefully about where it lives.

In this case, the differences between the two routines are 1) access to a member variable, 2) groupset vs. all groups, and 3) living in `LBSSolver` which does not have iterative schemes implemented vs. living in something in `iterative_methods`. On 1), we should not duplicate code when using a function argument will accomplish the same task. On 2) and 3) If the performance hit of an if statement in the outermost loop or defining group bounds at the top of the function is too large and a duplicate routine is necessary, we should think more carefully about where each routine lives. IMHO, both should live in LBSSolver or both should live in a class defined within `iterative_methods`.

wdhawkins: > `phi_new_local_` and `phi_old_local_` in `LBSSolver` span all groupsets. In the classic Richardson PR implementation this is why it is indexed `gsi + g`. You are already pulling in `lbs_solver_.PhiNewLocal()`, why can you not use `lbs_solver_.PhiOldLocal()`?

Yes, I understand that `phi_new_local` and `phi_old_local` span all groupsets. But, using `phi_old_local_` won't work. It does not contain the value of phi from the previous AGS iteration.

> On the pointwise routine, you are already duplicating the code, it just depends on where you do it. Alternatively, to not duplicate code, you could have an override that does not take a groupset in `LBSSolver`. A second option would be to loop over groupsets at the outermost level and do the check there. A third option would be to create a vector of groups outside at the top of the function based on the function input.

Yes, I think I like option 3 the best. I'll take a look... like I said, I'm sure it can be fixed, I'm just not a fan of passing -1 as a groupset id.

wdhawkins: >On 1), we should not duplicate code when using a function argument will accomplish the same task. 

I agree, and I'll move this routine into `math.h`/`math.cc`.

>On 2) and 3) If the performance hit of an if statement in the outermost loop or defining group bounds at the top of >the function is too large and a duplicate routine is necessary, we should think more carefully about where each >routine lives. IMHO, both should live in `LBSSolver` or both should live in an appropriate class defined within >`iterative_methods`.

Yes. As I mentioned in a comment in my Classic Richardson PR, I already have a new PR ready to go that pulls out all of the io functionality of `LBSSolver` into separate routines. It would be nice to do something similar with things like convergence checking... and hopefully it works out to be as clean as the io stuff did. But, I am actively trying not to add any additional code to `LBSSolver` unless it absolutely needs to live there.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: After thinking about this, I don't see how it's possible to do this by invoking a serial L2-error routine on each location. Computing the 2-norm of each individual vector and adding is incorrect. The only correct way to do this is to do a global reduce on the sum of the squares of the error and then have each location do a sqrt.

wdhawkins: We could create a routine in `lbs_solver` that does this type of convergence checking, but I think it might be best to keep it local to `ags_solver` for now. As I have mentioned, I hope to pull these routines out of `lbs_solver` at some point, and we could refactor this functionality at that time.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.h
wdhawkins: Changed to `ComputeL2PhiChange`.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: Done.

modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.cc
wdhawkins: Done.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
zhardy-lanl: > After thinking about this, I don't see how it's possible to do this by invoking a serial L2-error routine on each location. Computing the 2-norm of each individual vector and adding is incorrect. The only correct way to do this is to do a global reduce on the sum of the squares of the error and then have each location do a sqrt.

Why couldn't that be done in a function in math.h/.cc?

> We could create a routine in lbs_solver that does this type of convergence checking, but I think it might be best to keep it local to ags_solver for now. As I have mentioned, I hope to pull these routines out of lbs_solver at some point, and we could refactor this functionality at that time.

I am fine punting things, I suppose... But we should try to minimize introducing code that we admit needs a refactor before hitting master.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
zhardy-lanl: > It does not contain the value of phi from the previous AGS iteration.

Ah, this is a good point I did not think about. Each WGS solver updates the respective portion of the phi vectors. Once we get around to addressing where solutions should live, these types of things should be much more clear.

> Yes, I think I like option 3 the best. I'll take a look... like I said, I'm sure it can be fixed, I'm just not a fan of passing -1 as a groupset id.

I'm not a huge fan either. If you look at `ComputeLeakage` in `DiscreteOrdinatesSolver` you could do something similar. Pass a vector of groupset IDs. If it is empty, you compute for all groupsets, if it is not, you use the groupset IDs from the argument.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
```diff
@@ -75,6 +94,24 @@ AGSSolver::Solve()
...
+  {
+    double val = phi_new[i] - phi_old_[i];
+    norm += val * val;
+  }
+
+  double global_norm = 0.0;
+  mpi_comm.all_reduce<double>(norm, global_norm, mpi::op::sum<double>());
+  global_norm = std::sqrt(global_norm);
+
+  return global_norm;
```
zhardy-lanl, line 112:
```suggestion
  return std::sqrt(global_norm);
```
Just a suggestion

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: Fixed. I've added L2 and pointwise checks to `math.h/.cc`. These routines operate on `std::vector<double>`. I've also added analogous checks to `lbs_solver/iterative_algorithms/convergence.h/.cc` that operate on LBS-structured data.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/ags_solver.cc
wdhawkins: I believe I've resolved this issue. I've moved the convergence checks out of LBSSolver and into separate routines in `lbs_solver/iterative_algorithms/convergence.h/.cc`.

andrsd (changes requested):

framework/math/math.cc
```diff
@@ -681,4 +681,44 @@ operator-(const std::vector<double>& a, const std::vector<double>& b)
   return result;
 }
 
+double
+ComputeL2Change(std::vector<double>& x, std::vector<double>& y)
+{
+  assert(x.size() == y.size());
+
+  double norm = 0.0;
+  for (int i = 0; i < x.size(); ++i)
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/convergence.cc
```diff
@@ -0,0 +1,104 @@
...
+
+  auto grid_ptr = GetCurrentMesh();
+  auto& cell_transport_views = lbs_solver.GetCellTransportViews();
+  auto num_moments = lbs_solver.NumMoments();
+
+  double pw_change = 0.0;
+  for (const auto& cell : grid_ptr->local_cells)
+  {
+    auto& transport_view = cell_transport_views[cell.local_id_];
+    for (int i = 0; i < cell.vertex_ids_.size(); ++i)
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/convergence.cc
```diff
@@ -0,0 +1,104 @@
...
+    {
+      for (auto id : groupset_ids)
+      {
+        auto& groupset = lbs_solver.Groupsets()[id];
+        int gsi = groupset.groups_.front().id_;
+        for (int g = 0; g < groupset.groups_.size(); ++g)
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/iterative_methods/convergence.cc
```diff
@@ -0,0 +1,104 @@
...
+double
+ComputeL2PhiChange(LBSSolver& lbs_solver,
+                   std::optional<const std::reference_wrapper<std::vector<double>>> opt_phi_old)
+{
+  double norm = 0.0;
+  auto& phi_new = lbs_solver.PhiNewLocal();
+  std::vector<double>& phi_old =
+    opt_phi_old.has_value() ? opt_phi_old.value().get() : lbs_solver.PhiOldLocal();
+
+  for (int i = 0; i < phi_new.size(); ++i)
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

andrsd (approved):

zhardy-lanl (approved):

PR #357, opened 2024-09-03
Moving file i/o routines out of LBSSolver

wdhawkins:
This PR moves angular flux and moment i/o routines out of the LBSSolver class and into stand-alone functions.


ragusa (approved):
later, will we want to convert this to h5 ??

wdhawkins:
> later, will we want to convert this to h5 ??

Yes, absolutely. No more custom binary formats.

andrsd (approved):

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -0,0 +1,248 @@
...
+WriteFluxMoments(LBSSolver& lbs_solver,
+                 const std::string& file_base,
+                 std::optional<const std::reference_wrapper<std::vector<double>>> opt_src)
+{
+  // Open file
+  std::string file_name = file_base + std::to_string(opensn::mpi_comm.rank()) + ".data";
+  std::ofstream file(file_name, std::ofstream::binary | std::ofstream::out | std::ofstream::trunc);
+  OpenSnLogicalErrorIf(not file.is_open(), "Failed to open " + file_name + ".");
+
+  std::vector<double>& src = opt_src.has_value() ? opt_src.value().get() : lbs_solver.PhiOldLocal();
```
zhardy-lanl, line 21: Two quick questions: 1) why use `PhiOldLocal` instead of `PhiNewLocal`, and 2) why use this `std::optional<std::reference_wrapper<std::vector<double>>> opt_dest` stuff instead of just requiring a vector? On 2), this eliminates hard-coding `PhiOldLocal`.

zhardy-lanl (approved):
This is a great start to clean up LBSSolver. Just throwing the idea out there to steal some of David's mesh I/O implementation and encapsulate this into a class. That can be done here or we can address it later.

This `std::optional<std::reference_wrapper<std::vector<double>>> opt_dest` stuff is interesting. I haven't seen it before.

Marking as comment for now. I am not requesting changes, but would like a very small amount of clarification on a comment.

wdhawkins:
> This is a great start to clean up LBSSolver. Just throwing the idea out there to steal some of David's mesh I/O implementation and encapsulate this into a class. That can be done here or we can address it later.
> 
> This `std::optional<std::reference_wrapper<std::vector<double>>> opt_dest` stuff is interesting. I haven't seen it before.
> 
> Marking as comment for now. I am not requesting changes, but would like a very small amount of clarification on a comment.

Unlike the mesh i/o routines, these routines don't have much in common, so I opted not to encapsulate them in a class. Certainly, we could make the argument that the angular flux methods could be moved into a class, but, given their simplicity, I'd argue for leaving them as stand-alone functions.

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
wdhawkins: I chose `PhiOldLocal` solely because that was the default here: `lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_io_flux_moments.cc`.  I'm using `std::optional<std::reference_wrapper<std::vector<double>>> opt_dest` so that it's possible to optionally pass a reference to a vector.

zhardy-lanl: That's a fair point. Should we use `PhiNewLocal()` as the default? Is there any rationale to using `PhiOldLocal()`?

zhardy-lanl:
> Unlike the mesh i/o routines, these routines don't have much in common, so I opted not to encapsulate them in a class. Certainly, we could make the argument that the angular flux methods could be moved into a class, but, given their simplicity, I'd argue for leaving them as stand-alone functions.

What in particular are you referring that the mesh i/o routines have in common that the flux moment i/o routines do not? The mesh i/o routines are just readers and writers into particular formats. There is no underlying attributes, and all of the routines are static. It is just a collection of methods. How is this so different?

zhardy-lanl (approved):
Thank you for the changes!

PR #359, opened 2024-09-09
Refactoring NDArray for performance

wdhawkins:
The version of NDArray that is currently in OpenSn was changed considerably from the original version, and several design elements that were critical to good performance were lost. This PR attempts to recover some of that performance loss.

### DIFF
--- a/framework/data_types/ndarray.h
+++ b/framework/data_types/ndarray.h
@@ -6,10 +6,13 @@
 #include <cstddef>
 #include <type_traits>
 #include <cassert>
-#include <vector>
 #include <array>
+#include <memory>
+#include <initializer_list>
 #include <stdexcept>
-#include <string>
+#include <algorithm>
+#include <numeric>
+#include <vector>
 
 namespace opensn
 {
@@ -18,24 +21,18 @@ template <typename T>
 class NDArray
 {
 private:
-  size_t rank_;
-  std::vector<size_t> dimensions_;
-  std::vector<size_t> strides_;
-  size_t size_;
-  T* base_;
-
-  template <bool...>
-  struct bool_pack
-  {
-  };
-
-  template <class... U>
-  using conjunction = std::is_same<bool_pack<true, U::value...>, bool_pack<U::value..., true>>;
-
-  template <typename... U>
-  using AllIntegral = typename conjunction<std::is_integral<U>...>::type;
+  static constexpr size_t max_rank = 10;
 
 public:
+  /**
+   * Creates an empty array.
+   * \throw std::bad_alloc if memory allocation fails.
+   *
+   * This constructor creates an empty array and initializes the reference
+   * count to one.
+   */
+  NDArray() noexcept : rank_(0), size_(0), storage_(nullptr), dimensions_{}, strides_{} {}
+
   /**
    * Creates an array with the specified number of elements in each dimension,
    * from a vector-list.
@@ -48,29 +45,27 @@ public:
    */
   template <typename D>
   explicit NDArray(const std::vector<D>& dims)
-    : rank_(dims.size()), dimensions_(rank_), strides_(rank_), size_(0), base_(nullptr)
   {
-    static_assert(std::is_integral<D>::value,
-                  "NDArray dims argument must have vector of integral types.");
+    SetDimensions(dims);
+    Initialize();
+  }
 
-    // Populate dimensions
-    for (size_t i = 0; i < rank_; ++i)
-      dimensions_[i] = dims[i];
-
-    // Populate strides and size
-    size_ = 1;
-    for (size_t i = 0; i < rank_; ++i)
-    {
-      assert(dimensions_[i] > 0);
-      size_ *= dimensions_[i];
-      strides_[i] = 1;
-      for (size_t j = i + 1; j < rank_; ++j)
-        strides_[i] *= dimensions_[j];
-    }
-
-    base_ = new T[size_];
-    for (size_t i = 0; i < size_; ++i)
-      base_[i] = 0.0;
+  /**
+   * Creates an array with the specified number of elements in each dimension,
+   * from a vector-list and initializes the array.
+   *
+   * \param dims `std::vector` list of the number of elements in each
+   *             dimension.
+   * \param value Initial element value.
+   * \throw std::bad_alloc if memory allocation fails.
+   *
+   * This constructor creates an array with the specified size.
+   */
+  template <typename D>
+  explicit NDArray(const std::vector<D>& dims, T value)
+  {
+    SetDimensions(dims);
+    ValueInitialize(value);
   }
 
   /**
@@ -85,28 +80,27 @@ public:
    */
   template <typename D, size_t N>
   explicit NDArray(const std::array<D, N>& dims)
-    : rank_(N), dimensions_(N), strides_(N), size_(0), base_(nullptr)
   {
-    static_assert(std::is_integral<D>::value,
-                  "NDArray dims argument must have array of integral types.");
-    // Populate dimensions
-    for (size_t i = 0; i < N; ++i)
-      dimensions_[i] = dims[i];
+    SetDimensions(dims);
+    Initialize();
+  }
 
-    // Populate strides and size
-    size_ = 1;
-    for (size_t i = 0; i < N; ++i)
-    {
-      assert(dimensions_[i] > 0);
-      size_ *= dimensions_[i];
-      strides_[i] = 1;
-      for (size_t j = i + 1; j < N; ++j)
-        strides_[i] *= dimensions_[j];
-    }
-
-    base_ = new T[size_];
-    for (size_t i = 0; i < size_; ++i)
-      base_[i] = 0.0;
+  /**
+   * Creates an array with the specified number of elements in each dimension,
+   * from an array and initializes the array.
+   *
+   * \param dims `std::array` list of the number of elements in each
+   *             dimension.
+   * \param value Initial element value.
+   * \throw std::bad_alloc if memory allocation fails.
+   *
+   * This constructor creates an array with the specified size.
+   */
+  template <typename D, size_t N>
+  explicit NDArray(const std::array<D, N>& dims, T value)
+  {
+    SetDimensions(dims);
+    ValueInitialize(value);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ragusa (approved):

andrsd (approved):

framework/data_types/ndarray.h
```diff
@@ -19,23 +22,22 @@ class NDArray
...
-  {
-  };
-
-  template <class... U>
-  using conjunction = std::is_same<bool_pack<true, U::value...>, bool_pack<U::value..., true>>;
-
-  template <typename... U>
-  using AllIntegral = typename conjunction<std::is_integral<U>...>::type;
+  std::unique_ptr<T[]> storage_;
+  static constexpr size_t max_rank = 10;
```
zhardy-lanl, line 27: I just looked back at the coding standards and I could have sworn we had added something about class declaration structure along the lines to have the structure of routines -> attributes -> static routines -> static attributes with each having public/protected/private. I was going to ask if the declaration could be organized in this way, but I don't see anything on this in the coding standards... Does anyone else remember this discussion from long ago?

framework/data_types/ndarray.h
```diff
@@ -48,341 +50,147 @@ class NDArray
...
    *             dimension.
+   * \param value Initial element value.
    * \throw std::bad_alloc if memory allocation fails.
    *
    * This constructor creates an array with the specified size.
    */
-  template <typename D, size_t N>
-  explicit NDArray(const std::array<D, N>& dims)
-    : rank_(N), dimensions_(N), strides_(N), size_(0), base_(nullptr)
+  template <typename D>
```
zhardy-lanl, line 69: Is there a real purpose to templating essentially on different `int` types here?

framework/data_types/ndarray.h
```diff
@@ -48,341 +50,147 @@ class NDArray
...
-    std::swap(rank_, other.rank_);
-    std::swap(dimensions_, other.dimensions_);
-    std::swap(strides_, other.strides_);
-    std::swap(size_, other.size_);
-    std::swap(base_, other.base_);
+    if (this != &other)
+    {
+      NDArray<T> temp(other); // Copy constructor, uses swap to reduce redundancy
+      this->swap(temp);
+    }
```
zhardy-lanl, line 176:
```suggestion
    if (this != &other)
    {
      rank_ = other.rank_;
      size_ = other.size_;
      dimensions_ = other.dimensions_;
      strides_ = other.strides_;
      storage_ = other.storage_
      std::copy(other.storage_.get(), other.storage_.get() + size_, storage_.get());
    }
```
Why allocate a new NDArray here instead of implementing this more like the move assignment operator?

framework/data_types/ndarray.h
```diff
@@ -446,94 +242,104 @@ class NDArray
...
-   * Accesses the specified element.
+   * Accesses the specified element for an array with N dimensions.
    *
-   * \param args The indices of the desired element.
+   * \param args Indices for each dimension.
    * \return Read/write reference to the element.
    */
   template <typename... Args>
-  T& operator()(Args... args) noexcept
+  inline __attribute__((always_inline)) constexpr T& operator()(Args... args) noexcept
```
zhardy-lanl, line 257: This is some notation I haven't seen before... Could you explain to a layman what this means? In particular, I've never seen the `__attribute__((always_inline))` and don't fully understand why this should be marked `constexpr`

framework/data_types/ndarray.h
```diff
@@ -446,94 +242,104 @@ class NDArray
...
-    T* address = base_ + indices[N - 1];
-    for (size_t i = 0; i < N - 1; ++i)
-      address += strides_[i] * indices[i];
-    return *(address);
+    for (size_t i = 0; i < rank_; ++i)
+    {
+      if (indices[i] >= dimensions_[i])
+        throw std::out_of_range("Index out of bounds.");
+    }
+    return (*this)(args...);
```
zhardy-lanl, line 277:
```suggestion
    return storage[compute_index(args...)];
```
Just a suggestion that you can ignore. Never been a big fan of calling `()` in this way within a class.

framework/data_types/ndarray.h
```diff
@@ -446,94 +242,104 @@ class NDArray
...
 
-    T* address = base_ + indices[N - 1];
-    for (size_t i = 0; i < N - 1; ++i)
-      address += strides_[i] * indices[i];
-    return *(address);
+  /// Returns an iterator pointing to the beginning of the array.
+  inline constexpr T* begin() const noexcept { return storage_.get(); }
+
+  /// Returns a constant iterator pointing to the beginning of the array.
+  inline constexpr const T* cbegin() const noexcept { return storage_.get(); }
```
zhardy-lanl, line 314: Is it better to override `begin`/`end` in this case or use `cbegin`/`cend`?

framework/data_types/ndarray.h
```diff
@@ -446,94 +242,104 @@ class NDArray
...
+  inline constexpr bool empty() const noexcept { return size_ == 0; }
+
+  /// Returns a pointer to the underlying array data.
+  inline constexpr T* data() const noexcept { return storage_.get(); }
+
+  /// Returns the rank of the array.
+  inline constexpr size_t rank() const noexcept { return rank_; }
+
+  /// Returns the dimension of the array.
+  inline std::vector<size_t> dimension() const noexcept
```
zhardy-lanl, line 335: With `size_t` used here, should we have the template parameter `D` in routines above or just use `size_t`?

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
+    std::swap(rank_, other.rank_);
+    std::swap(size_, other.size_);
+    std::swap(storage_, other.storage_);
+    std::swap(dimensions_, other.dimensions_);
+    std::swap(strides_, other.strides_);
+  }
+
+private:
+  template <typename D>
+  void set_dimensions(const std::vector<D>& dims)
```
zhardy-lanl, line 391:
```suggestion
  void SetDimensions(const std::vector<D>& dims)
```

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
+  {
+    rank_ = dims.size();
+    if (rank_ > max_rank)
+      throw std::invalid_argument("Rank exceeds the maximum allowed value of 10.");
+    std::copy(dims.begin(), dims.end(), dimensions_.begin());
+  }
+
+  template <typename D, size_t N>
+  void set_dimensions(const std::array<D, N>& dims)
```
zhardy-lanl, line 400:
```suggestion
  void SetDimensions(const std::array<D, N>& dims)
```

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
+  {
+    rank_ = N;
+    static_assert(N <= max_rank, "Rank exceeds the maximum allowed value of 10.");
+    std::copy(dims.begin(), dims.end(), dimensions_.begin());
+  }
+
+  template <typename D>
+  void set_dimensions(const std::initializer_list<D>& dims)
```
zhardy-lanl, line 408:
```suggestion
  void SetDimensions(const std::initializer_list<D>& dims)
```

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
+  {
+    rank_ = dims.size();
+    if (rank_ > max_rank)
+      throw std::invalid_argument("Rank exceeds the maximum allowed value of 10.");
+    std::copy(dims.begin(), dims.end(), dimensions_.begin());
+  }
+
+  void initialize()
```
zhardy-lanl, line 416:
```suggestion
  void Initialize()
```

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
+    {
+      size_ *= dimensions_[i];
+      if (i > 0)
+        strides_[i - 1] = strides_[i] * dimensions_[i];
+    }
+
+    storage_ = std::make_unique<T[]>(size_);
+  }
+
+  void value_initialize(T value)
```
zhardy-lanl, line 430:
```suggestion
  void Initialize(T value)
```

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
+  {
+    initialize();
+    std::fill_n(storage_.get(), size_, value);
+  }
+
+  template <typename... Args>
+  inline __attribute__((always_inline)) constexpr size_t compute_index(Args... args) const noexcept
```
zhardy-lanl, line 437:
```suggestion
  inline __attribute__((always_inline)) constexpr size_t ComputeIndex(Args... args) const noexcept
```

framework/data_types/ndarray.h
```diff
@@ -545,37 +351,114 @@ class NDArray
...
                                 std::to_string(dimensions_[i]));
+      }
+    }
 
-    size_t index = indices[N - 1];
-    for (size_t i = 0; i < N - 1; ++i)
-      index += strides_[i] * indices[i];
-    return index;
+    return compute_index(args...);
   }
```
zhardy-lanl, line 373: It seems like this routine can be removed.

zhardy-lanl (changes requested):

framework/data_types/ndarray.h
wdhawkins: It's much faster if we swap than doing the `std::copy`.

framework/data_types/ndarray.h
wdhawkins: `inline` by itself is just a hint to the compiler. I'm using a gcc/clang specific compiler directive (`__attribute__((always_inline))`) to tell the compiler I really want you to inline this. It's a `constexpr` to try to get as much performance out of it as possible. It also calls a `constexpr` method.

framework/data_types/ndarray.h
wdhawkins: I think it's better to use `cbegin` and `cend`. This follows the convention set by the standard library.

framework/data_types/ndarray.h
andrsd: IIRC when I was playing with this class, this allows you to initialize  the `NDArray` with code like this: `NDArray<Real> array({5, 10});`. Here the dimensions are treated as `int`s. We don't always have the dimensions in the `size_t` type, so this way the compiler will do the right conversion for us.

framework/data_types/ndarray.h
andrsd: I remember the discussion and I do not see it in the coding standards either. Also, no issue capturing this either...

andrsd:
@wdhawkins Is the reason for `NDArray` using "column-major" format so that the data can be passed directly into libraries like lapack, magma, plasma, etc.?  The reason I am asking is that before #356, the matrix storage format was really row-major and the matrix operations were coded with that in mind. Many of those operations can be re-written to be column-major friendly. That should also help to recover the lost performance we observe in #356.

wdhawkins:
> @wdhawkins Is the reason for `NDArray` using "column-major" format so that the data can be passed directly into libraries like lapack, magma, plasma, etc.? The reason I am asking is that before #356, the matrix storage format was really row-major and the matrix operations were coded with that in mind. Many of those operations can be re-written to be column-major friendly. That should also help to recover the lost performance we observe in #356.

Unless I really goofed something, it should be using row-major order. I may add a routine that exports the underlying data as column-major for the sole purpose of interface with things like BLAS that typically expect that format.

andrsd:
Ok. I did a print out of the data and it is row-major. I misread the code that computes `strides_` 🤦 .
I guess I need to go look at the new operations introduced in #356. Maybe just too many mapping function calls. Maybe, I could try to have things like `GaussElimination` as class methods and then I could exploit the underlying data layout... 💡

framework/data_types/ndarray.h
wdhawkins: Fixed.

framework/data_types/ndarray.h
wdhawkins: I do... just for consistency purposes. Probably not a high priority though.

framework/data_types/ndarray.h
wdhawkins: I think just size_t. For this purpose, I don't think it's important to template.

framework/data_types/ndarray.h
wdhawkins: Fixed.

framework/data_types/ndarray.h
wdhawkins: Fixed.

framework/data_types/ndarray.h
wdhawkins: Fixed.

framework/data_types/ndarray.h
wdhawkins: Let me look at this. You're right that we call the copy constructor anyway, so, from a performance perspective, it's probably the same regardless.

andrsd: I think if you do just `storage_ = other.storage_;`, then you leak the memory associated with `storage_`. This is why I think the original code uses `swap`. The old pointer gets assigned into `other` and when it goes out of scope, the memory gets deallocated...

andrsd: But, this is not true anymore, since `storage_` is now `unique_ptr`... Just an explanation why `swap` was there...

framework/data_types/ndarray.h
wdhawkins: I've made changes to restructure the code following this coding standard.

framework/data_types/ndarray.h
wdhawkins: Yeah, I suspect that the `swap` predates my use of `unique_ptr`.

zhardy-lanl (approved):

PR #377, opened 2024-10-03
Fixing conditions that decide `FaceOrientation`

### LINKED ISSUE #367: Incorrectly setting incoming/parallel orientation of a face in SPDS
The code in `SPDS::PopulateCellRelationships` does this:
```
constexpr double tolerance = 1.0e-16;
...
FaceOrientation orientation = FaceOrientation::PARALLEL;
...
if (mu > tolerance)
    orientation = FaceOrientation::OUTGOING;
else if (mu < tolerance)
    orientation = FaceOrientation::INCOMING;
```

Which means parallel orientation is set only if `mu == 1.0e-16` which will "never" happen. It seems like the else-if branch should be:
```
else if (mu < -tolerance)
```
so that `PARALLEL` is from `-tolerance .. tolerance`.

Can @ragusa, @wdhawkins, or @zhardy-lanl confirm that this is a real bug?

andrsd:
Closes #367

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/spds/spds.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/spds/spds.cc
@@ -91,7 +91,7 @@ SPDS::PopulateCellRelationships(const Vector3& omega,
       {
         if (mu > tolerance)
           orientation = FOOUTGOING;
-        else if (mu < tolerance)
+        else if (mu < -tolerance)
           orientation = FOINCOMING;
 
         cell_face_orientations_[cell.local_id][f] = orientation;
@@ -127,7 +127,7 @@ SPDS::PopulateCellRelationships(const Vector3& omega,
         const double adj_mu = omega.Dot(adj_face.normal);
         if (adj_mu > tolerance)
           orientation = FOOUTGOING;
-        else if (adj_mu < tolerance)
+        else if (adj_mu < -tolerance)
           orientation = FOINCOMING;
 
         switch (orientation)
@@ -173,7 +173,7 @@ SPDS::PopulateCellRelationships(const Vector3& omega,
         }
       }
       // If not outgoing determine what it is dependent on
-      else
+      else if (cell_face_orientations_[cell.local_id][f] == FOINCOMING)
       {
         // if it is a cell and not bndry
         if (face.has_neighbor and not face.IsNeighborLocal(grid_))


ragusa:
@andrsd @zachhardy @wdhawkins : can you look at this part of the code, lines 175 and up, also below? Now that we are allowing orientation to be  PARALLEL with this PR, the following lines will be visiting when the orientation is not OUTGOING because of the `else `statement. So now, the else statement will be active when a face orientation is **either** INCOMING or PARALLEL (before, PARALLEL was **never** active). I am worried that those lines of codes should not happen when the face is PARALLEL. 
```
      // If not outgoing determine what it is dependent on
      else
      {
        // if it is a cell and not bndry
        if (face.has_neighbor_ and not face.IsNeighborLocal(grid_))
          location_dependencies.insert(face.GetNeighborPartitionID(grid_));
      }
```

wdhawkins:
> @andrsd @zachhardy @wdhawkins : can you look at this part of the code, lines 175 and up, also below? Now that we are allowing orientation to be PARALLEL with this PR, the following lines will be visiting when the orientation is not OUTGOING because of the `else `statement. So now, the else statement will be active when a face orientation is **either** INCOMING or PARALLEL (before, PARALLEL was **never** active). I am worried that those lines of codes should not happen when the face is PARALLEL.
> 
> ```
>       // If not outgoing determine what it is dependent on
>       else
>       {
>         // if it is a cell and not bndry
>         if (face.has_neighbor_ and not face.IsNeighborLocal(grid_))
>           location_dependencies.insert(face.GetNeighborPartitionID(grid_));
>       }
> ```

This is a good point. There may be several places in the code where we need to fix this if `FOPARALLEL` is going to be a valid value now. We'll want to expand the `if` test (or change it to a `switch`) to explicitly handle at least `FOINCOMING` and `FOOUTGOING`.

ragusa:
> > @andrsd @zachhardy @wdhawkins : can you look at this part of the code, lines 175 and up, also below? Now that we are allowing orientation to be PARALLEL with this PR, the following lines will be visiting when the orientation is not OUTGOING because of the `else `statement. So now, the else statement will be active when a face orientation is **either** INCOMING or PARALLEL (before, PARALLEL was **never** active). I am worried that those lines of codes should not happen when the face is PARALLEL.
> > ```
> >       // If not outgoing determine what it is dependent on
> >       else
> >       {
> >         // if it is a cell and not bndry
> >         if (face.has_neighbor_ and not face.IsNeighborLocal(grid_))
> >           location_dependencies.insert(face.GetNeighborPartitionID(grid_));
> >       }
> > ```
> 
> This is a good point. There may be several places in the code where we need to fix this if `FOPARALLEL` is going to be a valid value now. We'll want to expand the `if` test (or change it to a `switch`) to explicitly handle at least `FOINCOMING` and `FOOUTGOING`.

at least these files:

```
aah_sweep_chunk.cc
cbc_sweep_chunk.cc
lbs_curvilinear_sweep_chunk_pwl.cc

spds.cc
cbc_spds.cc

aah_fluds_common_data.cc
```

andrsd:
- `cbc_spds.cc` seems to be fine, it does `if (ori == INCOMING) {...} else if (ori == OUTGOING) {...}`, i.e. ignores `PARALLEL`
- `aah_fluds_common_data.cc` should be fine, it does `if (ori == SOMETHING) { ... }` and has no else branches and never tests for PARALLEL
- in `aah_sweep_chunk.cc` the test is `if (ori != SOMETHING) continue`, so it seems the test should be changed to `if (ori == PARALLEL or THE_OPPOSITE_OF_SOMETHING) continue;`
- the same for `cbc_sweep_chunk.cc` and `lbs_curvilinear_sweep_chunk_pwl.cc`

Am I on the right track?

andrsd:
I added a patch that fixes SPDS, so that PARALLEL faces are skipped...

andrsd:
> * in `aah_sweep_chunk.cc` the test is `if (ori != SOMETHING) continue`, so it seems the test should be changed to `if (ori == PARALLEL or THE_OPPOSITE_OF_SOMETHING) continue;`

I was thinking about this case and I think the condition should really be:
```
if (ori == OPPOSITE_OF_SOMETHING) { do stuff }
```
This was the original SOMETHING and PARALLEL gets ignored...

andrsd:
Incorporated discussed changed. Please, review when you got some time...

ragusa:
in `aah_sweep_chunks`, we have :
```
        if (face_orientations[f] != FaceOrientation::INCOMING)
          continue;
```
which means "do some work if INCOMING", and later we have :
```
        if (face_orientations[f] != FaceOrientation::OUTGOING)
          continue;
```
which means "do some work if OUTGOING", and later we have :


so, in effect, whenever a face is PARALLEL, it is always ignored.

In `aah_fluds_common_data`, only OUTGOING and INCOMING are dealt with, so PARALLEL is always ignored

So, I think this is ok. @wdhawkins will confirm.

ragusa (approved):
approved

zhardy-lanl (approved):

wdhawkins (approved):

PR #380, opened 2024-10-08
Adding support for Gmsh v4.1 ASCII format. Minor refactoring of Gmsh v2.2 ASCII reader.

### LINKED ISSUE #342: Gmesh mesh reader update
- [ ] allow for latest gmsh format to be read
- [ ] clean up or remove existing gmsh reader

wdhawkins:
This PR adds support for the Gmsh v4.1 ASCII format. The reader has been split into separate routines for v2.2 and v4.1.  The version-specific readers can be used directly, or FromGmsh() will select the correct reader based on the Gmsh format version in the file header. I've also done cleanup and minor refactoring of the v2.2 driver to make it consistent with the 4.1 reader.

Resolves #342.


ragusa (approved):

framework/mesh/io/mesh_io.h
```diff
@@ -19,6 +19,8 @@ class MeshIO
   FromEnsightGold(const UnpartitionedMesh::Options& options);
   static std::shared_ptr<UnpartitionedMesh> FromOBJ(const UnpartitionedMesh::Options& options);
   static std::shared_ptr<UnpartitionedMesh> FromGmsh(const UnpartitionedMesh::Options& options);
+  static std::shared_ptr<UnpartitionedMesh> FromGmshV41(const UnpartitionedMesh::Options& options);
+  static std::shared_ptr<UnpartitionedMesh> FromGmshV22(const UnpartitionedMesh::Options& options);
```
andrsd, line 23: If the user facing API is `FromGmsh`, then `FromGmshV41` and `FromGmshV22` should be protected or private. Unless you want users to be able to call those two, which it does not look like it is the case...

wdhawkins: Fixed.

andrsd (approved):

zhardy-lanl (approved):

PR #381, opened 2024-10-08
Reading/Writing angular fluxes to h5

quincy-huhn98:
Changing flux_io to use h5 file format


ragusa:
look it to use the .clang-format file we have in the repo when utilizing VS code. with https://code.visualstudio.com/docs/cpp/cpp-ide. I am not sure who else is using VS code but if you have issues you can ping @andrsd or @wdhawkins for guidance.

andrsd:
I don't use VS code. But if your clang-format gives you (@quincy-huhn98) different formatting, you want to double check its version. We are using 14. It is known that different clang-formats can produce different formatting even with the same config file...

zhardy-lanl (approved):

andrsd (approved):

wdhawkins (approved):

PR #383, opened 2024-10-15
Adding missing #include

andrsd:
This is fixing a compile issue with clang-14.

### DIFF
--- a/framework/data_types/ndarray.h
+++ b/framework/data_types/ndarray.h
@@ -13,6 +13,7 @@
 #include <algorithm>
 #include <numeric>
 #include <vector>
+#include <string>
 
 namespace opensn
 {


framework/data_types/ndarray.h
```diff
@@ -13,6 +13,7 @@
 #include <algorithm>
 #include <numeric>
 #include <vector>
+#include <string>
```
ragusa, line 16: out of curiosity, why some version of compilers do not complain and some do?

ragusa (approved):

wdhawkins (approved):

framework/data_types/ndarray.h
andrsd: I think it is how they #include internally in the C++ headers

Issue #378, opened 2024-10-03
Implicit integer conversions and float divide-by-zero issues


wdhawkins:
We have a handful of implicit integer conversions and floating point divide-by-zero issues that need to be fixed. They are easy to find with the following compile line:

`cmake -DCMAKE_CXX_FLAGS="-fsanitize=float-divide-by-zero,implicit-conversion -fno-sanitize-recover" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=float-divide-by-zero,implicit-conversion -fno-sanitize-recover" ..`

and running the test suite.

andrsd:
Now, that #406 is merged, I think this can be closed. Thoughts?

PR #385, opened 2024-10-15
Fixing errors found by clang sanitizer

andrsd:
Refs #378

### DIFF
--- a/framework/mesh/mesh_continuum/grid_face_histogram.cc
+++ b/framework/mesh/mesh_continuum/grid_face_histogram.cc
@@ -17,7 +17,7 @@ GridFaceHistogram::NumberOfFaceHistogramBins() const
 size_t
 GridFaceHistogram::MapFaceHistogramBins(size_t num_face_verts) const
 {
-  size_t category_counter = -1;
+  int64_t category_counter = -1;
   for (auto category : face_categories_)
   {
     category_counter++;
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/fluds/aah_fluds_common_data.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/fluds/aah_fluds_common_data.cc
@@ -181,7 +181,8 @@ AAH_FLUDSCommonData::SlotDynamics(const Cell& cell,
         bool found = false;
         for (auto& lock_box_slot : lock_box)
         {
-          if ((lock_box_slot.first == face.neighbor_id_) and (lock_box_slot.second == ass_face))
+          if ((static_cast<uint64_t>(lock_box_slot.first) == face.neighbor_id_) and
+              (lock_box_slot.second == ass_face))
           {
             lock_box_slot.first = -1;
             lock_box_slot.second = -1;
--- a/framework/materials/multi_group_xs/multi_group_xs.cc
+++ b/framework/materials/multi_group_xs/multi_group_xs.cc
@@ -329,7 +329,10 @@ MultiGroupXS::ComputeDiffusionParameters()
 
     // Compute the diffusion coefficient
     // Cap the value for when sigma_t - sigma_1 is near zero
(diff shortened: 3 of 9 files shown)


andrsd:
This should fix all but one error that my clang sanitizer found.

The one I am not able to track down is somewhere in loading OpenSn native cross-sections. I see this error:
```
path/include/c++/v1/istream:1458:23: runtime error: implicit conversion from type 'std::basic_string<char>::size_type' (aka 'unsigned long') of value 18446744073709551599 (64-bit, unsigned) to type 'std::streamsize' (aka 'long') changed the value to -17 (64-bit, signed)
```
This would suggest it is somewhere where the code works with the `stringstream`, but I don't see anything obviously wrong like with the other errors :-(  Another explanation is that this might be a false positive (which does happen...)

framework/mesh/mesh_continuum/grid_face_histogram.cc
```diff
@@ -17,7 +17,7 @@ GridFaceHistogram::NumberOfFaceHistogramBins() const
 size_t
 GridFaceHistogram::MapFaceHistogramBins(size_t num_face_verts) const
 {
-  size_t category_counter = -1;
+  int64_t category_counter = -1;
```
ragusa, line 20: how did you know that this integer should be a long one. it was `size_t `before. just curious

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/fluds/aah_fluds_common_data.cc
```diff
@@ -181,7 +181,8 @@ AAH_FLUDSCommonData::SlotDynamics(const Cell& cell,
         bool found = false;
         for (auto& lock_box_slot : lock_box)
         {
-          if ((lock_box_slot.first == face.neighbor_id_) and (lock_box_slot.second == ass_face))
+          if ((static_cast<uint64_t>(lock_box_slot.first) == face.neighbor_id_) and
+              (lock_box_slot.second == ass_face))
```
ragusa, line 185: @wdhawkins @zachhardy does anyone know what ass_face is, except sounding funny to a 4-year old.

ragusa (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/fluds/aah_fluds_common_data.cc
wdhawkins: Yes, I believe it's the cell neighbor face that is associated with this cell's face (hence ass_face = associated face).

framework/mesh/mesh_continuum/grid_face_histogram.cc
andrsd: `size_t` is unsigned, `int64_t` is signed. That was the problem. Assigning negative numbers into unsigned types is frowned upon by sanitizers....

modules/linear_boltzmann_solvers/discrete_ordinates_solver/sweep/fluds/aah_fluds_common_data.cc
ragusa: I am not so sure because we use a`dj_face, adj_cell`, where adj = adjacent. So, ass_face may be different than adj_face. Exampl using adj_ everywhere:

```
int af = -1;
for (const auto& adj_face : adj_cell.faces_)
{
++af;
std::set<uint64_t> afvids(adj_face.vertex_ids_.begin(),
 adj_face.vertex_ids_.end()); // adj_face vertex ids
```

zhardy-lanl: I had this discussion with Jan in the past. @wdhawkins is correct. It is associated face.

wdhawkins (approved):

PR #387, opened 2024-10-15
Write flux moments to H5

quincy-huhn98:
Change flux moment writing from arbitrary binary file to h5


zhardy-lanl (approved):

ragusa (approved):

andrsd:
Is there a plan to change what is actually gonna be stored inside the hdf5 file? Currently, there is a lot of duplicated or unneeded information stored in the file for the sake of having a simple read procedure, but that is just plainly wrong...

ragusa:
> Is there a plan to change what is actually gonna be stored inside the hdf5 file? Currently, there is a lot of duplicated or unneeded information stored in the file for the sake of having a simple read procedure, but that is just plainly wrong...

I believe this PR is a translate of the binary write into an H5 format. Has new data been added to what was already exported? I would be more than happy to revisit this or have Quincy revisit after we have a plan but if this PR does not add/remove data compared to what was done before, then we should let it go and make a plan during our next meeting.

andrsd:
If you want to just change the file format, then this is ok.

andrsd (approved):

zhardy-lanl:
> > Is there a plan to change what is actually gonna be stored inside the hdf5 file? Currently, there is a lot of duplicated or unneeded information stored in the file for the sake of having a simple read procedure, but that is just plainly wrong...
> 
> I believe this PR is a translate of the binary write into an H5 format. Has new data been added to what was already exported? I would be more than happy to revisit this or have Quincy revisit after we have a plan but if this PR does not add/remove data compared to what was done before, then we should let it go and make a plan during our next meeting.

What in particular are you referring to @andrsd? We briefly discussed this last week I think. I proposed going to attributes and the flux moments vector only. @ragusa brought up the issue of needing cell ids so that you can read the files in on a different number of processors and we decided to keep it this way for now. How would you structure this?

andrsd:
> What in particular are you referring to @andrsd?

1. 
```
  H5WriteDataset1D(file_id, "moments", moments);
  H5WriteDataset1D(file_id, "groups", groups);
```
This is not needed, that is a global info that you store in an attribute. You don't have to store `0..M` and `0..G` for every node. Then, in read, you do your nested for-loop to grab the values from the `values` vector.

2. You store number of nodes per cell, so that the dataset size is `number_of_cells`. Currently it is `number_of_cells * n_nodes * n_moments * n_groups`. The same for `cell_ids`.

Note how all the `push_back`s are done in the most-inner loop 🤮

andrsd:
> EDIT: Looking at it now, we write file `<prefix>_<rank>.h5` and read in the same. It doesn't even seem like we have the capability to read in on a different number of ranks.

If I am reading the code correctly, there is one file for all ranks, but each rank picks its own portion from it. This is probably ok for cases where you would use different partitioning between runs. But, may not be ok for case with millions of ranks - IDK - I never used that many, so IDK how would the OS and file system behave...

zhardy-lanl:
> > What in particular are you referring to @andrsd?
> 
> 
> ```
>   H5WriteDataset1D(file_id, "moments", moments);
>   H5WriteDataset1D(file_id, "groups", groups);
> ```
> 
> This is not needed, that is a global info that you store in an attribute. You don't have to store `0..M` and `0..G` for every node. Then, in read, you do your nested for-loop to grab the values from the `values` vector.
> 
> 2. You store number of nodes per cell, so that the dataset size is `number_of_cells`. Currently it is `number_of_cells * n_nodes * n_moments * n_groups`. The same for `cell_ids`.
> 
> Note how all the `push_back`s are done in the most-inner loop 🤮

The `moments` and `groups` are moment and group indices for a particular flux moment entry, not just the total number of moments and groups. Either way, I agree that they are not necessary. If you have the number of moments and groups in the header than you know the ordering on any given node on a cell and you don't need that information.

zhardy-lanl:
> > EDIT: Looking at it now, we write file `<prefix>_<rank>.h5` and read in the same. It doesn't even seem like we have the capability to read in on a different number of ranks.
> 
> If I am reading the code correctly, there is one file for all ranks, but each rank picks its own portion from it. This is probably ok for cases where you would use different partitioning between runs. But, may not be ok for case with millions of ranks - IDK - I never used that many, so IDK how would the OS and file system behave...

When we write, there is a separate file for all ranks:
```
std::string file_name = file_base + std::to_string(opensn::mpi_comm.rank()) + ".h5";
hid_t file_id = H5Fcreate(file_name.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
```

When we read, there is a flag called `single_file` which gets a file name via
```
const auto file_name =
  file_base + (single_file ? "" : std::to_string(opensn::mpi_comm.rank())) + ".h5";
hid_t file_id = H5Fopen(file_name.c_str(), H5F_ACC_RDONLY, H5P_DEFAULT);
```

Maybe a `single_file` flag should be added to `WriteFluxMoments` so that this can actually be used. Looking now, it does look like the correct data will get set on the correct ranks via the check `grid.IsCellLocal(cell_ids[i])`.

andrsd:
Just to be clear what exactly I am bitching about in 1). The size of `moments` and `groups` is not `M` or `G`, respectively. That would be ok, but still not needed. It is `number_of_cells * n_nodes * n_moments * n_groups`...

zhardy-lanl:
> Just to be clear what exactly I am bitching about in 1). The size of `moments` and `groups` is not `M` or `G`, respectively. That would be ok, but still not needed. It is `number_of_cells * n_nodes * n_moments * n_groups`...

Well it wouldn't make much sense for them to be `M` or `G` if it is a per flux moment entry?

andrsd:
> Well it wouldn't make much sense for them to be `M` or `G` if it is a per flux moment entry?

Yeah, it would not make sense either... I was not trying to say it should be that. Just the plain number of moments and groups stored as an attribute is correct.

quincy-huhn98:
I have eliminated the excess data storage. The h5 file now only contains the values as well as the attributes required to read it

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -15,99 +16,45 @@ LBSSolverIO::WriteFluxMoments(
...
-    file.write((char*)&num_cell_nodes, sizeof(uint64_t));
-
-    const auto nodes = discretization.GetCellNodeLocations(cell);
-    for (const auto& node : nodes)
-    {
-      file.write((char*)&node.x, sizeof(double));
-      file.write((char*)&node.y, sizeof(double));
-      file.write((char*)&node.z, sizeof(double));
-    } // for node
-  }   // for cell
```
andrsd, line 92: You don't want to remove this. The purpose of this code is:  You obtain a flux vector on mesh1. You try use the stored vector with mesh2. We need to check that the meshes are the same. Just the number of cells and nodes is not unique enough to prevent users to use the stored vector with a different mesh. This is why we store the vertex location and in read they are checked that they map on the mesh we already loaded.

andrsd: Also, the global IDs and number of nodes per cell needs to be stored in the file. In general case, every cell could have different number of nodes (the code supports polygon/polyhedron elements). And the global IDs are used in parallel, so that each rank picks its cells. It is this bit in the read function:
```
if (grid.IsCellLocal(cell_global_id))
```
So, bring these bits back.

andrsd (changes requested):

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
zhardy-lanl: A few comments/questions on this for @andrsd. First, the angular flux files in #381 (and preceding code) do not have grid information. Should this be added there? Also, would it be better to introduce a separate h5 file for the grid for these purposes? When we have transients, for example, if we dump the flux moments per time step we would then be writing the same grid information over and over. 

@quincy-huhn98 I would make two groups in the h5 file, one called "grid" or "mesh" and the other called "flux_moments" to separate the information. Everything currently written should go in the "flux_moments" group.

andrsd: I would start with a single file for both steady state and transient. Everything is going to be much simpler - both code-wise and easier for users (in case they need to transfer those files, post-processing, etc.).

Also, if the angular fluxes could be stored in the same file as scalar fluxes, even better :-) But I understand those get real huge real fast. 

One big reason I can see to start doing separate files is when the file size gets so big that the file system cannot handle that - if that's the direction of your thinking...

quincy-huhn98: I was still working out how to implement the grid, so I appreciate the direction @andrsd

quincy-huhn98:
I have made the changes discussed. I decided that if I do make the mesh validation a separate method I will do so on the angular flux pr.

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_io_flux_moments.cc
```diff
@@ -94,7 +94,6 @@ LBSReadSourceMoments(lua_State* L)
   const auto file_base = LuaArg<std::string>(L, 2);
   auto single_file_flag = LuaArgOptional<bool>(L, 3, false);
 
-  // Get pointer to solver
```
ragusa, line 97: just curious why these comments have been removed

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -15,99 +16,71 @@ LBSSolverIO::WriteFluxMoments(
...
-
-  // Write flux moments data
+      nodes_x.push_back(node.x);
+      nodes_y.push_back(node.y);
+      nodes_z.push_back(node.z);
+    }
+  }
+
+  // Write mesh data to h5
```
ragusa, line 60: maybe make the common more explciit that mesh is placed in its own group with group attributes

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -15,99 +16,71 @@ LBSSolverIO::WriteFluxMoments(
...
+  uint64_t num_local_cells = grid.local_cells.size();
+  uint64_t num_local_nodes = discretization.GetNumLocalDOFs(NODES_ONLY);
+  uint64_t num_moments = lbs_solver.NumMoments();
+  uint64_t num_groups = lbs_solver.NumGroups();
+  uint64_t num_local_dofs = num_local_nodes * num_moments * num_groups;
+  OpenSnLogicalErrorIf(src.size() != num_local_dofs, "Incompatible flux moments vector provided.");
 
-  // Write nodal positions
+  H5CreateAttribute(file_id, "num_moments", num_moments);
+  H5CreateAttribute(file_id, "num_groups", num_groups);
```
ragusa, line 41: maybe a comment here to state that those attributes are written int he root of the h5 file

ragusa (changes requested):
add the few lines of comments and then I can change my review status to approved. thx

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -164,86 +134,80 @@ LBSSolverIO::ReadFluxMoments(LBSSolver& lbs_solver,
...
 
     // Map the system nodes to file nodes
-    bool mapping_successful = true; // true until disproven
-    auto& mapping = file_cell_nodal_mapping[file_cell_global_id];
-    for (uint64_t n = 0; n < file_num_cell_nodes; ++n)
+    auto& mapping = file_cell_nodal_mapping[cell_global_id];
+    for (uint64_t n = 0; n < file_num_cell_nodes[c]; ++n)
     {
       bool mapping_found = false;
+      std::cout << std::endl;
```
andrsd, line 176: This looks like a left over statement from debugging. Remove this.

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -164,86 +134,80 @@ LBSSolverIO::ReadFluxMoments(LBSSolver& lbs_solver,
...
       for (uint64_t m = 0; m < nodes.size(); ++m)
-        if ((nodes[m] - file_nodes[n]).NormSquare() < 1.0e-12)
+        if ((nodes[m] - file_nodes[n]).NormSquare() < 1.0e-10)
```
andrsd, line 178: Was there a reason to change the tolerance?

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_io_flux_moments.cc
quincy-huhn98: I'm not sure. Those were removed by @zhardy-lanl in his fork that I got the original implementation from

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
quincy-huhn98: Changed the comment

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
quincy-huhn98: Added a comment

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
quincy-huhn98: I had changed it as part of debugging. I have reverted to the original

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_io_flux_moments.cc
zhardy-lanl: I removed the comments because it felt redundant to me to say we are getting a pointer to the solver when we are calling a routine called `GetStackItem` and assigning it to something that says `solver`. Feel free to add it back.

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -15,99 +16,72 @@ LBSSolverIO::WriteFluxMoments(
...
-  file.write((char*)&num_groups, sizeof(uint64_t));
+  uint64_t num_local_cells = grid.local_cells.size();
+  uint64_t num_local_nodes = discretization.GetNumLocalDOFs(NODES_ONLY);
+  uint64_t num_moments = lbs_solver.NumMoments();
+  uint64_t num_groups = lbs_solver.NumGroups();
+  uint64_t num_local_dofs = num_local_nodes * num_moments * num_groups;
+  OpenSnLogicalErrorIf(src.size() != num_local_dofs, "Incompatible flux moments vector provided.");
 
-  // Write nodal positions
+  // Write number of moments and groups to the root of th h5
```
zhardy-lanl, line 40:
```suggestion
  // Write number of moments and groups to the root of the h5
```

quincy-huhn98: Ahh good catch sorry

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_io_flux_moments.cc
zhardy-lanl: Just leave it then unless anyone else chimes in with an objection

andrsd (approved):

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -15,99 +16,72 @@ LBSSolverIO::WriteFluxMoments(
...
+  std::vector<uint64_t> cell_ids, num_cell_nodes;
+  std::vector<double> nodes_x, nodes_y, nodes_z;
   for (const auto& cell : grid.local_cells)
   {
-    const uint64_t cell_global_id = cell.global_id;
-    const uint64_t num_cell_nodes = discretization.GetCellNumNodes(cell);
-
-    file.write((char*)&cell_global_id, sizeof(uint64_t));
-    file.write((char*)&num_cell_nodes, sizeof(uint64_t));
+    cell_ids.push_back(cell.global_id);
```
wdhawkins, line 49: Since we know the sizes of these vectors, it's best to do a `.reserve` before doing `push_back`.

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
```diff
@@ -118,41 +92,38 @@ LBSSolverIO::ReadFluxMoments(LBSSolver& lbs_solver,
...
-  file.read((char*)&file_num_moments, sizeof(uint64_t));
-  file.read((char*)&file_num_groups, sizeof(uint64_t));
+  uint64_t file_num_groups;
+  H5ReadAttribute(file_id, "num_groups", file_num_groups);
 
   // Check compatibility with system macro info
+  const auto& grid = lbs_solver.Grid();
+  const auto& discretization = lbs_solver.SpatialDiscretization();
   const auto uk_man = lbs_solver.UnknownManager();
   const auto NODES_ONLY = UnknownManager::GetUnitaryUnknownManager();
```
wdhawkins, line 121: While you're refactoring, can you change `NODES_ONLY` so that it's not uppercase?

lua/modules/linear_bolzmann_solvers/lbs_solver/lbs_io_flux_moments.cc
ragusa: "just leave it" as meaning "put the comment back" or "keep it as is, i.e., with the comment currently removed"

zhardy-lanl: Just leave it meaning leave the comment out and don't change the state of the branch to put it back in.

modules/linear_boltzmann_solvers/lbs_solver/io/flux_moments_io.cc
quincy-huhn98: Added this

ragusa (approved):

wdhawkins (approved):

zhardy-lanl (approved):

PR #390, opened 2024-10-17
Changed `TextName` to `Name` and `text_name` to `name`

zhardy-lanl:
If anyone finds other occurrences I missed please let me know.


framework/math/unknown_manager/unknown_manager.h
```diff
@@ -187,7 +187,7 @@ class UnknownManager
   unsigned int AddUnknown(UnknownType unk_type, unsigned int dimension = 0);
 
   /// Maps the unknown's component within the storage of a node.
-  unsigned int MapUnknown(unsigned int unknown_id, unsigned int component = 0) const;
+  unsigned int MapUnknown(int unknown_id, unsigned int component = 0) const;
```
ragusa, line 190: just curious, why the change from `unsigned `int to `int ` for unknown_id? Do we allow those ids to be negative now?

framework/math/unknown_manager/unknown_manager.cc
```diff
@@ -107,23 +101,21 @@ UnknownManager::SetUnknownNumOffBlockConnections(unsigned int unknown_id, int nu
...
+UnknownManager::SetUnknownComponentNumOffBlockConnections(int unknown_id,
                                                           unsigned int component,
                                                           int num_conn)
 {
-  auto& log = Logger::GetInstance();
-
   if (unknown_id < 0 or unknown_id >= unknowns.size())
   {
-    log.LogAllError() << "UnknownManager failed call to SetUnknownComponentTextName,"
+    log.LogAllError() << "UnknownManager failed call to SetUnknownComponentName,"
```
andrsd, line 110: Can you add a patch that is fixing the member function name to `SetUnknownComponentNumOffBlockConnections` which would be the correct name? Also the same problem on line `118`.

andrsd (changes requested):

framework/math/unknown_manager/unknown_manager.h
zhardy-lanl: Because in almost every routine where `unknown_id` was used there was a check for a negative value

framework/math/unknown_manager/unknown_manager.cc
zhardy-lanl: Done.

andrsd (approved):

ragusa (approved):

PR #392, opened 2024-10-29
Adding unstructured mesh scaling studies

wdhawkins:
Adds inputs, mesh geometry, cross-section data files, and Jupyter notebooks for unstructured mesh scaling studies.


tools/scaling/node_to_node/unstructured_mesh/strong/64rpn/strong_scaling.geo
```diff
@@ -0,0 +1,63 @@
+// Parameters
+cube_size = 10.0;          // Side length of the cube
+
+// Calculate characteristic length based on the desired number of tetrahedra
+approx_element_size = cube_size / 39;
```
andrsd, line 5: `gmsh` takes a command line argument to prescribe a variable. So, you can use just one .geo file and parameterize it from a command line.

If you changed this line to:
```
approx_element_size = cube_size /  divisor;
```
then, you could generate the mesh like this:
```
gmsh -3 -setnumber divisor 39 strong_scaling.geo -o mesh.msh
```

andrsd (commented):
Just a suggestion: it may be better to do the dir structure like this:
```
+ scaling
+-+ 64rpn
  +-- weak
  +-- strong
```
Then, you could put the xs files in `scaling/64rpn` and point the input files there.

And you could have just one geo file sitting at the same level and be used by the weak/strong scaling cases.

wdhawkins:
> Just a suggestion: it may be better to do the dir structure like this:
> 
> ```
> + scaling
> +-+ 64rpn
>   +-- weak
>   +-- strong
> ```
> 
> Then, you could put the xs files in `scaling/64rpn` and point the input files there.
> 
> And you could have just one geo file sitting at the same level and be used by the weak/strong scaling cases.

That will definitely work for the xs file... that's what I've been doing in the run, but for the weak scaling case, it's probably better to keep the `.geo` file for each case. It takes a bit of iterating to get get right values for the right number of cells for each case.

tools/scaling/node_to_node/unstructured_mesh/strong/64rpn/strong_scaling.geo
wdhawkins: Oh... now that's nice. I'll take one `.geo` file and write a script that generates all of the runs.

tools/scaling/node_to_node/unstructured_mesh/strong/generate_scaling_study.py
```diff
@@ -0,0 +1,65 @@
...
+gmsh_binary = "/Applications/Gmsh.app/Contents/MacOS/gmsh"
+
+# Node counts
+nnodes =  [ 1,  2,  4,  8, 16, 32, 64, 128, 256, 512 ]
+
+# Tasks (MPI ranks) per node
+ntasks = 64
+
+# Gmsh divisor
+# Generates an unstrudtured mesh with 268,938 tets. At 64 tasks per node, this comes to ~4200 cells
```
andrsd, line 60: "unstrudtured" -> "unstructured"

andrsd (changes requested):

tools/scaling/node_to_node/unstructured_mesh/strong/64rpn/strong_scaling.geo
wdhawkins: Fixed!

tools/scaling/node_to_node/unstructured_mesh/strong/generate_scaling_study.py
wdhawkins: Fixed.

wdhawkins:
> > Just a suggestion: it may be better to do the dir structure like this:
> > ```
> > + scaling
> > +-+ 64rpn
> >   +-- weak
> >   +-- strong
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, you could put the xs files in `scaling/64rpn` and point the input files there.
> > And you could have just one geo file sitting at the same level and be used by the weak/strong scaling cases.
> 
> That will definitely work for the xs file... that's what I've been doing in the run, but for the weak scaling case, it's probably better to keep the `.geo` file for each case. It takes a bit of iterating to get get right values for the right number of cells for each case.

Since the addition of the script greatly simplified things, I think I've addressed this. I did decide to keep copies of the `.geo` file and the `.xs` file in both the `strong` and `weak` directories so that it's possible to just grab the entire directory and have everything needed for that particular scaling study.

andrsd:
> I did decide to keep copies of the `.geo` file and the `.xs` file in both the `strong` and `weak` directories so that it's possible to just grab the entire directory and have everything needed for that particular scaling study.

Treating weak and strong scaling as individual is fine with me. 

On another subject: Do you want to leave the hard-coded paths in the script? Maybe the path to `gmsh` could be just `gmsh`, so that if it is on PATH it will just work (this is how I setup my machines). And if `gmsh` is installed via a package manager this will just work as well - for example ubunutu provides such a [package](https://packages.ubuntu.com/oracular/gmsh).

wdhawkins:
> > I did decide to keep copies of the `.geo` file and the `.xs` file in both the `strong` and `weak` directories so that it's possible to just grab the entire directory and have everything needed for that particular scaling study.
> 
> Treating weak and strong scaling as individual is fine with me.
> 
> On another subject: Do you want to leave the hard-coded paths in the script? Maybe the path to `gmsh` could be just `gmsh`, so that if it is on PATH it will just work (this is how I setup my machines). And if `gmsh` is installed via a package manager this will just work as well - for example ubunutu provides such a [package](https://packages.ubuntu.com/oracular/gmsh).

Fixed. I had forgotten about this. Thanks.

andrsd (approved):

tools/scaling/node_to_node/unstructured_mesh/strong/strong_scaling.ipynb
ragusa, line 1: there is a typo:
dane_nodes = [1, 2, 4, 8, 16, 32, 64, 128, 265]
should be
dane_nodes = [1, 2, 4, 8, 16, 32, 64, 128, **256**]

tools/scaling/node_to_node/unstructured_mesh/weak/weak_scaling.ipynb
ragusa, line 1: there is a typo:
dane_nodes = [1, 2, 4, 8, 16, 32, 64, 128, 265]
should be
dane_nodes = [1, 2, 4, 8, 16, 32, 64, 128, **256**]

tools/scaling/node_to_node/unstructured_mesh/strong/strong_scaling.lua
```diff
@@ -0,0 +1,67 @@
...
+    {
+      groups_from_to = gs1,
+      angular_quadrature_handle = pquad0,
+      angle_aggregation_type = "single",
+      angle_aggregation_num_subsets = 1,
+      groupset_num_subsets = 1,
+      inner_linear_method = "krylov_richardson",
+      l_abs_tol = 1.0e-6,
+      l_max_its = 9,
+      gmres_restart_interval = 10,
```
ragusa, line 57: why is this needed when the linear method is "krylov_richardson"?

if my question makes sense, i think there is a general need to revisit the management of solver options. if so, please create an issue

ragusa (changes requested):
typos in the jupyter notebooks. the notebooks are not super easy to review, one has to view them as a separate file, not a big deal but we should do this.

tools/scaling/node_to_node/unstructured_mesh/strong/strong_scaling.lua
wdhawkins: It's not necessary. Just a leftover from the original input I used for the scaling input.  Fixed.

tools/scaling/node_to_node/unstructured_mesh/strong/strong_scaling.ipynb
wdhawkins: Fixed.

tools/scaling/node_to_node/unstructured_mesh/weak/weak_scaling.ipynb
wdhawkins: Fixed.

ragusa (approved):

PR #393, opened 2024-10-30
Cleaning up use of PETSc iterative method names

wdhawkins:
This PR cleans up the use of PETSc iterative method names and fixes an issue in `WGSLinearSolver` where we were checking for the wrong name. This PR also restricts the use of the PETSc name to the `LinearSolver `class.

### DIFF
--- a/framework/math/linear_solver/linear_solver.cc
+++ b/framework/math/linear_solver/linear_solver.cc
@@ -7,16 +7,9 @@
 namespace opensn
 {
 
-LinearSolver::LinearSolver(const std::string& iterative_method,
+LinearSolver::LinearSolver(const std::string& petsc_iterative_method,
                            std::shared_ptr<LinearSolverContext> context_ptr)
-  : solver_name_(iterative_method), iterative_method_(iterative_method), context_ptr_(context_ptr)
-{
-}
-
-LinearSolver::LinearSolver(const std::string& solver_name,
-                           const std::string& iterative_method,
-                           std::shared_ptr<LinearSolverContext> context_ptr)
-  : solver_name_(solver_name), iterative_method_(iterative_method), context_ptr_(context_ptr)
+  : context_ptr_(context_ptr), petsc_iterative_method_(petsc_iterative_method)
 {
 }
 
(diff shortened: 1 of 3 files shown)


ragusa (approved):

andrsd (approved):

zhardy-lanl (approved):
Thanks for this fix. I may submit another PR that does more cleanup on this. It is strange that `LinearSolver` takes particular string names but `Groupset` uses an enum class that gets mapped to incompatible string names.

wdhawkins:
> Thanks for this fix. I may submit another PR that does more cleanup on this. It is strange that `LinearSolver` takes particular string names but `Groupset` uses an enum class that gets mapped to incompatible string names.

They all used to be the same name (the PETSc names), but we needed a way to distinguish classic Richardson from PETSc Richardson. That led to `classic_richardson` and `krylov_richardson`. I think opensn should stick to `classic_richardson`, `krylov_richardson`, `gmres`, and `bcgs`. This should correspond to the `IterativeMethod` enum. Then we can map to the correct PETSc names via the `IterativeMethodPETScName` function. 

An alternative I think might be more clear is to use: `classic_richardson`, `petsc_richardson`, `petsc_gmres`, `petsc_bcgs` for opensn and then map those to the correct PETSc name. I also think `LinearSolver` could become something like `PETScLinearSolver`.

ragusa:
> An alternative I think might be more clear is to use: `classic_richardson`, `petsc_richardson`, `petsc_gmres`, `petsc_bcgs` for opensn and then map those to the correct PETSc name. I also think `LinearSolver` could become something like `PETScLinearSolver`.

I like this alternative. It is very clear. I propose that we change it to that.

wdhawkins:
> What about shipping the enum class into `LinearSolver` and having the mapping to strings live there? I like the idea of using the new names.

I like that. I'll see about putting up a new PR this weekend with these changes.

zhardy-lanl:
> > What about shipping the enum class into `LinearSolver` and having the mapping to strings live there? I like the idea of using the new names.
> 
> I like that. I'll see about putting up a new PR this weekend with these changes.

I'll try to get that done today so that I can more easily get the SCDSA stuff fixed. If I don't have anything in by the weekend ping me before you start working on it.

zhardy-lanl:
> An alternative I think might be more clear is to use: classic_richardson, petsc_richardson, petsc_gmres, petsc_bcgs for opensn and then map those to the correct PETSc name. I also think LinearSolver could become something like PETScLinearSolver.

On this point, `ClassicRichardson` derives from `LinearSolver` so there would need to be a parent `LinearSolver` class still with a derived `PETScLinearSolver`.

wdhawkins:
> > An alternative I think might be more clear is to use: classic_richardson, petsc_richardson, petsc_gmres, petsc_bcgs for opensn and then map those to the correct PETSc name. I also think LinearSolver could become something like PETScLinearSolver.
> 
> On this point, `ClassicRichardson` derives from `LinearSolver` so there would need to be a parent `LinearSolver` class still with a derived `PETScLinearSolver`.

Probably the way to go. I think that would clean things up nicely since `ClassicRichardson` doesn't need all of the PETSc stuff currently in `LinearSolver`.

PR #394, opened 2024-10-31
Linear Solver Enum

zhardy-lanl:
This PR follow up on #393 and does two things:

1) Moves the `IterativeMethod` enum from LBS into the `LinearSolver` class. Two static methods were added to `LinearSolver`. `IterativeMethodName`  is public and outputs the method name in all caps for logging purposes in various places. `PETScIterativeMethodName` is private and outputs the appropriate PETSc option for the requested solver type. If this method is called when `iterative_method_` is not set to a PETSc solver an error is thrown. At present, this will throw an error only if `LinearSolver` is constructed with `IterativeMethod::CLASSIC_RICHARDSON`. This highlights the need for having a base `LinearSolver` class with a separate `PETScLinearSolver` derived class.

2) Method names were changed from `KRYLOV_*` to `PETSC_*`. Propagated through to the input files, this lead to `gmres` -> `petsc_gmres` and `krylov_richardson` -> `petsc_richardson`.


ragusa (approved):
I have noted that no input files needed to be changed to use` petsc_bcgs`  because we have no such test. Should we add one (in a separate PR)?

framework/math/linear_solver/linear_solver.cc
```diff
@@ -79,21 +72,17 @@ LinearSolver::Setup()
...
-    KSPSetType(ksp_, "richardson");
-  else
-    KSPSetType(ksp_, iterative_method_.c_str());
+  const auto petsc_iterative_method = PETScIterativeMethodName(iterative_method_);
+  KSPSetType(ksp_, petsc_iterative_method.c_str());
 
   ApplyToleranceOptions();
 
-  if (iterative_method_ == "gmres")
+  if (petsc_iterative_method == "gmres")
```
andrsd, line 85: Why not: `iterative_method_ == IterativeMethod::PETSC_GMRES`? This is IMO better than comparing strings...

zhardy-lanl: Good point. I just left the same comparison used before because I was moving quickly. Fixed.

andrsd (approved):

wdhawkins (approved):

PR #396, opened 2024-11-05
Do not warn about C11 extensions when compiling with clang

andrsd:
Since boost is our new dependency, this warning gets triggered, thus silencing it to get a clean
build.


andrsd:
Just a comment to point out that is this [C11](https://en.cppreference.com/w/c/keyword/_Atomic) not C++11...

zhardy-lanl (approved):

andrsd:
Oh, this is against `boost-1.73.0`.
The `1.82.0` does not trigger that warning...

PR #399, opened 2024-11-07
Code standard: Unifying getter's names

andrsd:


andrsd:
This unifies the getter's names by removing the `Get` from the method name as stated in the code standards.

However, there are problems with several APIs where removing the Get prefix yields a type name and that creates compile issues. The list of these APIs is:

- `SpatialDiscretization`
- `Unknown`
- `UnknownManager`
- `ScalarValue`
- `InputParameters` (this is a big one)
- `CellMapping`
- `CoordinateSystem`
- `LogicalVolume`
- `BasicOptions`
- `TimeStepper`
- `SPDS`
- `FLUDS`
- `FaceNodalMapping`
- `SweepChunk`
- `UnitCellMatrices`
- `AGSSolver`
- `WGSContext`
- `QuadratureOrder` (this is an enum, so a `Enum` postfix could resolve this one)

ragusa:
Because of the conflicts, would it not be better or easier to keep and use Get? To be honest, I am not quite sure why setting something uses Set but getting something should not use Get. To me, there's value in the parallelism or symmetry this creates.

PR #400, opened 2024-11-07
Sweep chunk zeroing bug

zhardy-lanl:
The routine `SweepChunk::ZeroDestinationPhi` which gets called within `SweepScheduler::ZeroFluxDataStructures` which gets called within `SweepWGSContext::ApplyInverseTransportOperator` zeros the entire `destination_phi` vector (set to `LBSSolver::phi_new_local_`). This is a problem because sweep chunks operate on groupsets and flux moments vector  is defined over all groups. For single groupset problems, this does not pose an issue, but for multiple groupset problems, it means that with the standard machinery of groupset scoped copying `phi_new_local_` at the end of a solve will be zero for all groupsets other than the last groupset. As a result, the AGS solver actually converges only on the last groupset instead of the full flux moments vector. This PR rewrites the `SweepChunk::ZeroDestinationPhi` routine to zero only the portions of `destination_phi` that correspond to the groupset the sweep chunk is operating on. This results in small diffs on existing tests.

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_adjoint_solver/lbs_adj_solver.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_adjoint_solver/lbs_adj_solver.cc
@@ -101,7 +101,7 @@ DiscreteOrdinatesAdjointSolver::Execute()
         const auto dof_map = cell_view.MapDOF(i, m, 0);
 
         for (int g : set_group_numbers)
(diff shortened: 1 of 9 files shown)


wdhawkins (approved):

wdhawkins:
Nice catch! This bug has been lurking for a long time... and we're close to being able to completely get rid of `phi_old_local`! 🎉

ragusa (approved):

andrsd (approved):

PR #402, opened 2024-11-08
WGS Context Cleanup

zhardy-lanl:
The existing `WGSContext` stores a reference to an `LBSSolver`. The derived classes store this and a reference to the derived class (`DiffusionDFEMSolver` or `DiscreteOrdinatesSolver`). This seems redundant and unnecessary. Instead, storing an `std::shared_ptr<LBSSolver>` and casting where necessary seems more reasonable. This PR implements this.

One test fails: `test/framework/tutorials/tutorial_93_raytracing.lua`. The test segfaults during the call to `Finalize` and I cannot figure out why. Any advice on what the issue is would be much appreciated!

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_solver/iterative_methods/sweep_wgs_context.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_solver/iterative_methods/sweep_wgs_context.cc
@@ -27,8 +27,7 @@ SweepWGSContext::SweepWGSContext(DiscreteOrdinatesSolver& lbs_solver,
     sweep_scheduler(lbs_solver.SweepType() == "AAH" ? SchedulingAlgorithm::DEPTH_OF_GRAPH
                                                     : SchedulingAlgorithm::FIRST_IN_FIRST_OUT,
                     *groupset.angle_agg,
-                    *sweep_chunk),
-    lbs_ss_solver(lbs_solver)
+                    *sweep_chunk)
 {
 }
 
@@ -136,7 +135,7 @@ SweepWGSContext::PostSolveCallback()
   // currently used in OpenSn). This step also zeros out balance variables and computes the correct
   // in-flow and out-flow. Classic Richardson calls this solely to compute balance quantities (note
   // that it does cost us an extra sweep that is technically not necessary).
-  lbs_ss_solver.ZeroOutflowBalanceVars(groupset);
+  dynamic_cast<DiscreteOrdinatesSolver&>(lbs_solver).ZeroOutflowBalanceVars(groupset);
   const auto scope = lhs_src_scope | rhs_src_scope;
   set_source_function(groupset, lbs_solver.QMomentsLocal(), lbs_solver.PhiOldLocal(), scope);
   sweep_scheduler.SetDestinationPhi(lbs_solver.PhiNewLocal());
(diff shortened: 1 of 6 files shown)


wdhawkins:
@zachhardy So `tutorial_93_raytracing.lua` looks to be a bit different from the other tests in that it does a bit of lua, calls some compiled code, and then executes more lua that creates a solver and runs a problem. The new shared pointer code ends up calling the destructor for LBSSolver once the compiled code completes then, when we pick up the second bit of Lua code and try to create and use the solver, bad things happen. There's certainly a lot more to it than this, but it looks like `std::shared_ptr` may not be what we want here.

andrsd:
I see this segfault with tests that use more than 1 groupset, not just the raytracing tutorial. Try running, for example,  `test/modules/linear_boltzmann_solvers/transport_steady/transport_1d_1.lua` from a terminal, not via the `run_tests` script. 

IMO, the problem is that you are creating the shared pointer in the ctor of `WGSContext`.  If you want a shared pointer, I think you want to pass `std::shared_ptr<LBSSolver>` as an argument into `WGSContext` ctor, i.e. create the shared pointer outside the context class.

Also, if you want to store just the reference to `LBSSolver` and dynamic cast to a child class where needed, this works: 
```
auto & solver  = dynamic_cast<DiscreteOrdinatesSolver &>(lbs_solver_reference);
``` 
It is just that you cannot check that the cast was ok using the `nullptr` comparison. But this should throw a `bad_cast` exception if things go wrong.

zhardy-lanl:
> I see this segfault with tests that use more than 1 groupset, not just the raytracing tutorial. Try running, for example, `test/modules/linear_boltzmann_solvers/transport_steady/transport_1d_1.lua` from a terminal, not via the `run_tests` script.
> 
> IMO, the problem is that you are creating the shared pointer in the ctor of `WGSContext`. If you want a shared pointer, I think you want to pass `std::shared_ptr<LBSSolver>` as an argument into `WGSContext` ctor, i.e. create the shared pointer outside the context class.
> 
> Also, if you want to store just the reference to `LBSSolver` and dynamic cast to a child class where needed, this works:
> 
> ```
> auto & solver  = dynamic_cast<DiscreteOrdinatesSolver &>(lbs_solver_reference);
> ```
> 
> It is just that you cannot check that the cast was ok using the `nullptr` comparison. But this should throw a `bad_cast` exception if things go wrong.

Thank you for this! Most recent update uses `dynamic_cast` instead.

andrsd (approved):

wdhawkins (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_solver/iterative_methods/sweep_wgs_context.cc
```diff
@@ -136,7 +135,7 @@ SweepWGSContext::PostSolveCallback()
   // currently used in OpenSn). This step also zeros out balance variables and computes the correct
   // in-flow and out-flow. Classic Richardson calls this solely to compute balance quantities (note
   // that it does cost us an extra sweep that is technically not necessary).
-  lbs_ss_solver.ZeroOutflowBalanceVars(groupset);
+  dynamic_cast<DiscreteOrdinatesSolver&>(lbs_solver).ZeroOutflowBalanceVars(groupset);
```
ragusa, line 138: just curious why the dynamic cast here? I am always suspicious of something underlying. can we add a virtual function ZeroOutflowBalanceVars in the base class so that we don't need the dynamic_cast?

zhardy-lanl: Because the underlying type of `lbs_solver` is `LBSSolver` and not `DiscreteOrdinatesSolver`. Prior to this PR, this context stored an `lbs_solver` of type `LBSSolver&` and `lbs_ss_solver` of type `DiscreteOrdinatesSolver&` (a reference to the same object but cast to each respective type. We can certainly add a virtual method to `LBSSolver`, but I hesitate to go down that path because `LBSSolver` is already so massive and we plan to clean it up.

PR #403, opened 2024-11-08
Extra Sweeps Fix

zhardy-lanl:
### Highlights

- Only do a post-solve sweep for Krylov methods, excluding one-sweep PETSc Richardson.
- Only do a single sweep within the SCDSA eigenvalue acceleration method.

Update (11/12/2024):

After trying to find a bug with @wdhawkins for some time, we concluded that the failing test was just a bad test. `transport_3d_1_poly_qmom_part1.lua` did one sweep, wrote the source moments to file, and then printed field function operation results on an unconverged solution. Without the additional sweep, this changes the values printed in the `part1` test because it was unconverged. The change in sweep behavior changed the behavior in `part2` slightly because it changed what was written to the source moments files. The `part1` test was changed to check for completion only because the values are meaningless. The `part2` gold values were very slightly modified to account for the changed source moments files.

### DIFF
--- a/modules/linear_boltzmann_solvers/executors/pi_keigen_scdsa.cc
+++ b/modules/linear_boltzmann_solvers/executors/pi_keigen_scdsa.cc
@@ -171,31 +171,24 @@ PowerIterationKEigenSCDSA::Execute()
     // Set the fission source
     SetLBSFissionSource(phi_old_local_, false);
     Scale(q_moments_local_, 1.0 / k_eff_);
-
-    auto Sf_ell = q_moments_local_;
     auto Sf0_ell = CopyOnlyPhi0(front_gs_, q_moments_local_);
 
-    // This solves the inners for transport
+    // Store the last power iteration scalar flux
+    auto phi0_ell = CopyOnlyPhi0(front_gs_, phi_old_local_);
+
+    // Solve transport inners
     ags_solver_->Solve();
 
-    // lph_i = l + 1/2,i
-    auto phi0_lph_i = CopyOnlyPhi0(front_gs_, phi_new_local_);
-
-    // Now we produce lph_ip1 = l + 1/2, i+1
-    q_moments_local_ = Sf_ell; // Restore 1/k F phi_l
-    SetLBSScatterSource(phi_new_local_, true);
-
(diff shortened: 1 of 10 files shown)


andrsd (approved):

modules/linear_boltzmann_solvers/executors/pi_keigen_scdsa.cc
```diff
@@ -176,17 +176,8 @@ PowerIterationKEigenSCDSA::Execute()
     auto Sf0_ell = CopyOnlyPhi0(front_gs_, q_moments_local_);
 
     // This solves the inners for transport
+    auto phi0_lph_i = CopyOnlyPhi0(front_gs_, phi_old_local_);
```
ragusa, line 179: I know the variable had that name before this PR, but what does it mean? Can a comment be added?

modules/linear_boltzmann_solvers/executors/pi_keigen_scdsa.cc
```diff
@@ -254,7 +245,7 @@ PowerIterationKEigenSCDSA::Execute()
     double reactivity = (k_eff_ - 1.0) / k_eff_;
 
     // Check convergence, bookkeeping
-    k_eff_change = fabs(k_eff_ - k_eff_prev) / k_eff_;
+    k_eff_change = std::fabs(k_eff_ / k_eff_prev - 1.0);
     k_eff_prev = k_eff_;
```
ragusa, line 249: Should some of these be moved inside the verbose statement if they are only used there?

wdhawkins (approved):

modules/linear_boltzmann_solvers/executors/pi_keigen_scdsa.cc
zhardy-lanl: Moved reactivity to the verbose statement.

modules/linear_boltzmann_solvers/executors/pi_keigen_scdsa.cc
zhardy-lanl: Modified notation and added comments.

ragusa (approved):

PR #406, opened 2024-11-09
Adding additional checks to nightly sanitizer test

wdhawkins:
This PR along with #385 resolves issue #378.

### DIFF
--- a/.github/workflows/builds.yaml
+++ b/.github/workflows/builds.yaml
@@ -51,24 +51,20 @@ jobs:
         shell: bash
         run: |
           export MODULEPATH=/scratch-local/software/modulefiles
-          module load opensn/clang/14.0.6
-          mkdir build && cd build
-          cmake -DCMAKE_CXX_FLAGS="-O2 -fsanitize=address,undefined,nullability \
-                -fsanitize-address-use-after-scope -fno-sanitize-recover" \
-                -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined,nullability \
-                -fsanitize-address-use-after-scope -fno-sanitize-recover" ..
-          make -j && cd .. 
+          module load opensn/clang/17.0.6
+          cmake --preset clang+debug+sanitizer
+          cmake --build --preset clang+debug+sanitizer -j64
       - name: test
         shell: bash
         run: |
           export MODULEPATH=/scratch-local/software/modulefiles
-          module load opensn/clang/14.0.6
+          module load opensn/clang/17.0.6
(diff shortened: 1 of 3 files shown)


andrsd:
I don't want to throw a monkey wrench into your work, but if you want to do something cool with this, consider this. Add a file called `CMakePresets.json` into repo root:
```
{
  "version": 6,
  "configurePresets": [
    {
      "name": "debug+sanitizer",
      "displayName": "Debug with sanitizer",
      "generator": "Unix Makefiles",
      "binaryDir": "build-debug-sanitizer",
      "cacheVariables": {
        "CMAKE_BUILD_TYPE": "Debug",
        "CMAKE_CXX_FLAGS_DEBUG": "-g -Wall -fsanitize=address,float-divide-by-zero,implicit-conversion -fno-sanitize-recover -fno-omit-frame-pointer",
        "CMAKE_C_FLAGS_DEBUG": "-g -Wall -fsanitize=address,float-divide-by-zero,implicit-conversion -fno-sanitize-recover -fno-omit-frame-pointer"
      }
    }
  ],
  "buildPresets": [
    {
      "name": "debug+sanitizer",
      "displayName": "Debug with sanitizer",
      "configurePreset": "debug+sanitizer",
      "configuration": "Debug with sanitizer"
    }
  ]
}
```

Then, one can do:
```
cmake --preset=debug+sanitizer
cmake --build --preset=debug+sanitizer
```

You can do this in the workflow, or any other machine where you want to run the sanitizer target. The build will be configured and built in `build-debug-sanitizer` dir, but you can change that by changing `binaryDir` variable. And you probably want to change the FLAGS to match your new ones and maybe use `RelWithDebInfo` instead of `Debug` for the build type (this is from another project where I want to sue the Debug settings). And probably change the name of the preset as well...

No need to do this in this PR, just an idea how to make the sanitization work simpler for us. Other presets can be added later...

ragusa (approved):

wdhawkins:
> No need to do this in this PR, just an idea how to make the sanitization work simpler for us. Other presets can be added later...
> 
> NOTE: one runs the cmake command from the repo root

I like this a lot.  My only complaint is that I can't query compiler ids from within a preset. It would be nice to be able to disable a preset if the user isn't use a compatible compiler.

andrsd:
I don't think presets are meant to be this smart. They only lump variables together so you don't have to type the whole thing and make a typo. If you need configs for different compilers, you just create multiple presets like "release-clang", "release-gcc", etc. and use it on the specific platform. At least that's what I saw one project to do...

andrsd (approved):

Issue #409, opened 2024-11-14
Explore alternative partitioners


wdhawkins:
METIS typically works well and is exceptionally fast on a single core with a large numbers of cells. Unfortunately, it can generate zero-cell partitions. We need to find a partitioner that is both fast and robust or at least allows for a rebalancing step. SCOTCH is too slow for meshes with lots of cells.  It may also be necessary to do our partitioning outside of PETSc. In any case, this is an area that needs exploration and testing.

PR #410, opened 2024-11-14
Simple partition rebalancing

wdhawkins:
This PR adds a simple partition rebalancing algorithm that allows the code to continue execution if the partitioner creates partitions with zero cells. The algorithm computes the average number of cells per partition and moves cell from heavier partitions to the partition with zero cells. This algorithm is a last resort as it will likely destroy the spatial locality of the partitioning.

See issue #409.

### DIFF
--- a/framework/mesh/mesh_generator/mesh_generator.cc
+++ b/framework/mesh/mesh_generator/mesh_generator.cc
@@ -198,7 +198,52 @@ MeshGenerator::PartitionMesh(const UnpartitionedMesh& input_umesh, int num_parti
   std::vector<int64_t> cell_pids =
     partitioner_->Partition(cell_graph, cell_centroids, num_partitions);
 
+  RebalancePartitions(cell_pids, num_partitions);
+
   return cell_pids;
+}
+
+void
+MeshGenerator::RebalancePartitions(std::vector<int64_t>& cell_pids, int num_partitions)
+{
+  // Count the number of cells in each partition
+  std::vector<int> cells_per_partition(num_partitions, 0);
+  for (int partition : cell_pids)
+    cells_per_partition[partition]++;
+
+  // Check if any partition has zero cells
+  if (std::none_of(cells_per_partition.begin(),
+                   cells_per_partition.end(),
+                   [](int count) { return count == 0; }))
+  {
+    return;
+  }
+
+  // Redistributed cells from heavy partitions
+  int total_cells = cell_pids.size();
+  int target = total_cells / num_partitions;
+
+  for (int partition = 0; partition < num_partitions; ++partition)
+  {
+    while (cells_per_partition[partition] > target)
+    {
+      auto it = std::min_element(cells_per_partition.begin(), cells_per_partition.end());
+      int min_partition = std::distance(cells_per_partition.begin(), it);
+      if (min_partition == partition or cells_per_partition[min_partition] >= target)
+        break;
+
+      for (auto& cell_pid : cell_pids)
+      {
+        if (cell_pid == partition)
+        {
+          cell_pid = min_partition;
+          cells_per_partition[partition]--;
+          cells_per_partition[min_partition]++;
+          break;
+        }
+      }
+    }
+  }
 }
 
 std::shared_ptr<MeshContinuum>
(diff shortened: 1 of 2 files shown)


framework/mesh/mesh_generator/mesh_generator.cc
```diff
@@ -198,9 +198,54 @@ MeshGenerator::PartitionMesh(const UnpartitionedMesh& input_umesh, int num_parti
...
+  for (int partition : cell_pids)
+    cells_per_partition[partition]++;
+
+  // Check if any partition has zero cells
+  if (not std::any_of(cells_per_partition.begin(),
+                      cells_per_partition.end(),
+                      [](int count) { return count == 0; }))
+  {
+    return;
+  }
```
zhardy-lanl, line 220: I'm not sure if it is a coding standard thing, but should we use braces for one-liners like this?

zhardy-lanl (approved):

ragusa (approved):

framework/mesh/mesh_generator/mesh_generator.cc
```diff
@@ -198,9 +198,54 @@ MeshGenerator::PartitionMesh(const UnpartitionedMesh& input_umesh, int num_parti
...
+  int total_cells = cell_pids.size();
+  int target = total_cells / num_partitions;
+
+  for (int partition = 0; partition < num_partitions; ++partition)
+  {
+    while (cells_per_partition[partition] > target)
+    {
+      auto it = std::min_element(cells_per_partition.begin(), cells_per_partition.end());
+      int min_partition = std::distance(cells_per_partition.begin(), it);
+      if (min_partition == partition || cells_per_partition[min_partition] >= target)
```
andrsd, line 232: Code standard: `||` -> `or`

andrsd (changes requested):

framework/mesh/mesh_generator/mesh_generator.cc
wdhawkins: If the `if` statement has been a single line, I would agree that it wouldn't be necessary to use `{ }`. Since the `if` statement is multi-line, I think the braces help with clarity.

framework/mesh/mesh_generator/mesh_generator.cc
wdhawkins: Fixed.

framework/mesh/mesh_generator/mesh_generator.cc
ragusa: I agree with Daryl. The if statement that s long, so the curly braces help with readability in my opinion

andrsd: I understood that the discussion was about:

```
if (not std:any())
  return;
```
vs.
```
if (not std::any_of())
{
  return;
}
```

Are you guys arguing that the length of the conditional of the if-statement would determine the use of the braces for the "body"?

ragusa: I believe the discussion is about curly braces or not, for one-liners. I suppose we are pondering when to allow the use of `{...}` when the conditional is complex and we think the use of braces enhance readability. To me, they do.

on readability, is this below not equivalent and a bit shorter?
```
if (std::none_of(cells_per_partition.begin(),
                 cells_per_partition.end(),
                 [](int count) { return count == 0; }))
```

wdhawkins: > Are you guys arguing that the length of the conditional of the if-statement would determine the use of the braces for the "body"?

If the body of the conditional spans multiple lines, I always use braces. If the body fits on a single line, I don't use braces. For me, it helps with clarity when I'm quickly scanning code.

wdhawkins: > on readability, is this below not equivalent and a bit shorter?
> 
> ```
> if (std::none_of(cells_per_partition.begin(),
>                  cells_per_partition.end(),
>                  [](int count) { return count == 0; }))
> ```

It is. I'll make this change.

wdhawkins: Fixed.

andrsd (approved):

PR #411, opened 2024-11-16
Improving performance of assembling routines in `DiffusionPWLCSolver` and `DiffusionMIPSolver`

andrsd:
This PR improves the assembling routines in `DiffusionPWLCSolver` and `DiffusionMIPSolver`:

- the dof map is computed just once
- the local contributions for the matrix and right-hand side vectors are accumulated into a dense matrix/vector, then added to the global matrix, vector

Impact observed on the C5G7 benchmark:

### `AssembleAand_b` routine:

| Case | `main` | `optimized` | Speedup |
| --- | ---: | ---: | ---:|
| `DiffusionMIPSolver` | 11.189 | 5.619 | 1.9915 |
| `DiffusionPWLCSolver` | 1.422 | 0.919 | 1.5467 |

### `Assemble_b` routine:

| Case | `main` | `optimized` | Speedup |
| --- | ---: | ---: | ---:|
| `DiffusionMIPSolver` | 0.170 | 0.153 | 1.1114 |
| `DiffusionPWLCSolver` | 0.238 | 0.223 | 1.0654 |

All times are in seconds.

### DIFF
--- a/modules/linear_boltzmann_solvers/lbs_solver/acceleration/diffusion_mip_solver.cc
+++ b/modules/linear_boltzmann_solvers/lbs_solver/acceleration/diffusion_mip_solver.cc
@@ -75,9 +75,17 @@ DiffusionMIPSolver::AssembleAand_b_wQpoints(const std::vector<double>& q_vector)
 
     const auto& xs = mat_id_2_xs_map_.at(cell.material_id);
 
+    DenseMatrix<double> cell_A(num_nodes, num_nodes);
+    Vector<double> cell_rhs(num_nodes);
+    Vector<int64_t> cell_idxs(num_nodes);
     // For component/group
     for (size_t g = 0; g < num_groups; ++g)
     {
+      cell_A.Set(0.);
+      cell_rhs.Set(0.);
+      for (size_t i = 0; i < num_nodes; ++i)
+        cell_idxs(i) = sdm_.MapDOF(cell, i, uk_man_, 0, g);
+
       // Get coefficient and nodal src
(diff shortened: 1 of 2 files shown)


ragusa (approved):
Good speedup!

PS: the method names are a bit inconsistent (not your fault).
`assembleAand_b` and `assemble_b` do not show symmetry in their names. should we change the first one to `assemble_A_and_b` ?

andrsd:
> `assembleAand_b` and `assemble_b` do not show symmetry in their names. should we change the first one to `assemble_A_and_b` ?

I think we talked about this once before, but no decision was made. Strictly speaking, the names should be `AssembleAAndB` and `AssembleB`, but capital `B` would suggest `b` might be a matrix. That's probably why there is `_` in the name. Alternatively `AssembleMatrixAndRhs` and `AssembleRhs` - we use `PascalCase` for function names...

ragusa:
> > `assembleAand_b` and `assemble_b` do not show symmetry in their names. should we change the first one to `assemble_A_and_b` ?
> 
> I think we talked about this once before, but no decision was made. Strictly speaking, the names should be `AssembleAAndB` and `AssembleB`, but capital `B` would suggest `b` might be a matrix. That's probably why there is `_` in the name. Alternatively `AssembleMatrixAndRhs` and `AssembleRhs` - we use `PascalCase` for function names...

ok. I could be fine with this.

wdhawkins (approved):

zhardy-lanl (approved):

PR #413, opened 2024-11-19
Adding CUDA support to CMake

wdhawkins:
This PR adds CUDA support to our CMake configuration. It has been tested with both a cuBLAS sweep kernel and a custom CUDA sweep kernel. 

This PR also bumps our required CMake version to 3.18. This is necessary to enable auto detection of the CUDA compute architecture. How well it works is questionable, but it's better than nothing. The architecture can also be set via the command line if auto detection fails.

I'm definitely not a CMake expert, so suggestions to improve this are appreciated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.20.2)
 
 project(opensn
     VERSION 0.0.1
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -32,13 +32,25 @@ if (CMAKE_CXX_FLAGS_NATIVE STREQUAL "")
     endif()
 endif()
 
+option(ENABLE_CUDA "Enable CUDA support" OFF)
```
andrsd, line 35: Rename `ENABLE_CUDA` to `OPENSN_WITH_CUDA`. The prefix is important when OpenSn is used as a library, to prevent name collisions with other CMAKE variables defined by other projects. There is no namespacing for variables in cmake, so projects have to add the prefix themselves...

And we are using `WITH`. For example, see below `OPENSN_WITH_LUA`.

CMakeLists.txt
```diff
@@ -125,12 +139,24 @@ else()
...
+      framework/*.cc
+      modules/*.cc
+      modules/*.cu
+  )
+else()
+  file(GLOB_RECURSE LIBOPENSN_SRCS CONFIGURE_DEPENDS
+      framework/*.cc
+      modules/*.cc
+  )
+endif()
```
andrsd, line 153: Can you try this instead:
```
file(GLOB_RECURSE LIBOPENSN_SRCS CONFIGURE_DEPENDS
    framework/*.cc
    modules/*.cc
)

if (OPENSN_WITH_CUDA)
    file(GLOB_RECURSE LIBOPENSN_CUDA_SRCS CONFIGURE_DEPENDS
        modules/*.cu
    )
    list(APPEND LIBOPENSN_SRCS ${LIBOPENSN_CUDA_SRCS})
endif()

andrsd (changes requested):

ragusa (approved):

zhardy-lanl (approved):

CMakeLists.txt
wdhawkins: Fixed.

CMakeLists.txt
wdhawkins: Fixed.

andrsd (approved):

PR #420, opened 2024-11-22
Refactoring API for computing norms

### LINKED ISSUE #351: Refactoring vector norms
Currently, there are these functions:

- `Vec1Norm`
- `Vec2Norm`
- `VecInfinityNorm`
- `VecPNorm`

and there is also `enum class NormType : int { L1_NORM, L2_NORM, LINF_NORM }`.

I suggest this gets refactored. To match the upcoming change for matrix and vector operations, the functions names would be:

- `NormL1`, `NormL2`, etc.
- or, there could be only `Norm(Vector, NormType)`

andrsd:
Functions are renamed as:
- Vec1Norm -> L1Norm
- Vec2Norm -> L2Norm
- VecInfinityNorm -> LInfNorm
- VecPNorm -> LpNorm

Closes #351

### DIFF
--- a/framework/math/math.cc
+++ b/framework/math/math.cc
@@ -100,7 +100,7 @@ Mult(const std::vector<double>& x, const double& val)
 }
 
 double
-Vec1Norm(const std::vector<double>& x)
+L1Norm(const std::vector<double>& x)
 {
   // Local Variables
   size_t n = x.size();
@@ -113,7 +113,7 @@ Vec1Norm(const std::vector<double>& x)
 }
 
 double
-Vec2Norm(const std::vector<double>& x)
+L2Norm(const std::vector<double>& x)
 {
   // Local Variables
   size_t n = x.size();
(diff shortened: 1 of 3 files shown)


wdhawkins (approved):

ragusa (approved):

zhardy-lanl:
Just want to ask the question before giving the green light. How does everyone feel about `NormL1` versus `L1Norm`. The latter is my preference because it is the way we use it is spoken.

andrsd:
I think I suggested `NormL1` because after stripping `Vec` from `Vec1Norm` we would get `1Norm` and that's a no go. However with `L1Norm`, i.e. having the function start with `L`, we do not have that problem. So, either one works for me...

ragusa:
Most people think of "L1 Norm" rather than "Norm L1." So I agree with Zach and we should go with
- `L1Norm`
- `L2Norm`
- `LpNorm`
- `LInfNorm` is a slight preference of mine over `LInfinityNorm` just because it's shorter

andrsd:
Changes were made...

zhardy-lanl (approved):

PR #424, opened 2024-11-23
computing volumes per material ID

ragusa:
Adding functionality to compute the volume of each material ID.

### DIFF
--- a/framework/mesh/mesh_continuum/mesh_continuum.cc
+++ b/framework/mesh/mesh_continuum/mesh_continuum.cc
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 #include "framework/mesh/mesh_continuum/mesh_continuum.h"
+#include "framework/math/spatial_discretization/finite_element/piecewise_linear/piecewise_linear_continuous.h"
 #include "framework/mesh/mesh_continuum/grid_face_histogram.h"
 #include "framework/mesh/mesh_continuum/grid_vtk_utils.h"
 #include "framework/mesh/logical_volume/logical_volume.h"
@@ -650,6 +651,80 @@ MeshContinuum::SetBoundaryIDFromLogical(const LogicalVolume& log_vol,
 
   if (global_num_faces_modified > 0 and grid_bndry_id_map.count(bndry_id) == 0)
     grid_bndry_id_map[bndry_id] = boundary_name;
+}
+
+void
+MeshContinuum::ComputeVolumePerMaterialID() const
+{
+  auto grid_ptr = GetCurrentMesh();
+  const auto& grid = *grid_ptr;
+  std::shared_ptr<SpatialDiscretization> sdm_ptr = PieceWiseLinearContinuous::New(grid);
+  const auto& sdm = *sdm_ptr;
+
+  // Create a map to hold local volume with local material as key
+  std::map<int, double> material_volumes;
+  for (auto& cell : local_cells)
+  {
+    const auto& cell_mapping = sdm.GetCellMapping(cell);
+    const double volume = cell_mapping.CellVolume();
+    material_volumes[cell.material_id] += volume;
+  }
+
+  // Collect all local material IDs
+  std::set<int> unique_material_ids;
+  for (const auto& pair : material_volumes)
+    unique_material_ids.insert(pair.first);
+  // convert set to vector
+  std::vector<int> local_material_ids(unique_material_ids.begin(), unique_material_ids.end());
+  int local_size = local_material_ids.size();
+
+  // Initialize vector to hold sizes from all processes
+  std::vector<int> all_sizes(mpi_comm.size());
+  // Gather all local material ID sizes from all processes
+  mpi_comm.all_gather(local_size, all_sizes);
+
+  // Compute the displacement and total size
+  std::vector<int> displs(mpi_comm.size(), 0);
+  int total_size = 0;
+  for (int i = 0; i < mpi_comm.size(); ++i)
+  {
+    displs[i] = total_size;
+    total_size += all_sizes[i];
+  }
+
+  // Initialize vector to hold all material IDs from all processes
+  std::vector<int> global_material_ids(total_size);
+  // Gather all material IDs at root
+  mpi_comm.all_gather(local_material_ids, global_material_ids, all_sizes, displs);
+
+  // Create a union of all unique material IDs
+  std::set<int> global_unique_material_ids(global_material_ids.begin(), global_material_ids.end());
+
+  // Assign unique material IDs for global reduction
+  global_material_ids.assign(global_unique_material_ids.begin(), global_unique_material_ids.end());
+  std::vector<double> local_volumes(global_material_ids.size(), 0.0);
+  std::vector<double> global_volumes(global_material_ids.size(), 0.0);
+
+  // Fill local volumes vector based on the local material volumes
+  // and perform the reduction one material at a time
+  std::map<int, double> global_material_volumes;
+  for (size_t i = 0; i < global_material_ids.size(); ++i)
+  {
+    if (material_volumes.find(global_material_ids[i]) != material_volumes.end())
+      local_volumes[i] = material_volumes[global_material_ids[i]];
+
+    mpi_comm.all_reduce(local_volumes[i], global_volumes[i], mpi::op::sum<double>());
+    global_material_volumes[global_material_ids[i]] = global_volumes[i];
+  }
+
+  // Output the volumes per material_id on the root process
+  if (opensn::mpi_comm.rank() == 0)
+  {
+    for (const auto& pair : global_material_volumes)
+      log.Log() << "Material ID: " << pair.first << " Volume: " << std::setprecision(12)
+                << pair.second << std::endl;
+    log.Log() << std::endl;
+  }
 }
 
 } // namespace opensn
--- a/framework/mesh/mesh_continuum/mesh_continuum.h
+++ b/framework/mesh/mesh_continuum/mesh_continuum.h
@@ -17,9 +17,7 @@ class MPICommunicatorSet;
 class GridFaceHistogram;
 class MeshGenerator;
 
-/**
- * Stores the relevant information for completely defining a computationaldomain.
- */
+/// Encapsulates all the necessary information required to fully define a computational domain.
 class MeshContinuum
 {
 public:
@@ -44,8 +42,7 @@ public:
 
   static std::shared_ptr<MeshContinuum> New() { return std::make_shared<MeshContinuum>(); }
 
-  /**Method to be called if cells and nodes have been transferred
-   * to another grid.*/
+  /// Method to be called if cells and nodes have been transferred to another grid.
   void ClearCellReferences()
   {
     local_cells_.clear();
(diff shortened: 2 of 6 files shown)


framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -652,4 +653,88 @@ MeshContinuum::SetBoundaryIDFromLogical(const LogicalVolume& log_vol,
...
+  for (size_t i = 0; i < global_material_ids.size(); ++i)
+  {
+    global_material_volumes[global_material_ids[i]] = global_volumes[i];
+  }
+
+  // Output the volumes per material_id on the root process
+  if (opensn::mpi_comm.rank() == 0)
+  {
+    for (const auto& pair : global_material_volumes)
+      log.Log() << "Material ID: " << pair.first << std::setprecision(12)
```
andrsd, line 734: it would make more sense to have the `std::setprecision` in front of `pair.second`.

framework/mesh/mesh_continuum/mesh_continuum.h
```diff
@@ -186,6 +186,11 @@ class MeshContinuum
 
   void SetOrthoAttributes(const OrthoMeshAttributes& attrs) { ortho_attributes_ = attrs; }
 
+  /**
+   * Compute volume per material id's
+   */
```
andrsd, line 191: Single line doxygen comments are `///`, [code standard](https://open-sn.github.io/opensn/coding_standard.html#comments)

lua/framework/mesh/mesh_compute_vol.cc
```diff
@@ -0,0 +1,29 @@
+// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#include "lua/framework/mesh/mesh_compute_vol.h"
+#include "framework/mesh/mesh_continuum/mesh_continuum.h"
+#include "framework/runtime.h"
+#include "framework/logging/log.h"
+#include "lua/framework/console/console.h"
+// #include "framework/mesh/logical_volume/logical_volume.h"
```
andrsd, line 9: Remove this line

lua/framework/mesh/mesh_compute_vol.cc
```diff
@@ -0,0 +1,29 @@
...
+{
+
+RegisterLuaFunctionInNamespace(MeshComputeVolumePerMaterialID, mesh, ComputeVolumePerMaterialID);
+
+using namespace opensn;
+
+int
+MeshComputeVolumePerMaterialID(lua_State* L)
+{
+  // const std::string fname = "mesh.ComputeVolumePerMaterialID";
```
andrsd, line 21: Remove this line

lua/framework/mesh/mesh_compute_vol.h
```diff
@@ -0,0 +1,16 @@
...
+#pragma once
+
+#include "lua/framework/lua.h"
+
+namespace opensnlua
+{
+
+/**
+ * Compute the volume per material ID
+ */
```
andrsd, line 13: Single line comments are `///`

andrsd (changes requested):

test/framework/mesh/tests.json
```diff
@@ -27,5 +27,24 @@
...
+      {
+        "type": "KeyValuePair",
+        "key": "[0]  Material ID: 0 Volume:",
+        "goldvalue": 12.5,
+        "abs_tol": 1.0e-6
+      },
+      {
+        "type": "KeyValuePair",
+        "key": "[0]  Material ID: 1 Volume:",
+        "goldvalue": 12.5,
```
andrsd, line 45: It would make the test more robust if it was set up such that volume of material 1 was different than volume of material 0.

andrsd:
Also, squash the 3 commits into one...

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: fixed

framework/mesh/mesh_continuum/mesh_continuum.h
ragusa: fixed

lua/framework/mesh/mesh_compute_vol.h
ragusa: fixed

lua/framework/mesh/mesh_compute_vol.cc
ragusa: removed

lua/framework/mesh/mesh_compute_vol.cc
ragusa: removed

test/framework/mesh/tests.json
ragusa: done

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -652,4 +653,88 @@ MeshContinuum::SetBoundaryIDFromLogical(const LogicalVolume& log_vol,
...
+  std::vector<int> local_material_ids(unique_material_ids.begin(), unique_material_ids.end());
+  int local_size = local_material_ids.size();
+
+  // Initialize vector to hold sizes from all processes
+  std::vector<int> all_sizes(mpi_comm.size());
+  // Gather all local material ID sizes from all processes
+  mpi_comm.all_gather(local_size, all_sizes);
+
+  // Compute the displacement and total size
+  std::vector<int> displs(mpi_comm.size(), 0); // Initialize displacement vector
```
zhardy-lanl, line 687:
```suggestion
  std::vector<int> displs(mpi_comm.size(), 0);
```
Seems redundant.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -652,4 +653,88 @@ MeshContinuum::SetBoundaryIDFromLogical(const LogicalVolume& log_vol,
...
+  int total_size = 0;
+  for (int i = 0; i < mpi_comm.size(); ++i)
+  {
+    displs[i] = total_size;
+    total_size += all_sizes[i];
+  }
+
+  // Initialize vector to hold all material IDs from all processes
+  std::vector<int> all_material_ids(total_size);
```
zhardy-lanl, line 696:
```suggestion
  std::vector<int> global_material_ids(total_size);
```

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -652,4 +653,88 @@ MeshContinuum::SetBoundaryIDFromLogical(const LogicalVolume& log_vol,
...
+    mpi_comm.all_reduce(loc_vol, glo_vol, mpi::op::sum<double>());
+    global_volumes[i] = glo_vol;
+  }
+
+  // Combine results into a map on each process
+  std::map<int, double> global_material_volumes;
+  for (size_t i = 0; i < global_material_ids.size(); ++i)
+  {
+    global_material_volumes[global_material_ids[i]] = global_volumes[i];
+  }
```
zhardy-lanl, line 728:
```suggestion
  // Fill local volumes vector based on the local material volumes
  // and perform the reduction one material at a time
  std::map<int, double> global_material_volumes;
  for (size_t i = 0; i < global_material_ids.size(); ++i)
  {
    if (material_volumes.find(global_material_ids[i]) != material_volumes.end())
      local_volumes[i] = material_volumes[global_material_ids[i]];
      
    mpi_comm.all_reduce(local_volumes[i], global_volumes[i], mpi::op::sum<double>());
    global_material_volumes[global_material_ids[i]] = global_volumes[i];
  }
```
Would this work?

framework/mesh/mesh_continuum/mesh_continuum.h
```diff
@@ -186,6 +186,9 @@ class MeshContinuum
 
   void SetOrthoAttributes(const OrthoMeshAttributes& attrs) { ortho_attributes_ = attrs; }
 
+  /// Compute volume per material id's
+  void ComputeVolumePerMaterialID() const;
```
zhardy-lanl, line 190:
```suggestion
  void ComputeVolumePerMaterialID() const;
```
Redundant comment. When the volumetric source conversion is finished we should remove `ID` from this.

zhardy-lanl (changes requested):
I think it would be useful to find some way to print out a table of containing the material name and volume (and later density and mass) as well. Doesn't need to be done in this PR.

framework/mesh/mesh_continuum/mesh_continuum.h
ragusa: I do not know what to do. I orginally had `/** ... */` because other declarations just above were like that, then David asked me to go with `/// ...` and now you ask me to remove it altogether.

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: good question. I think the `mpi_comm.all_reduce` did not allow this when i coded it 5 months ago. Since then, David added that functionality. I will test it

framework/mesh/mesh_continuum/mesh_continuum.h
zhardy-lanl: David was correct that one-line comments should be `///` per our coding standards. My comment is that the documentation is redundant because it just restates the function name and that it is not necessary. @andrsd can weigh in on his thoughts about comments like this.

wdhawkins: +1 for me on that. If the function of the method is obvious from its name, I don't think it needs a comment.

andrsd: I prefer to add comments even if they seem like repeating the obvious in the function name for one and only one reason: The comment will show up in the doxygen page and when looked at by user they will see some text rather then nothing. Which can seem like developers did not care enough to document API, or they can feel it is experimental API they should not be using. Potentially also search may find the words, users are looking for...

zhardy-lanl: I suppose that is true. That is a lot of redundant comments throughout the codebase though... I think that commenting the function name with spaces is as documented as no comment to be honest. I don't know how the doxygen search feature is built but I don't know if it is built like Google where I could just search something like "computing volumes per material ID" and get a result. I have always searched based off of class name or function name.

ragusa: I see value in having this referenced by doxygen. For all one-liners, I am converting this header to use `/// ...` I hope this is ok with everyone.

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: the reoson things look a bit convoluted in that routine is because a sub-domain may only have a few of the material IDs. So the `local_material_ids` are the ones we found in that partition, while  `all_material_ids` is the union of all the `local_material_ids`. The ids themselves do not changed values (contrarily to local_dof and global_dof), so i didn't want to give the local-vs-global flavor to the names.

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: it works with the updated mpi communicator. thanks for the suggestion.

andrsd (approved):

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: I have removed  `all_material_ids` and re-assigned it. should be good now

ragusa:
@wdhawkins @zhardy-lanl : the test machine never finished the regression suite. not sure why. I restarted the workflow and it passed. please provide approval/comment when ready.

andrsd:
[This](https://github.com/Open-Sn/opensn/actions/runs/12108999557) shows that the tests passed in about 10 mins. What makes you think it did not finish?

andrsd:
I see. I have seen the `run_tests` script fail weirdly. It happens to me once in a while, but I am not able to replicate that. So, it could be you got bit by that.

Just a language note: Attempt 1 *finished* but with an error message. "Never finished" to me means that the script would timed out on the github runner and was terminated. This did happened to me once or twice...

wdhawkins (approved):

PR #425, opened 2024-11-24
Creating LBSSolverVecOps class for group-scoped vector operations

wdhawkins:
This PR moves group-scoped vector operations out of LBSSolver and DiscreteOrdinatesSolver and into a standalone LBSSolverVecOps class.  It consolidates the base and derived vector ops methods into a single helper class.

I'm hoping this PR will generate some discussion on the future of LBSSolver. If we want LBSSolver to be a fancy container class that encapsulates problem parameters and solution data - with most of the heavy lifting delegated to helper classes -this PR fits in with that vision. Along those lines, I could see us splitting out other functionality like fission ops, etc. into helper classes.

If we want to retain the existing LBSSolver structure, I can easily replace the base and derived vector ops methods with the new, refactored versions.


andrsd:
The original code had polymorphism on these methods, so the behavior of LSBSolver and DiscreteOrdinatesSolver was different. In some methods, the pure LBS ignored the `DelayedPsi` while the DO used them. I guess the MG diffusion solver does not call these, so all works as expected?

andrsd (approved):

modules/linear_boltzmann_solvers/lbs_solver/vecops/lbs_solver_vecops.h
```diff
@@ -0,0 +1,106 @@
...
+#include <petscksp.h>
+#include <vector>
+
+namespace opensn
+{
+
+class LBSSolver;
+class LBSGroupset;
+
+class LBSSolverVecOps
```
zhardy-lanl, line 19:
```suggestion
class LBSVecOps
```
Just a suggestion

modules/linear_boltzmann_solvers/lbs_solver/vecops/lbs_solver_vecops.h
zhardy-lanl, line 1: I would not make a subdirectory for a single `.[h|cc]` file. Eventually we can maybe have a `helpers` directory that houses these types of things, but for now I don't think this is necessary.

zhardy-lanl (changes requested):

modules/linear_boltzmann_solvers/lbs_solver/vecops/lbs_solver_vecops.h
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/lbs_solver/vecops/lbs_solver_vecops.h
wdhawkins: Fixed.

zhardy-lanl (approved):

ragusa (approved):

PR #427, opened 2024-11-24
SCDSA-AAH + Multi-Sweep Fixes

### LINKED ISSUE #428: AAH for the one-sweep methods on the C5G7 benchmark
The AAH sweeper does not currently work with one-sweep methods on the C5G7 benchmark. The reason for this is not currently known. This issue dates back to ChiTech.

zhardy-lanl:
This PR changes the behavior of one-sweep methods on the C5G7 benchmark to use the CBC sweeper instead of AAH.

Closes #428


ragusa:
The changes to the cc file are just cosmetic, right? it's about renaming some vars. The really change is the switch to the methods called for in the input file, correct?

test/modules/linear_boltzmann_solvers/transport_keigen/c5g7/c5g7.lua
```diff
@@ -11,11 +11,13 @@ aquad.OptimizeForPolarSymmetry(pquad, 4.0 * math.pi)
 
 -- Solver
 if string.find(k_method, "scdsa") or string.find(k_method, "smm") then
-  inner_linear_method = "classic_richardson"
+  inner_linear_method = "petsc_richardson"
```
wdhawkins, line 14: Did classic Richardson not work with CBC?

zhardy-lanl: It should work with both I would imagine (as you saw). I will revert this back to classic Richardson tonight or tomorrow morning. As the methods are coded up, they allow for multi-sweep. The single sweep is just enforced in the input file. I think warnings should probably be printed for AAH/single-sweep settings.

andrsd (approved):

ragusa (approved):

wdhawkins:
I think there is still an issue here. I agree that the changes to `pi_keigen_scdsa.cc` in this PR and #403 work well with CBC, but I still have problems with the changes in #403 and AAH. Prior to #403, multi-sweep AAH + SCDSA would converge C5G7 to the correct answer with 32 ranks. #403 + mult-sweep AAH + SCDSA does not converge on 32 ranks. The changes to `pi_keigen_scdsa.cc` in this PR don't fix that issue.

zhardy-lanl:
> I think there is still an issue here. I agree that the changes to `pi_keigen_scdsa.cc` in this PR and #403 work well with CBC, but I still have problems with the changes in #403 and AAH. Prior to #403, multi-sweep AAH + SCDSA would converge C5G7 to the correct answer with 32 ranks. #403 + mult-sweep AAH + SCDSA does not converge on 32 ranks. The changes to `pi_keigen_scdsa.cc` in this PR don't fix that issue.

This is very strange. I will try to hunt this down. The changes in this PR are cosmetic. Do you see anything in #403 (or #402) for that matter that pops out as fishy?

wdhawkins:
> > I think there is still an issue here. I agree that the changes to `pi_keigen_scdsa.cc` in this PR and #403 work well with CBC, but I still have problems with the changes in #403 and AAH. Prior to #403, multi-sweep AAH + SCDSA would converge C5G7 to the correct answer with 32 ranks. #403 + mult-sweep AAH + SCDSA does not converge on 32 ranks. The changes to `pi_keigen_scdsa.cc` in this PR don't fix that issue.
> 
> This is very strange. I will try to hunt this down. The changes in this PR are cosmetic. Do you see anything in #403 (or #402) for that matter that pops out as fishy?

I'll take a look. I'm not very familiar with SCDSA, but I do know that reverting to the pre-403 version of that file works. I'll take a look at the code and see if anything jumps out.

zhardy-lanl:
**UPDATE**
We now know what was happening. SCDSA needs the difference between two successive iterations. When multiple sweeps were turned on, `phi0_star` was actually `num_sweeps` iterations in front of `phi0_ell`. The same behavior for AAH was observed with CBC. With that said here is a summary of the fix:

- Throw an error for one-sweep AAH in SCDSA and SMM with a descriptive message. This is a temporary fix we can revisit later if we deem it necessary.
- If more than one sweep is requested, decrease `max_iterations` by one and do an extra sweep after doing the solve to ensure the requested number of sweeps is what the user actually gets.

It seems that everything works as expected now. Thanks to @wdhawkins for going through the code in detail with me.

wdhawkins:
A quick note here.  Nothing that needs to be changed for this PR. We need to be able to query the DO solver for the percentage of lagged angular fluxes. We throw an error for AAH with 1 sweep right now, but we're assuming we have lagged fluxes. For an orthogonal grid with KBA that shouldn't be the case (not sure what Metis will do). Anyway, it's already on my list to add this as part of the revamp of iteration output. I'll create an issue to change this error check to check for lagged fluxes. I'll also add an orthogonal regression test that shouldn't need multiple sweeps with AAH.

zhardy-lanl:
> A quick note here. Nothing that needs to be changed for this PR. We need to be able to query the DO solver for the percentage of lagged angular fluxes. We throw an error for AAH with 1 sweep right now, but we're assuming we have lagged fluxes. For an orthogonal grid with KBA that shouldn't be the case (not sure what Metis will do). Anyway, it's already on my list to add this as part of the revamp of iteration output. I'll create an issue to change this error check to check for lagged fluxes. I'll also add an orthogonal regression test that shouldn't need multiple sweeps with AAH.

We should also allow allow for one-sweep AAH in serial

wdhawkins:
> > A quick note here. Nothing that needs to be changed for this PR. We need to be able to query the DO solver for the percentage of lagged angular fluxes. We throw an error for AAH with 1 sweep right now, but we're assuming we have lagged fluxes. For an orthogonal grid with KBA that shouldn't be the case (not sure what Metis will do). Anyway, it's already on my list to add this as part of the revamp of iteration output. I'll create an issue to change this error check to check for lagged fluxes. I'll also add an orthogonal regression test that shouldn't need multiple sweeps with AAH.
> 
> We should also allow for one-sweep AAH in serial

Oh... of course.  Good point!

wdhawkins (approved):

PR #432, opened 2024-11-28
Use C++ header files

ragusa:
Draft to get your opinion


zhardy-lanl (approved):

andrsd (approved):

wdhawkins:
I think `mathinc_def.h` is only used in a couple of places. It could probably go away all together.

ragusa:
I have followed Daryl's suggestion to remove the file altogether and to add the include's in the only header that was using `mathh_incdef.h`

PR #438, opened 2024-12-07
cleaning some files

ragusa:


framework/math/dense_matrix.h
```diff
@@ -30,7 +30,7 @@ class DenseMatrix : public NDArray<TYPE, 2>
    *
    * \param rows Number of rows
    * \param cols Number of columns
-   * \param intitial_value Value to initialize the matrix elements with
+   * \param inittial_value Value to initialize the matrix elements with
```
andrsd, line 33: "inittial_value" -> "initial_value"

framework/data_types/varying.h
```diff
@@ -151,7 +151,7 @@ class Varying
       }
     }
 
-    bool operator!=(const VaryingType& that) const override { return not(*this == that); }
+    bool operator!=(const VaryingType& that) const override { return *this != that; }
```
andrsd, line 154: clang reports this warning here:
```
warning: all paths through this function will call itself [-Winfinite-recursion]
    bool operator!=(const VaryingType& that) const override { return *this != that; }
```
that's why the original code uses `==` and negation.

framework/math/dense_matrix.h
```diff
@@ -415,7 +415,7 @@ GaussElimination(DenseMatrix<TYPE>& A, Vector<TYPE>& b, unsigned int n)
   }
 
   // Back substitution
-  for (int i = n - 1; i >= 0; --i)
+  for (auto i = n - 1; i >= 0; --i)
```
andrsd, line 418: For this for-loop to work, `i` must become negative (that's the stopping condition). However, the type is inferred from `n` which is `unsigned int`, so `i` will never be negative.

You have 3 options:
1. change type of `n`
2. keep the explicit `int i`
3. rewrite the loop so that `i` is only positive.

This is most likely what is causing the test suite to fail.

framework/object_factory.h
```diff
@@ -125,7 +125,7 @@ class ObjectFactory
   {
     auto& object_maker = GetInstance();
 
-    const std::string name = object_name;
+    const std::string& name = object_name;
```
andrsd, line 128: You might as well just remove the "alias" and use `object_name`

framework/object_factory.h
```diff
@@ -148,7 +148,7 @@ class ObjectFactory
...
```
andrsd, line 151: same here...

framework/object_factory.h
```diff
@@ -170,7 +170,7 @@ class ObjectFactory
   {
     auto& object_maker = GetInstance();
 
-    const std::string name = block_name;
+    const std::string& name = block_name;
```
andrsd, line 173: and here

framework/object.cc
```diff
@@ -12,9 +12,7 @@ Object::GetInputParameters()
   return {}; // Returns an empty block
 }
 
-Object::Object()
-{
-}
+Object::Object() = default;
```
andrsd, line 15: This is fine, but maybe even better is to move this into the header file.

andrsd (changes requested):

framework/runtime.cc
```diff
@@ -57,7 +54,7 @@ Initialize()
   SystemWideEventPublisher::GetInstance().PublishEvent(Event("ProgramStart"));
 
   // Disable internal HDF error reporting
-  H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
+  H5Eset_auto2(H5E_DEFAULT, nullptr, nullptr);
```
wdhawkins, line 57: We should be cautious here. We are using the `C` interface for `HDF5`, and `nullptr` is not always a drop-in replacement for `NULL`. While it's likely to work most of the time, there are specific instances where things can go awry. If we are using the `C` interface to a third-party library, we should probably stick with `C` constructs and use `NULL`.

framework/utils/hdf_utils.h
```diff
@@ -177,7 +177,7 @@ H5ReadDataset1D(hid_t id, const std::string& name)
     if (dataspace != H5I_INVALID_HID)
     {
       hsize_t dims[1];
-      if (H5Sget_simple_extent_dims(dataspace, dims, NULL) == 1)
+      if (H5Sget_simple_extent_dims(dataspace, dims, nullptr) == 1)
```
wdhawkins, line 180: Same as above with `NULL` vs. `nullptr`.

framework/math/dense_matrix.h
wdhawkins: I would suggest option (2). Checking against 0 is going to be more efficient than (3), and this code is particularly performance sensitive.

wdhawkins:
I'm going to close this PR since I think all of the issues have been addressed by our `clang-tidy` fixes.

PR #456, opened 2024-12-21
Refactor `CheckPointInsideCell` and fix for polyhedra

loganharbour:
https://github.com/Open-Sn/opensn/pull/261 broke `CheckPointInsideCell` for polyhedra. Before the refactoring, if a single tetrahedron was found that contains the point, it was considered contained (see [here](https://github.com/wdhawkins/openSn/blob/fb1e7b055d910577acc437c0c71acb34704b186b/framework/mesh/mesh_continuum/mesh_continuum.cc#L975) for old version). After the refactoring, the point needed to be contained within _all_ tetrahedra. This corrects that issue.

Impressively, this bug didn't break anything. This probably needs unit tests.

### DIFF
--- a/framework/mesh/mesh_continuum/mesh_continuum.cc
+++ b/framework/mesh/mesh_continuum/mesh_continuum.cc
@@ -381,45 +381,48 @@ MeshContinuum::CheckPointInsideCell(const Cell& cell, const Vector3& point) cons
 {
   const auto& grid_ref = *this;
 
+  // Check each cell edge. A point inside the cell will return a negative value. A point on either
+  // edge will return a zero value, and a point outside the cell will return a positive value.
   if (cell.GetType() == CellType::SLAB)
   {
     const auto& v0 = grid_ref.vertices[cell.vertex_ids[0]];
     const auto& v1 = grid_ref.vertices[cell.vertex_ids[1]];
-
-    // Check each cell edge. A point inside the cell will return a negative value. A point on either
-    // edge will return a zero value, and a point outside the cell will return a positive value.
-    if (((v0.z - point.z) * (v1.z - point.z)) > 0.0)
-      return false;
+    return (((v0.z - point.z) * (v1.z - point.z)) <= 0.0);
   }
-  else if (cell.GetType() == CellType::POLYGON)
+
+  // Check each face of the polygon. A point inside the face will give a negative value, a point
+  // on the face will give a zero value, and a point outside the face will give a positive value.
+  // If the point is inside all faces, it is inside the polygon.
+  if (cell.GetType() == CellType::POLYGON)
   {
-    // Check each face of the polygon. A point inside the face will give a negative value, a point
-    // on the face will give a zero value, and a point outside the face will give a positive value.
-    // If the point is inside all faces, it is inside the polygon.
     for (const auto& face : cell.faces)
     {
       const auto& vcp = point - face.centroid;
       if (vcp.Dot(face.normal) > 0.0)
         return false;
     }
+    return true;
   }
-  else if (cell.GetType() == CellType::POLYHEDRON)
+
+  // Check each tetrahedron within the polyhedron. If the point is contained within one
+  // of the tetrahedra, it is contained within the polyhedron.
+  if (cell.GetType() == CellType::POLYHEDRON)
   {
-    // Divide each polyhedron into tetrahedral sides. For each side, check if the point is inside
-    // the tetrahedron. If the point is inside all tetrahedral sides, it is inside the polyhedron.
-    auto InsideTet =
-      [](const Vector3& point, const Vector3& v0, const Vector3& v1, const Vector3& v2)
+    // Helper for returning whether or not the given point is considered "inside" a plane,
+    // where inside will mean on the interior of a tetrahedron (determined by the ordering
+    // of the vertices)
+    const auto InsidePlane = [&point](const std::array<Vector3, 3>& v)
     {
-      const auto& v01 = v1 - v0;
-      const auto& v02 = v2 - v0;
+      const auto v01 = v[1] - v[0];
+      const auto v02 = v[2] - v[0];
       const auto n = v01.Cross(v02).Normalized();
-      const auto c = (v0 + v1 + v2) / 3.0;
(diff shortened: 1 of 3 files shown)


framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -416,7 +416,7 @@ MeshContinuum::CheckPointInsideCell(const Cell& cell, const Vector3& point) cons
       const auto c = (v0 + v1 + v2) / 3.0;
       const auto pc = point - c;
 
-      if (pc.Dot(n) > 0.0)
+      if (pc.Dot(n) >= -1.e-12)
```
andrsd, line 419: should not this really be: `std::abs(pc.Dot(n)) >= 1e-12)`?

ragusa:
Wouldn't it also be better to check that the length of pc is greater than epsilon? If not, then point is also centroid and it is inside for convex polygons. If pc is greater, then fall back on standard test.

framework/mesh/mesh_continuum/mesh_continuum.cc
loganharbour: No, that would be everything _but_ zero

loganharbour:
> Wouldn't it also be better to check that the length of pc is greater than epsilon? If not, then point is also centroid and it is inside for convex polygons. If pc is greater, then fall back on standard test.

I did this in part.

Digging more, I found a bug. Updated the comment to detail the issue.

loganharbour:
Added exhaustive unit testing for `MeshContinuum::CheckPointInsideCell()`. Along with a basic unit testing framework (because something like gtest doesn't exist).

I think that testing like this should really be supported by something like gtest if possible. I'm glad to add that capability if desired, but I wanted to check opinions first.

loganharbour:
> I personally would like to see google test being used for all unit testing. It is easy to learn and quite powerful. One "disadvantage" is that it needs to be somehow integrated into the run_test script. I don't like the idea that developers has to run 2 things for testing.

I'd agree with this. Primarily because relying on a well-trusted external library will reduce maintenance burden. I wrote this quickly just so that I could reasonably throw together tests for a method that desperately needed them. The test itself would be very easily transitioned to gtest.

ragusa:
I agree that we should do this. To me unit testing will no only make the code more robust by revealing issues, it will also force us to decipher some of the code that no one on this team has looked at deeply enough to understand it. Correct me if I am wrong, but one cannot create a unit test if one does not fully understand a function. 

Rather than waiting for perfection, I suggest that what Logan has invested in be switched to gtest so that we have a begging of unit testing.

loganharbour:
Contains the content from https://github.com/Open-Sn/opensn/pull/464 (to add unit testing capability), which should be merged before this.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -381,45 +381,49 @@ MeshContinuum::CheckPointInsideCell(const Cell& cell, const Vector3& point) cons
...
+      const auto v02 = v[2] - v[0];
       const auto n = v01.Cross(v02).Normalized();
-      const auto c = (v0 + v1 + v2) / 3.0;
+      const auto c = (v[0] + v[1] + v[2]) / 3.0;
       const auto pc = point - c;
-
-      if (pc.Dot(n) > 0.0)
+      const auto pc_norm = pc.Norm();
+      // Point is at c
+      if (pc_norm < 1.e-12)
```
zhardy-lanl, line 423:
```suggestion
      // Point is at c
      if (pc.Norm() < 1.e-12)
```

framework/mesh/mesh_continuum/mesh_continuum.h
```diff
@@ -163,6 +163,11 @@ class MeshContinuum
   /// Compute volume per material id's
   void ComputeVolumePerMaterialID() const;
 
+  /// Get the face vertices of a tetrahedron contained within the given face and
+  /// side of a polyhedron.
```
zhardy-lanl, line 167: Please use `/** */` for multi-line comments, where the opening and closing should each be on their own line.

framework/mesh/mesh_vector.h
```diff
@@ -170,18 +170,6 @@ struct Vector3
...
-   * \f$ \vec{\vec{T}} = \vec{x} \otimes \vec{y} \f$
-   */
-  TensorRank2Dim3 OTimes(const Vector3& that) const;
-
-  /**
-   * Dot product of vector and a rank-2 tensor.
-   * \f$ \vec{w} = \vec{x} \bullet \vec{\vec{T}} \f$
-   */
-  Vector3 Dot(const TensorRank2Dim3& that) const;
-
```
zhardy-lanl, line 184: I assume these were just unused?

zhardy-lanl (changes requested):
I assume the intent is for #464 to go in first?

framework/mesh/mesh_vector.h
loganharbour: Yep. Declarations exited, but no definition

ragusa (approved):

doc/install_macos_easy.md
```diff
@@ -140,6 +140,7 @@ To check if the code compiled correctly execute the test scripts:
 ```shell
 cd /path/to/opensn
 test/run_tests -j<N>
+build/test/opensn-unit
```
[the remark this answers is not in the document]
loganharbour: They don't need to be. The linux instructions use cmake to run the tests, and that target has been updated to also run the unit tests

loganharbour:
This is good. Comments addressed and resolved.

ragusa (approved):

andrsd (changes requested):
Needs interactive rebase. The two "address review" patches should be changed to fixup patches and merged into appropriate patches.

loganharbour:
> Needs interactive rebase. The two "address review" patches should be changed to fixup patches and merged into appropriate patches.

Is this is part of the standard here?

I typically avoid doing this so what the changes made are explicit to reviewers.

andrsd:
> Is this is part of the standard here?

[Yes](https://open-sn.github.io/opensn/devguide/workflow.html#fixing-your-branch)

Issue #458, opened 2024-12-27
Circular dependency in lbs_vecops and LBSSolver


carlo-fiorina:
https://github.com/Open-Sn/opensn/blob/main/modules/linear_boltzmann_solvers/lbs_solver/lbs_vecops.h

There is a possible circular dependency here:
lbs_vecops.h contains (#include "modules/linear_boltzmann_solvers/lbs_solver/lbs_solver.h")
lbs_discrete_ordinates_solver.cc contains :#include "modules/linear_boltzmann_solvers/lbs_solver/lbs_vecops.h"
but lbs_discrete_ordinates_solver is derived from LBSSolver
 
#pragma once avoids recursive inclusions and compilation errors, but this is not ideal. #pragma once silently skips problematic re-inclusions.

**Possible solution?**
I guess lbs_vecops.h  should not include lbs_solver.h. I could not yet find out why it does that but it seems to me that vector operations should be independent of the existence of a solver that uses them.

wdhawkins:
That looks like a bug. I'll fix that. That include is likely the result of `lbs_vecops` being derived (not in a C++ sense) from an already existing class.  The entire `linear_boltzmann_solver` module is slated for a major refactoring in the next couple of months that will fix a lot of the issues that you're seeing.

ragusa:
closed by #462, #463

PR #461, opened 2025-01-06
Moving `GroupsetScopedCopy()` into `lbs_vecops.cc`

wdhawkins:
Moves `GroupsetScopedCopy()` into `lbs_vecops.cc` and removes `lbs_solver.h` from `lbs_vecops.h`. This resolves issue #458, but is a workaround ahead of refactoring `modules/linear_boltzmann_solvers`.


andrsd (approved):

ragusa (approved):
The issue of circular dependencies is larger than this PR and #458 only scratched the surface. You may close #458, but I will open a new issue for circular dependencies because we need to address this.

zhardy-lanl (approved):

PR #467, opened 2025-01-15
Correcting installation instructions

### LINKED ISSUE #459: Minor inconsistency in documentation
Quick install, step 4, says to:
$ export CMAKE_PREFIX_PATH=/path/to/dependencies/directory:$CMAKE_PREFIX_PATH`

However, /configure_dependencies.py, once completed, suggests to export CMAKE_PREFIX_PATH, but also PETSC_DIR and LD_LIBRARY_PATH

andrsd:
Closes #459

### DIFF
--- a/doc/install_linux.md
+++ b/doc/install_linux.md
@@ -178,7 +178,7 @@ libraries to your `CMAKE_PREFIX_PATH` environment variable. This can be
 accomplished with the following command:
 
 ```bash
-    $ export CMAKE_PREFIX_PATH=/path/to/dependencies/directory:$CMAKE_PREFIX_PATH`
+    $ export CMAKE_PREFIX_PATH=/path/to/dependencies${CMAKE_PREFIX_PATH:+:${CMAKE_PREFIX_PATH}}`
 ```
 
 **Important:** It may be a good idea to add the `CMAKE_PREFIX_PATH` variable to
--- a/doc/install_linux_easy.md
+++ b/doc/install_linux_easy.md
@@ -82,7 +82,7 @@ libraries to your `CMAKE_PREFIX_PATH` environment variable. This can be
 accomplished with the following command:
 
 ```bash
-$ export CMAKE_PREFIX_PATH=/path/to/dependencies/directory:$CMAKE_PREFIX_PATH`
+$ export CMAKE_PREFIX_PATH=/path/to/dependencies${CMAKE_PREFIX_PATH:+:${CMAKE_PREFIX_PATH}}`
 ```
 
 **Important:** It may be a good idea to add the `CMAKE_PREFIX_PATH` variable to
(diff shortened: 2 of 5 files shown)


andrsd:
On case you are not familiar with this syntax:
```
export A=value${A:+:${A}}
```
If `A` is already set, then `A` will be `value:${A}`. If `A` was not set, then `A` will be `value` (compare with `export A=value:${A}` where `A` would end up being `value:`)

wdhawkins (approved):

doc/install_linux.md
```diff
@@ -178,7 +178,7 @@ libraries to your `CMAKE_PREFIX_PATH` environment variable. This can be
 accomplished with the following command:
 
 ```bash
-    $ export CMAKE_PREFIX_PATH=/path/to/dependencies/directory:$CMAKE_PREFIX_PATH`
+    $ export CMAKE_PREFIX_PATH=/path/to/dependencies${CMAKE_PREFIX_PATH:+:${CMAKE_PREFIX_PATH}}`
```
zhardy-lanl, line 181: Could you explain this syntax here? In particular, the omission of the `/` after `dependencies` and the `:+:`.   I haven't seen this before and I would worry that this would confuse a user trying to install the code.

andrsd: - I removed the `/directory` part to be consistent with the rest of instructions (in the MD files and in the `configure_dependencies` script)
- The `:+:` is explained [here](https://github.com/Open-Sn/opensn/pull/467#issuecomment-2593812815)

ragusa: # Explanation of `export A=value${A:+:${A}}`

The command:

```bash
export A=value${A:+:${A}}
```
is a Bash shell expression that sets or updates the environment variable A based on its current value.

## Key Concepts
1. **export:**
   - Makes the variable `A` available to child processes of the current shell.
 2. **Parameter Expansion (```${...}```)**:
  -```${A}```: Expands to the current value of `A`.
  - ```${A:+something}```: A conditional expansion. If `A` is set (not null), it expands to something. Otherwise, it expands to nothing (an empty string).

## What This Command Does
1. **Base Value:**
  - ```A=value``` sets `A` to the string `value` initially.
2. **Conditional Append:**
  - The ```${A:+:${A}}``` part checks if `A` already has a value.
  - If `A` is set (even if empty), it appends :```${A}``` to the new value of `A`.

## Example Scenarios
**Case 1: `A` is not set**
Before running the command:
```bash
echo $A
# Output: (nothing, as A is unset)
```
After running the command:
```
bash
export A=value${A:+:${A}}
echo $A
# Output: value
```
Here, ```${A:+:${A}}``` expands to an empty string because `A` was unset.

**Case 2: A is set to something**
Before running the command:
```bash
export A=foo
```
After running the command:
```
bash
export A=value${A:+:${A}}
echo $A
# Output: value:foo
```
Here, ```${A:+:${A}}``` expands to `:foo` because `A` was already set to `foo`.

**Case 3: A is set but empty**
Before running the command:
```bash
export A=
```
After running the command:
```
bash
export A=value${A:+:${A}}
echo $A
# Output: value:
```
```${A:+:${A}}``` expands to `:` because `A` was set, even though it was empty.

ragusa: not me, chatGPT

zhardy-lanl (approved):

PR #469, opened 2025-01-16
Refactoring restart capability

wdhawkins:
This PR refactors our restart capability and includes:

1. Restart routines have been moved out of `lbs_solver` and into the appropriate executors.
2. Steady-state problems only write a restart once the problem has converged.
3. Restart writes are now controlled with the `restart_writes_enabled` option. `write_restart_time_interval` is now optional and does not apply to steady-state problems.
4. Lagged angular fluxes are now saved to restart dumps by default. They are only included if necessary, and this behavior can be controlled with the `write_delayed_psi_to_restart` option.
5. The API for `H5ReadDataset1D` has changed to match the other H5 methods. `H5ReadDataset1D` now takes a reference to an `std:vector` and returns a bool indicating success or failure.

I am using `c5g7_restart.lua` as the test problem for the delayed angular flux capability. The restart files generated by this problem could be considered large-ish (~10MB * 4). If we think that's too large for the repository, I'll see about finding a problem that generates smaller restart files.


framework/utils/hdf_utils.h
```diff
@@ -184,6 +184,7 @@ H5ReadDataset1D(hid_t id, const std::string& name)
         {
           data.clear();
           data.shrink_to_fit();
+          retval = false;
```
ragusa, line 187: i guess this is a short name for returned value. Can we use a longer, more descriptive name?

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.lua
```diff
@@ -0,0 +1,144 @@
...
+--    fieldfunc.SetProperty(slices[k],SLICE_POINT,{x = 0.0, y = 0.0, z = 0.8001})
+--    fieldfunc.SetProperty(slices[k],ADD_FIELDFUNCTION,fflist[k])
+--    --fieldfunc.SetProperty(slices[k],SLICE_TANGENT,{x = 0.393, y = 1.0-0.393, z = 0})
+--    --fieldfunc.SetProperty(slices[k],SLICE_NORMAL,{x = -(1.0-0.393), y = -0.393, z = 0.0})
+--    --fieldfunc.SetProperty(slices[k],SLICE_BINORM,{x = 0.0, y = 0.0, z = 1.0})
+--    fieldfunc.Initialize(slices[k])
+--    fieldfunc.Execute(slices[k])
+--    fieldfunc.ExportToPython(slices[k])
+--end
+
```
ragusa, line 108: can be deleted

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.lua
```diff
@@ -0,0 +1,144 @@
...
+      gmres_restart_interval = 100,
+    },
+  },
+}
+
+lbs_options = {
+  scattering_order = 0,
+  --restart_writes_enabled = true,
+  --write_delayed_psi_to_restart = true,
+  --write_restart_path = "transport_3d_5_cycles_2_restart/transport_3d_5_cycles_2",
```
ragusa, line 79: delete commented out lines

ragusa (changes requested):
some quick changes.

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.lua
wdhawkins: Fixed.

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.lua
wdhawkins: I think these are worth leaving. They show how to generate the restart data for this test.

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.lua
```diff
@@ -0,0 +1,130 @@
+-- 3D Transport test with Vacuum BCs.
+-- SDM: PWLD
+-- Test: Max-value1=6.55387e+00
+--       Max-value2=1.02940e+00
+
+num_procs = 4
+
+--############################################### Check num_procs
```
andrsd, line 8: Could you remove the `#`s from all these comments?

andrsd (changes requested):

test/modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.lua
wdhawkins: Fixed.

framework/utils/hdf_utils.h
zhardy-lanl: Alternatively, this could be named `success` or something along those lines

modules/linear_boltzmann_solvers/executors/lbs_steady_state.cc
```diff
@@ -45,16 +46,114 @@ SteadyStateSolver::Execute()
...
 
+  auto& options = lbs_solver_.Options();
+
+  if (not options.read_restart_path.empty())
+  {
+    if (ReadRestartData())
+      log.Log() << "Successfully read restart data." << std::endl;
+    else
+      log.Log() << "Failed to read restart data." << std::endl;
+  }
```
zhardy-lanl, line 57: This is fine, but I thought I would ask. Would you consider reading in restart data part of initialization or execution? I lean towards initialization. If you do as well, consider moving this to `SteadyStateSolver::Initialize`.

Additionally, maybe we could hide these messages inside of `ReadRestartData`.

modules/linear_boltzmann_solvers/executors/lbs_steady_state.cc
```diff
@@ -45,16 +46,114 @@ SteadyStateSolver::Execute()
...
-  if (lbs_solver_.Options().use_precursors)
+  if (options.restart_writes_enabled)
+  {
+    if (WriteRestartData())
+    {
+      lbs_solver_.UpdateRestartWriteTime();
+      log.Log() << "Successfully wrote restart data." << std::endl;
+    }
+    else
+      log.Log() << "Failed to write restart data." << std::endl;
```
zhardy-lanl, line 70: Could this be hidden inside of `WriteRestartData` to make this more clean?

zhardy-lanl, line 70: Could this be hidden inside of `WriteRestartData` to make this more clean?

modules/linear_boltzmann_solvers/executors/lbs_steady_state.cc
wdhawkins: Definitely not opposed to moving it to initialization.  My preference is to return success or failure and let the developer handle the condition appropriately. Since all of these methods are local to the solver, I don't see any reason not to move the output to the specific restart routines.

modules/linear_boltzmann_solvers/executors/lbs_steady_state.cc
zhardy-lanl: Answered with response to other comment.

andrsd (approved):

framework/utils/hdf_utils.h
wdhawkins: Fixed.

ragusa (approved):

zhardy-lanl (approved):

PR #475, opened 2025-01-29
Removing mesh stack

andrsd:
Solvers now accept `mesh` parameter instead of grabbing a mesh object from the mesh stack.


ragusa (approved):

ragusa:
but a lot of QA is failing currently.

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -579,10 +579,14 @@ MeshContinuum::SetUniformMaterialID(int mat_id)
   const auto& ghost_ids = cells.GetGhostGlobalIDs();
   for (uint64_t ghost_id : ghost_ids)
     cells[ghost_id].material_id = mat_id;
+
+  mpi_comm.barrier();
+  opensn::log.Log() << program_timer.GetTimeString() << " Done setting material id " << mat_id
+                    << " to all cells";
```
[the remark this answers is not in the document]
andrsd: this could be something that crept in from a rebase. let me check...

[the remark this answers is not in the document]
andrsd: The diff does not show the reason for this change.  However, it is because in the lua layer, there was a function:
```
void
MeshSetUniformMaterialID(int mat_id)
{
 auto mesh = GetCurrentMesh();
 mesh->SetUniformMaterialID(mat_id);
 mpi_comm.barrier();
 opensn::log.Log() << program_timer.GetTimeString() << " Done setting material id " << mat_id
                   << " to all cells";
}
```

That is now gone, because I bound `MeshContinuum::SetUniformMaterialID` directly. So, to preserve the functionality, the printout statement with the barrier was moved inside `MeshContinuum::SetUniformMaterialID`.

So, not a bug...

andrsd:
> but a lot of QA is failing currently.

That's resolved now

lua/lib/mesh.h
```diff
@@ -4,21 +4,15 @@
...
-void MeshSetupOrthogonalBoundaries();
-void MeshSetMaterialIDFromFunction(const char* lua_fname);
-void MeshSetBoundaryIDFromFunction(const char* lua_fname);
-void MeshExportToPVTU(const std::string& file_name);
-void MeshComputeVolumePerMaterialID();
+void MeshSetMaterialIDFromFunction(std::shared_ptr<opensn::MeshContinuum> grid,
+                                   const char* lua_fname);
+void MeshSetBoundaryIDFromFunction(std::shared_ptr<opensn::MeshContinuum> grid,
+                                   const char* lua_fname);
+void MeshExportToPVTU(std::shared_ptr<opensn::MeshContinuum> grid, const std::string& file_name);
```
loganharbour, line 16: You don't need the smart pointers here (within the definition), but they're required to be passed as such as a result of the lua bindings, correct?

loganharbour (approved):

lua/lib/mesh.h
loganharbour: After looking at the input, this can be disregarded. Answered my own question in the export calls.

loganharbour:
Sorry for the need to rebase 😬

lua/lib/mesh.h
andrsd: Yes, objects from Lua arrive usually as `shared_ptr` and `MeshIO::ToPVTU` expects a `shared_ptr`.
But also [this](https://open-sn.github.io/opensn/devguide/coding_standard.html#pointers).

Issue #124, opened 2024-02-14
Fix curvilinear sweeper with triangular elements


zhardy-lanl:
The curvilinear sweeper breaks (segfault failure if my memory is correct) when using triangular elements. The cause is not currently understood.

PR #481, opened 2025-02-12
Initial refactoring of angular quadrature classes

wdhawkins:
This PR begins the process of refactoring the angular quadrature classes and makes the following changes:

1. Renames product quadrature routines (including user-facing names). The new naming scheme is QuadratureType + Dimension + Geometry, so we now have names similar to the following:
`GLProductQuadrature1DSlab` -> Gauss-Legendre product quadrature, 1D, slab geometry
`GLCProductQuadrature2DRZ` -> Gauss-Legendre Chebyshev product quadrature, 2D, RZ geometry
`GLCProductQuadrature3DXYZ` -> Gauss-Legendre Chebyshev product quadrature, 3D, XYZ geometry
2. Removes the custom quadrature capability.
3. Standardizes the quadrature calling convention.
4. When specifying `Npolar` the user now supplies the total number of polar angles, and, when specifying `Nazimuthal`, the user specifies the total number of angles in 4 quadrants. `Npolar` is checked to make sure it is divisible by 2, and `Nazimuthal` is checked to make sure it is divisible by 4.
5. `OptimizaForPolarSymmetry()` is no longer a user-facing function and is handled automatically for 2D quadratures
6. The dimension of the quadrature is now stored and checked against the dimension of the mesh.


framework/math/quadratures/angular/product_quadrature.cc
```diff
@@ -164,49 +161,70 @@ AngularQuadratureProdGL::AngularQuadratureProdGL(int Nphemi, bool verbose) : Pro
...
-  for (unsigned int i = 0; i < azimu_ang.size(); ++i)
-    for (unsigned int j = 0; j < polar_ang.size(); ++j)
+  for (auto i = 0; i < azimu_ang.size(); ++i)
+    for (auto j = 0; j < polar_ang.size(); ++j)
       weights.emplace_back(2 * gc_azimu.weights[i] * gl_polar.weights[j]);
 
   // Initialize
   AssembleCosines(azimu_ang, polar_ang, weights, verbose);
+
+  OptimizeForPolarSymmetry(4.0 * M_PI);
```
ragusa, line 195: what is the 4pi for? Don't we want, by default to normlaize to sum of weights = 1 no matter what?

test/modules/linear_boltzmann_solvers/transport_steady_cyl/transport_2d_cyl_2_multigroup.lua
```diff
@@ -60,7 +60,7 @@ mg_src = xs.IsotropicMultiGroupSource.FromArray(source)
 material0:SetIsotropicMGSource(mg_src)
 
 -- Setup Physics
-pquad0 = aquad.CreateCylindricalProductQuadrature(GAUSS_LEGENDRE_CHEBYSHEV, 4, 8)
```
ragusa, line 63: in RZ, why are the integers passed in arguments the same before and after your changes? I thought that the number of polar angles was, previously, the number of positive angles and now it is the total number of levels. I was expecting a factor 2. Maybe a similar question for the second entry as well.

framework/math/quadratures/angular/product_quadrature.cc
wdhawkins: For this commit, I left the default as it was. There are so many other changes that I was purposefully trying to not have to modify any of the outputs. I have a second PR after this one is merged that makes that change since it's likely to affect the results of some tests.

test/modules/linear_boltzmann_solvers/transport_steady_cyl/transport_2d_cyl_2_multigroup.lua
wdhawkins: They shouldn't be. I just accidentally missed them in my mass editing of the inputs. I've fixed those inputs. Nice catch!

framework/math/quadratures/angular/product_quadrature.cc
ragusa: we have to be careful. a test that has a scalar flux as qoi will NOT change its value. so, if in the next PR, you need to re-gold a results, we have to scrutinize that

test/modules/linear_boltzmann_solvers/transport_steady_cyl/transport_2d_cyl_2_multigroup.lua
ragusa: i didn't scrutinize every input file. should i do that or will the regression test catch the issues? it's a pain to have to go through all those input files, for you as author of the PR, for me as reviewer. if only the original developer talked to experts to learn what the convention in the field was ...

ragusa: I have a problem with this. none of the 3 tests that failed is this input file. so the regression suite didn't not catch that input change because the results was unaffected. that's a bad test!

wdhawkins: It is a pain. And I went through them twice and still missed some. I modified them a test directory at a time, so that's likely why I missed these. I'll go back through all of them a third time just to make sure I got them all... and make the correct changes.

wdhawkins: > I have a problem with this. none of the 3 tests that failed is this input file. so the regression suite didn't not catch that input change because the results was unaffected. that's a bad test!

I agree. We have #124 open to address this.

framework/math/quadratures/angular/curvilinear_product_quadrature.cc
```diff
@@ -0,0 +1,417 @@
...
+                << ", omega = " << omegas[k].PrintStr()
+                << ", fac_diamond_difference = " << fac_diamond_difference_[k]
+                << ", fac_streaming_operator = " << fac_streaming_operator_[k] << std::endl;
+    const auto sum_weights = std::accumulate(weights.begin(), weights.end(), 0.0);
+    log.Log() << "sum(weights) = " << sum_weights << std::endl;
+  }
+}
+
+void
+GLCProductQuadrature2DRZ ::InitializeParameters()
```
andrsd, line 367: Interesting that the space between `GLCProductQuadrature2DRZ` and `::` was not found by clang-format check.

framework/math/quadratures/angular/product_quadrature.h
```diff
@@ -67,28 +61,25 @@ class ProductQuadrature : public AngularQuadrature
   }
 };
 
-class AngularQuadratureProdGL : public ProductQuadrature
+class GLProductQuadrature1DSlab : public ProductQuadrature
 {
 public:
-  /// Constructor for Angular Gauss-Legendre.
-  explicit AngularQuadratureProdGL(int Np, bool verbose = false);
+  /// Constructor for 1D slab Gauss-Legendre product qaudrature
```
andrsd, line 67: "qaudrature" -> "quadrature"

andrsd (changes requested):

framework/math/quadratures/angular/curvilinear_product_quadrature.cc
wdhawkins: I'm surprised by that too!  Fixed.

framework/math/quadratures/angular/product_quadrature.h
wdhawkins: Good catch.  Fixed.

andrsd (approved):

ragusa (approved):

PR #487, opened 2025-02-13
Don't append extensions to empty restart file names

wdhawkins:
This PR fixes a small issue where we incorrectly appended extensions to empty restart file names. This resulted in attempts to read restart files that did not exist.


ragusa (approved):

ragusa:
we do not need 3 reviews for this.

andrsd (approved):

PR #494, opened 2025-02-18
Simplifications to SweepScheduler

wdhawkins:
Simplifies scheduling rules and removes unnecessary calls to AngleSetAdvance. Tested at scale.


ragusa (approved):
We need more like this. Shaving off little bit little.

andrsd (approved):

PR #496, opened 2025-02-19
Removing `Material` concept from the code

### LINKED ISSUE #431: Remove materials, and only use cross section objects
Once #430 is closed, materials will only be responsible for storing `MultiGroupXS` objects, which makes them unnecessary. Because `LBSSolver` operates on `matid_to_xs_map_` (not materials), this should be set directly in options. Within the `LBSOptions` there should be an option called something like `xs` or `xs_map` that has the syntax
```
xs_map = {<block1> = xs[1], <block2> = xs[2], ...}
```
that ultimately populates the `matid_to_xs_map_` attribute.

A separate API call like 
```
lbs.AssignXSToBlocks(phys, xs_map)
```
could and probably should be implemented.

andrsd:
Remove `Material` from the code.
Added `xs_map` instead as an option for the solver.

Closes #431


wdhawkins:
I'll have to test this PR. I'm a bit concerned that it doesn't look like there is any concept now of materials in the visualization output.  This is important since we need to be able to do things like subtract a material from the problem to be able to see inside it. For example, if I have a cube of air with a a sphere on the inside, it's useful to be able to subtract the air to visualize the sphere. Does this PR prevent that?

andrsd:
There should still be `block_id`s and you should be able to achieve what you describe above with that, I think. It can be that after this PR, the IDs are not written into the output files. I should check that as well...

ragusa:
I looked at the lines that were removed that Daryl might have questioned. They were for vtk to obj. I am not even sure if we use this functionality at all and why we even have it. However, it would be necessary to check that the mesh, when exported, is still showing blovk_id (instead of, formerly, material)

ragusa (approved):

wdhawkins:
I've tested visualization with this PR, and everything looks good.

wdhawkins (approved):

Issue #471, opened 2025-01-23
Simplify LBSSolver class


wdhawkins:
The following issues need to be addresses as part of the refactoring of LBSSolver:

- [ ] Implement generic boundary objects
- [ ] Boundary conditions and sources should be passed to the solver via the `lbs_solver` block in the input, not the `lbs_options` block
- [ ] Generic groupset objects
- [ ] Convert flux moment vectors to `std::vector<std::vector<>>` to eliminate complex copy operations (LBSVecOps)
- [ ] Move grid-level information to more appropriate locations (grid nodal mappings, grid face histograms, etc.)
- [ ] Simplify cell-level discretization mappings (unit cell matrices, cell mappings, cell lbsview)

andrsd:
`lbs_curvilinear_solver.h` contains `DiscreteOrdinatesCurvilinearSolver`. I am renaming it to `DiscreteOrdinatesCurvilinearProblem`. Do you want the file be `discrete_ordinates_curvilinear_solver.{h|cc}`?

andrsd:
Also, `SweepChunkPwlrz` is in `lbs_curvilinear_sweep_chunk_pwl.h`. Do you want me to rename the file to something like `curvilinear_sweep_chunk_pwl.h` and rename the class to match the file or match the file name to the class name? Thoughts?

PR #500, opened 2025-02-24
Moved volumes and areas to cells and faces

zhardy-lanl:
Moved cell volumes and face areas to the respective `Cell` and `CellFace` objects from `CellMapping`. Prior to this PR, face normals lived in `CellFace` and were computed in the `MeshGenerator::SetupCell` and cell volumes and areas lived in `CellMapping` were computed in `CellMapping::ComputeCellVolumeAndAreas`. Because the routines used to compute these geometric quantities only used mesh-level vertex data, it was not necessary to house these separately. All routines to compute this data has been moved under `ComputeGeometricInfo` within `MeshContinuum`, `Cell`, and `CellFace`. All data gets initialized in the `MeshGenerator` objects after cell and face definitions are completed.

Refs #471

### DIFF
--- a/framework/mesh/cell/cell.cc
+++ b/framework/mesh/cell/cell.cc
@@ -319,10 +319,12 @@ Cell::ComputeGeometricInfo(const MeshContinuum* grid)
       {
         const auto& v0 = grid->vertices[face.vertex_ids[0]];
         const auto& v1 = grid->vertices[face.vertex_ids[1]];
-        const auto subnormal = (v0 - centroid).Cross(v1 - centroid);
-        volume += 0.5 * subnormal.Norm();
-        break;
+
+        const auto e0 = v1 - v0;
+        const auto e1 = centroid - v0;
+        volume += 0.5 * std::fabs(e0.x * e1.y - e0.y * e1.x);
       }
+      break;
     }
 
     // The volume of a polyhedron is the sum of the sub-tetrahedrons
--- a/framework/mesh/cell/cell.cc
+++ b/framework/mesh/cell/cell.cc
@@ -43,45 +43,6 @@ CellTypeName(const CellType type)
       return "NONE";
   }
 }
-Cell::Cell(const Cell& other)
-  : cell_type_(other.cell_type_),
-    cell_sub_type_(other.cell_sub_type_),
-    global_id(other.global_id),
-    local_id(other.local_id),
-    partition_id(other.partition_id),
-    centroid(other.centroid),
-    material_id(other.material_id),
-    vertex_ids(other.vertex_ids),
-    faces(other.faces)
-{
-}
-
-Cell::Cell(Cell&& other) noexcept
-  : cell_type_(other.cell_type_),
-    cell_sub_type_(other.cell_sub_type_),
-    global_id(other.global_id),
-    local_id(other.local_id),
-    partition_id(other.partition_id),
-    centroid(other.centroid),
-    material_id(other.material_id),
-    vertex_ids(std::move(other.vertex_ids)),
-    faces(std::move(other.faces))
-{
-}
-
-Cell&
-Cell::operator=(const Cell& other)
-{
-  global_id = other.global_id;
-  local_id = other.local_id;
-  partition_id = other.partition_id;
-  centroid = other.centroid;
-  material_id = other.material_id;
-  vertex_ids = other.vertex_ids;
-  faces = other.faces;
-
-  return *this;
-}
 
 bool
 CellFace::IsNeighborLocal(const MeshContinuum* grid) const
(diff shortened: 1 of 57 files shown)


framework/mesh/cell/cell.cc
```diff
@@ -179,43 +140,62 @@ CellFace::GetNeighborAdjacentFaceIndex(const MeshContinuum* grid) const
...
-    double area = 0.0;
-    auto& v2 = centroid;
+    // The face of a polyhedron is a polygon. The area can be computed
+    // by summing the volume of triangles formed with each edge and
+    // the centroid. The normal must be computed as the area-weighted
+    // average of the normal vector on each sub-triangle.
+    area = 0.0;
+    normal = Vector3(0.0, 0.0, 0.0);
     const auto num_verts = vertex_ids.size();
     for (uint64_t v = 0; v < num_verts; ++v)
```
ragusa, line 184: would `auto` work here?

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -19,109 +19,140 @@ namespace opensn
...
+        else if (n.Dot(ihat) > 0.999)
+          boundary_name = "XMAX";
+        else if (n.Dot(jhat) < -0.999)
+          boundary_name = "YMIN";
+        else if (n.Dot(jhat) > 0.999)
+          boundary_name = "YMAX";
+        else if (n.Dot(khat) < -0.999)
+          boundary_name = "ZMIN";
+        else if (n.Dot(khat) > 0.999)
+          boundary_name = "ZMAX";
```
ragusa, line 143: I was wondering if using three nine's was enough. it feels a bit loose (it was in the code before, but it is time to tighten this a bit?)

ragusa (approved):

framework/mesh/cell/cell.cc
zhardy-lanl: Would auto work for what in particular?

framework/mesh/mesh_continuum/mesh_continuum.cc
zhardy-lanl: That's a good point. I don't know how many nines are really appropriate, but I'm happy to up them if you'd like. Just let me know what you'd like to see.

framework/mesh/cell/cell.cc
ragusa: Line 184. The loop index.

zhardy-lanl: Ah, yes. That can be made `auto`. For the counting loops (i.e. not range-based), I'm not sure how I feel about using auto. We generally don't throughout the rest of the code base. We should discuss adopting a standard on this. I will leave this until we've had that discussion.

zhardy-lanl: My vote is to stick with the type you want. It is the simple solution that doesn't require any modification. Plus, a newer developer will likely get confused when seeing `make_range` all over the place. I know I would have when I first started working on the code.

ragusa: the reason for my comment was that there were several locations where the type of the index was changed to auto in this PR. For example, see screenshot, but there are other occurrences.  
image
For consistency, maybe the line I pointed out could be changed too.

For the larger picture, I think auto should be used (but not with make_range). The situations were auto cannot be used (as when the type is an unsigned int but the loop is decreasing, so -1 must be reached, would be an example of where auto cannot be used. However, the fact that auto is not employed should also be telling the developer something about that loop, and i like that feeling (if all other loops are changed to auto, of course).

zhardy-lanl: This particular example is of a range-based loop whereas the comment you left was on a counting loop. I think we should always use `auto` on range based loops, and from @andrsd comment, likely use actual types for counting loops.

framework/mesh/cell/cell.cc
```diff
@@ -283,13 +263,96 @@ CellFace::ToString() const
...
+  // Compute cell centroid
+  centroid = Vector3(0.0, 0.0, 0.0);
+  for (const auto& vid : vertex_ids)
     centroid += grid->vertices[vid];
   centroid /= static_cast<double>(vertex_ids.size());
+
+  // Compute face geometric data
+  unsigned int f = 0;
+  for (auto& face : faces)
+    face.ComputeGeometricInfo(grid, f++);
```
andrsd, line 300: Probably `++f` instead of `f++`?
It does not make a lot of difference here, but I think consistency is required with the increment operator...

framework/mesh/cell/cell.h
```diff
@@ -67,9 +55,8 @@ class CellFace
   /// Determines the neighbor's associated face.
   int GetNeighborAdjacentFaceIndex(const MeshContinuum* grid) const;
 
-public:
-  /// Computes the face area.
-  double ComputeFaceArea(const MeshContinuum* grid) const;
+  /// Computes the geometric info on the face.
+  void ComputeGeometricInfo(const MeshContinuum* grid, unsigned int f = 0);
```
andrsd, line 59: You most likely do not want a default value on `f`

framework/mesh/cell/cell.h
```diff
@@ -80,50 +67,41 @@ class CellFace
...
-{
-private:
-  /// Primary type, i.e. SLAB, POLYGON, POLYHEDRON
-  const CellType cell_type_;
-  /// Sub-type i.e. SLAB, QUADRILATERAL, HEXAHEDRON
-  const CellType cell_sub_type_;
+  /// Flag indicating whether face has a neighbor
+  bool has_neighbor = false;
+  /// If face has neighbor, contains the global_id, otherwise, contains boundary_id.
+  uint64_t neighbor_id = 0;
```
andrsd, line 73: This is just a discussion point. I think at some point we talked about not doing this (i.e. initialize member variables in a header file) and having all inits in the ctor initializer list. I don't think this was standardized, so I am bringing this up, if it should be standardized or we keep doing this...

framework/mesh/mesh_generator/mesh_generator.cc
```diff
@@ -347,72 +210,141 @@ MeshGenerator::SetupCell(const UnpartitionedMesh::LightWeightCell& raw_cell,
...
+      if (min_partition == partition or cells_per_partition[min_partition] >= target)
+        break;
+
+      for (auto& cell_pid : cell_pids)
+      {
+        if (cell_pid == partition)
+        {
+          cell_pid = min_partition;
+          cells_per_partition[partition]--;
+          cells_per_partition[min_partition]++;
```
andrsd, line 342: Should this use prefix operators?

framework/mesh/mesh_generator/mesh_generator.cc
```diff
@@ -347,72 +210,141 @@ MeshGenerator::SetupCell(const UnpartitionedMesh::LightWeightCell& raw_cell,
...
+  communicator.broadcast(cell_pids, root);
+}
+
+void
+MeshGenerator::RebalancePartitions(std::vector<int64_t>& cell_pids, const int num_partitions)
+{
+  // Count the number of cells in each partition
+  std::vector<int> cells_per_partition(num_partitions, 0);
+  for (const auto partition : cell_pids)
+    cells_per_partition[partition]++;
```
andrsd, line 313: Should this use prefix operators?

modules/diffusion/fv_diffusion_solver.cc
```diff
@@ -172,7 +172,7 @@ FVDiffusionSolver::Initialize()
   log.Log() << "Global num cells: " << grid.GetGlobalNumberOfCells();
 
   // BIDs
-  auto globl_unique_bndry_ids = grid.GetDomainUniqueBoundaryIDs();
+  auto globl_unique_bndry_ids = grid.GetUniqueBoundaryIDs();
```
andrsd, line 175: Maybe change `globl` to `global`?

modules/diffusion/mg_diffusion_solver.cc
```diff
@@ -300,7 +300,7 @@ MGDiffusionSolver::Initialize()
   InitializeMaterials(unique_material_ids);
 
   // BIDs
-  auto globl_unique_bndry_ids = grid.GetDomainUniqueBoundaryIDs();
+  auto globl_unique_bndry_ids = grid.GetUniqueBoundaryIDs();
```
andrsd, line 303: Here as well...

andrsd (changes requested):

framework/mesh/cell/cell.h
zhardy-lanl: The reason this is here is because the face index is only used for 1D faces. I can make this a required parameter, though. I thought it may be better to have the required call signature be the 2D/3D case rather than 1D, but can change that. Thoughts?

framework/mesh/cell/cell.h
zhardy-lanl: I'm a little agnostic to this I think. I see a value in being able to go into a class and see what the attributes and defaults are in one place rather than having to look through a declaration and then a constructor.

framework/mesh/mesh_generator/mesh_generator.cc
zhardy-lanl: Probably so. This was just moved around not changed. I'll make it prefix.

framework/mesh/cell/cell.h
zhardy-lanl: Since I pass the face index anyways it looks like, I will remove the default value

framework/mesh/cell/cell.cc
zhardy-lanl: It should be `f++` here. I don't want to pass face index 1 when I am in fact on face index 0. I can alternatively use `int face = -1` and use a prefix, but I am not a huge fan of that

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: maybe 0.99999 (five 9's)

framework/mesh/cell/cell.cc
andrsd: @zhardy-lanl I see, my bad. This where `enumerate` would be quite handy :-) I am fine with the post-fix increment. Starting from `-1` would be confusing...

andrsd:
Thanks for fixing the `globl` to `global` 👍

andrsd (approved):

PR #501, opened 2025-02-25
Implement the Python interface for the first LBS example

quocdang1998:
Implementation of the Python interface of OpenSn for the first example. To run it, execute:

python setup.py build_ext --inplace    # for compiling
pushd pushd tutorials/lbs/first/       # go to the .py directory
mpiexec -n 4 python3 first_example.py  # execute
popd                                   # return to the original directory

User can install the Python interface to the Python library directory with pip, using:

pip install .

Note: only the first example works for now. Further development will be required in the future!


pyopensn/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+    src/fieldfunc.cpp
+    src/logvol.cpp
+    src/main.cpp
+    src/mat.cpp
+    src/mesh.cpp
+    src/parameter.cpp
+    src/solver.cpp
+    src/source.cpp
+    src/xs.cpp
+)
```
andrsd, line 39: You may want to switch this to `file(GLOB_RECURSE <VAR> CONFIGURE_DEPENDS...`. It is what the main CMakeLists.txt uses. People usually forget to add a new file into CMake and then they wonder what is going on. That's why we prefer the `GLOB`

pyopensn/src/aquad.cpp
```diff
@@ -0,0 +1,132 @@
+/*
+ * Created on Sun, February 23
+ *
+ * Copyright (c) 2025 The OpenSn Authors <https://open-sn.github.io/opensn/>
+ */
```
andrsd, line 5: These copyright notices should be using SPDX format, like [so](https://github.com/Open-Sn/opensn/blob/8dc54ff130546259027095aed84160b9e3aa03bc/framework/runtime.h#L1-L2)

Also, please, update `.licenserc.json` to check inside `pyopensn`

pyopensn/src/aquad.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include "pyapi.hpp"
+
+#include <memory>     // std::shared_ptr, std::make_shared
```
andrsd, line 9: Do not do the comment which explains why you `#include` a header file. We don't do that anywhere, so keep the code consistent with the rest of the code base.

pyopensn/src/pyapi.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include "framework/parameters/parameter_block.h"   // opensn::ParameterBlock
+#include "framework/parameters/input_parameters.h"  // opensn::InputParameters
+
+namespace opensn {
+
+// Environment for running Python
+// ------------------------------
+
+/** @brief Environment initializer and finalizer for OpenSn.*/
```
andrsd, line 25: The coding standards says:
- single line doco-comment use `///`
- multi-line use `/** */`

[See](https://open-sn.github.io/opensn/devguide/coding_standard.html#comments)

tutorials/lbs/first/first_example.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+    )
+    grid = meshgen.Execute()
+
+    # set material
+    grid.SetUniformMaterialID(0)
+
+    # add material
+    materials = []
+    materials.append(AddMaterial("Material_A"))
```
andrsd, line 51: Just headsup: PR #496 will drop the material concept.

tutorials/lbs/first/first_example.py
```diff
@@ -0,0 +1,101 @@
...
+    strength = []
+    for g in range(num_groups):
+        strength.append(1.0)
+    mg_src = VolumetricSource(block_ids = [0], group_strength = strength)
+
+    # initialize quadrature
+    nazimu = 1
+    npolar = 2
+    pquad = ProductQuadrature("gauss-legendre-chebyshev", nazimu, npolar)
+    pquad.OptimizeForPolarSymmetry(4.0 * math.pi)
```
andrsd, line 68: Another heads-up: PR #481 will remove the `OptimizeForPolarSymmetry` from the public API.

tutorials/lbs/first/first_example.py
```diff
@@ -0,0 +1,101 @@
...
+
+sys.path.append(os.path.abspath("../../../"))
+
+from pyopensn.mesh import OrthogonalMeshGenerator, KBAGraphPartitioner
+from pyopensn.mat import AddMaterial
+from pyopensn.xs import LoadFromOpenSn
+from pyopensn.source import VolumetricSource
+from pyopensn.aquad import ProductQuadrature
+from pyopensn.solver import DiscreteOrdinatesSolver, SteadyStateSolver
+from pyopensn.fieldfunc import FieldFunctionGridBased
```
andrsd, line 21: This is comment for everybody. Should we teach users to use the submodule name, i.e. what is being done in lua now?

Example
```
mg = meshgen.OrthogonalMeshGenerator(...)
```

pyopensn/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+    PRIVATE
+    $<INSTALL_INTERFACE:include/opensn>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}/external
+    ${MPI4PY_INCLUDE_DIR}
+)
+message(STATUS "PyOpenSn ${CMAKE_BINARY_DIR}")
```
andrsd, line 55: On my machine I get:
```
PyOpenSn /path/to/opensn/build/temp.macosx-11.1-arm64-cpython-39/pyopensn.__init__
```
What is the purpose of this printout? Is that supposed to help with something? What would that be?

tutorials/lbs/first/first_example.py
quocdang1998: Tomorrow, I'll ask Daryl what the new interface would look like. For now, I'll just keep it as in the first example and save the remove for the next commit.

tutorials/lbs/first/first_example.py
quocdang1998: Acknowledged. I'll ask Daryl about this.

andrsd:
General comment on building. We teach people to use cmake/make to build opensn while here the instructions are: `python setup.py build_ext --inplace`. This builds the code and the python extension, which is good. But, this is a big inconsistency. People might be coming trying to use cmake/make and then wonder why things do not work. Recall that we use the `opensn -i <input file>` approach at the moment. While with python interface the approach is `python3 script.py`...

pyopensn/CMakeLists.txt
quocdang1998: This one is dead code for debug purpose only. I'll remove it in the patch.

quocdang1998:
Right now, user can still use the Lua interface. The classic mkdir build/cmake/make -j still work, but it will build the default library and the Lua interface ONLY (Python interface is not compiled). This is still compatible with the legacy approach.

andrsd:
Yes. I was not trying to say the current approach was changed and/or was not working anymore. More that to raise awareness that something we have been using for 2 years and were teaching users to use will be changed unless a different solution is found. Something to think about: what would be "the way" to run OpenSn in case Lua goes away.

quocdang1998:
I think that we may have to bring this into discussion with everybody on Wed.
I do agree that the moving from Lua to Python will change significantly the way users call OpenSn. In Lua, OpenSn is compiled into an executable, whereas in Python OpenSn is compiled into a module.
Personally, I prefer the "module" approach rather than the executable approach. It is Python native and we can employ the features from multiple libraries in the same script. However, this would need further discussions with everyone else in the meeting tomorrow before making the decision.

wdhawkins:
> Yes. I was not trying to say the current approach was changed and/or was not working anymore. More that to raise awareness that something we have been using for 2 years and were teaching users to use will be changed unless a different solution is found. Something to think about: what would be "the way" to run OpenSn in case Lua goes away.

I think the correct thing to do for the Python interface is to have the Python console analog of the Lua console. We should have an OpenSn binary using the Python interface in the same way we have an OpenSn binary using the Lua interface.  The binary with the Python console should use the existing CLI options (or most of them anyway... for example, the binary with the Python console should not use `--lua`).

tutorials/lbs/first/first_example.py
andrsd: This is an [example](https://github.com/Open-Sn/opensn/pull/496/files#diff-011753cafd0170bae6a7a174884dcf4d296b63aac311e7849ca37ea29200afb1). The material stuff will be completely gone. And XS mapping will be added using the input parameters (see #431 for more details)

pyopensn/src/aquad.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <memory>
+#include <stdexcept>
+
+#include "framework/math/quadratures/angular/angular_quadrature.h"
+#include "framework/math/quadratures/angular/product_quadrature.h"
+#include "framework/math/quadratures/angular/sldfe_sq_quadrature.h"
+#include "framework/logging/log.h"
+
+namespace opensn {
```
andrsd, line 14: For the failing clang-format check. We use the `.clang-format` file from the root of the repo, so you want to set your IDE to use that one. You are most likely using something else. 

Also, make sure that you are using clang-format-14 from your IDE (IIRC there is a plugin for this where you can specify the path to the clang-format binary). It is ok to have different clang-format version and different compiler version (if you are using newer llvm). The problem is that newer versions of clang-format can produce different formatting with the same `.clang-format` file. The version 14 is what the CI uses, so this is why we recommend to use that locally, so you don't have to fight with it.

tutorials/lbs/first/first_example.py
```diff
@@ -0,0 +1,101 @@
...
+    n_cells = 10
+    length = 2.0
+    xmin = - length / 2
+    dx = length / n_cells
+    for i in range(n_cells + 1):
+        nodes.append(xmin + i * dx)
+
+    # generate the mesh
+    meshgen = OrthogonalMeshGenerator(
+        node_sets = [nodes, nodes],
```
andrsd, line 36: As for the python formatting, we have `.flake8` file in the root repo. If you install `flake8`, you can run:
```
cd /path/to/repo
flake8 tutorials
```

And it will tell you locally the problems you need to fix.

setup.py
```diff
@@ -0,0 +1,119 @@
...
+if __name__ == "__main__":
+    setup(
+        name = "pyopensn",
+        author = "The OpenSn Authors",
+        description = "Python extension for OpenSn",
+        packages = ["pyopensn"],
+        ext_modules = [CMakeExtension("pyopensn.__init__")],
+        cmdclass = {"build_ext": CMakeBuilder},
+        install_requires = ["mpi4py", "numpy"]
+    )
```
andrsd, line 119: I will make the comment here, but it is more general that this file only.

If we are building the python binding inside `pyopensn`, I see that `libopensn.0.0.1.dylib`, `libopensn.0.dylib`, and `libopensn.dylib` are also created there.

My preference would be to have the binaries build in the build dir. But then running the python script needs to point there, which may be more difficult to achieve (it depends how the scripts are executed).

Backup plan is to just add:
```
pyopensn/libopensn.0.0.1.dylib
pyopensn/libopensn.0.dylib
pyopensn/libopensn.dylib
```
into `.gitignore` - we don't want people to accidentally stage and commit these files.

andrsd: I see you did actually add `pyopensn/libopensn*.so*` into `.gitignore`, so I guess we just need the `dylib` version of the line

tutorials/lbs/first/first_example.py
ragusa: i am fine with doing the import as in line 15, so line 35 looks good as is to me.

PR #504, opened 2025-02-27
An embedded Python console for OpenSn

wdhawkins:
This PR is now a combination of my original Python console PR and Danny's Python module PR (#501). It introduces an embedded Python console for the C++ OpenSn binary, and a separate Python OpenSn module. The C++ console interface and command-line options mirror the Lua interface where possible. Both the binary and module can run the first OpenSn example tutorial.

To build the C++ binary with the Python console (build with Python will disable the Lua interface):
`cd build; cmake -DOPENSN_WITH_PYTHON=True ..`

To build the Python module:
`python3 setup.py build_ext --inplace` or `pip install .`

I've temporarily excluded the `test` directory from Python builds since it contains some C++ code that directly calls Lua routines.

With a bit more work to add in missing bindings, we will have a fully functional Python interface to OpenSn.

I've updated this PR to use the new quadrature classes.

@andrsd @quocdang1998 ~GCC proved to be difficult with this PR. In particular, it wasn't exporting symbols with the correct visibility in `libopensnpy.so`. What I have here is not exactly an elegant fix, but it works. If you have any other ideas, please let me know. Just setting the visibility via CMake didn't work. The other option would be to pass `-fvisibility=default` to the compiler, but that seems a bit heavy-handed. Using `-fvisiblity=default` for GCC does not resolve the issue.~ This was resolved by using a static library.

With this last commit, both the embedded console and the module run example1.py for Clang and GCC.


CMakeLists.txt
```diff
@@ -228,6 +248,24 @@ if(OPENSN_WITH_LUA)
...
+    # 
+    #add_custom_target(test
+    #  COMMAND
+    #  ${CMAKE_BINARY_DIR}/test/opensn-unit &&
+    #          ${CMAKE_SOURCE_DIR}/test/run_tests
+    #          -d ${CMAKE_SOURCE_DIR}/test
+    #          --exe ${CMAKE_BINARY_DIR}/test/opensn-test
+    #          -w3
+    #  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    #)
```
andrsd, line 263: We probably do not need this commented-out section. When we have testing with python input files, it will be added here...

pyopensn/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+find_package(pybind11 CONFIG)
+
+# Construct module
+file(GLOB_RECURSE PY_INTERFACE_FILE_LIST ${CMAKE_CURRENT_SOURCE_DIR}/*.cc)
+pybind11_add_module(__init__ ${PY_INTERFACE_FILE_LIST})
+
+# Module properties
+#set_target_properties(__init__ PROPERTIES
+#    INTERPROCEDURAL_OPTIMIZATION ON
+#)
```
andrsd, line 29: If we don't need this, then delete it.

python/lib/parameter.cc
```diff
@@ -0,0 +1,113 @@
...
+  TO_PARAMBLOCK(MeshContinuum);
+  TO_PARAMBLOCK(MeshGenerator);
+  TO_PARAMBLOCK(SurfaceMesh);
+  TO_PARAMBLOCK(PointSource);
+  TO_PARAMBLOCK(VolumetricSource);
+  TO_PARAMBLOCK(Solver);
+
+  // throw and return
+  throw std::invalid_argument("Unsupported argument type.");
+  return ParameterBlock();
```
andrsd, line 87: If we `throw`, we do we need a return here?

andrsd (changes requested):

python/lib/parameter.cc
wdhawkins: I haven't tested it, but I think the compiler will complain about a missing return statement if we take it out.

pyopensn/CMakeLists.txt
wdhawkins: Yes. I had forgotten about this. I need to fix CMake so this is optional if it find all of the prereqs. This doesn't not work on our CI machine since it requires LLVMGold, and I don't install it.

python/lib/parameter.cc
wdhawkins: Fixed. Turns out it's not necessary.

python/lib/py_app.h
```diff
@@ -0,0 +1,32 @@
...
+{
+
+class PyApp
+{
+public:
+  explicit PyApp(const mpi::Communicator& comm);
+  int Run(int argc, char** argv);
+
+private:
+  int InitPetSc(int argc, char** argv);
```
zhardy-lanl, line 23:
```suggestion
  int InitPetsc(int argc, char** argv);
```

python/lib/py_app.cc
```diff
@@ -0,0 +1,193 @@
...
+  console.BindModule(wrap_mesh_generator);
+  console.BindModule(wrap_graph_partitioner);
+  console.BindModule(wrap_solver);
+  console.BindModule(wrap_source);
+  console.BindModule(wrap_multigroup_xs);
+  console.BindModule(wrap_create_load);
+}
+
+int
+PyApp::InitPetSc(int argc, char** argv)
```
zhardy-lanl, line 46:
```suggestion
PyApp::InitPetsc(int argc, char** argv)
```

ragusa (approved):

python/lib/py_app.cc
ragusa: will be fixed in #512

CMakeLists.txt
```diff
@@ -39,6 +39,11 @@ if (OPENSN_WITH_CUDA)
 endif()
 option(OPENSN_WITH_DOCS "Enable documentation" OFF)
 option(OPENSN_WITH_LUA "Build with lua support" ON)
+option(OPENSN_WITH_PYTHON "Build with python support" OFF)
+
+if(OPENSN_WITH_PYTHON)
+    set(OPENSN_WITH_LUA OFF CACHE BOOL "Python interface requested, disabling Lua" FORCE)
+endif()
```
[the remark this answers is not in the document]
ragusa: good point. but we plan on deprecating lua sooner rather than later.

python/lib/py_api.h
```diff
@@ -0,0 +1,25 @@
...
+namespace py = pybind11;
+
+namespace opensn
+{
+
+/// Environment initializer and finalizer for OpenSn.
+class PyEnv
+{
+public:
+  PyEnv(void);
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

pyopensn/py_env.cc
```diff
@@ -0,0 +1,57 @@
...
+  py::object mpi_module = py::module::import("mpi4py.MPI");
+  py::object comm_world = mpi_module.attr("COMM_WORLD");
+  mpi_comm = mpi::Communicator(*(::PyMPIComm_Get(comm_world.ptr())));
+  if (mpi_comm.rank() == 0)
+  {
+    std::cout << program << " version " << GetVersionStr() << "\n";
+    std::cout << Timer::GetLocalDateTimeString() << " Running " << program << " with "
+              << mpi_comm.size() << " processes.\n\n";
+  }
+  // initialize PETSc and Opensn
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

andrsd (approved):

Issue #267, opened 2024-05-15
Material object, Mesh blockID, external volumetric source object , and boundary source object


ragusa:
- [ ] material ID in the mesh should be changed to a block ID (or a marker)
- [ ] Currently, the mapping between the mesh material ID and the material stack  is implicit. This is error prone. A material should be created and then a certain number (a list or a lua table) of block IDs should be associated with that material. This is the equivalent of the user saying: "Material matname is associatd with this set of block IDs"
- [ ] this new mapping from material stack to block ID should then be inverted so that, once on a cell, we can quickly identify the "index" of the material in the stack
- [ ] the material object should no longer have an add material routine
- [ ] the material object should no longer have the external isotropic source associated with it. A new object should contain the external source information. there will be a user input mapping  that does "Source srcname is associatd with this set of block IDs". this new mapping from source stack to block ID should then be inverted so that, once on a cell, we can quickly identify the "index" of the sourcein the stack
- [ ] once the external source is set, it will be attached to the sover. 
- [ ] there should be a boundary source object. there will be a user input mapping  that does "BoundarySource bsrcname is associatd with this set of boundary IDs"

zhardy-lanl:
Replaced with #70, #429, #430, and #431

zhardy-lanl:
Replaced with #70, #429, #430, and #431

PR #510, opened 2025-03-04
Replacing material ID with block ID

andrsd:
This finishes task 1 in #267

### DIFF
--- a/modules/diffusion/dfem_diffusion_solver.cc
+++ b/modules/diffusion/dfem_diffusion_solver.cc
@@ -284,7 +284,7 @@ DFEMDiffusionSolver::Execute()
     const auto cc_nodes = cell_mapping.GetNodeLocations();
     const auto fe_vol_data = cell_mapping.MakeVolumetricFiniteElementData();
 
-    const auto imat = cell.material_id;
+    const auto iblk = cell.block_id;
 
     // Assemble volumetric terms
     for (size_t i = 0; i < num_nodes; ++i)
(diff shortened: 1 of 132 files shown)


modules/diffusion/dfem_diffusion_solver.cc
```diff
@@ -302,7 +302,7 @@ DFEMDiffusionSolver::Execute()
     const auto cc_nodes = cell_mapping.GetNodeLocations();
     const auto fe_vol_data = cell_mapping.MakeVolumetricFiniteElementData();
 
-    const auto imat = cell.material_id;
+    const auto iblk = cell.block_id;
```
ragusa, line 305: just noting this occurrence, but could we use something else for a name like `block_id` or `iblock`. i just find that `iblk`, even though as short as `imat` could be obscure but i leave this as a comment, not a requested change. @wdhawkins  opinion?

modules/diffusion/mg_diffusion_solver.cc
```diff
@@ -709,14 +709,14 @@ MGDiffusionSolver::AssembleAbext()
     const auto fe_vol_data = cell_mapping.MakeVolumetricFiniteElementData();
     const size_t num_nodes = cell_mapping.GetNumNodes();
 
-    const auto& xs = matid_to_xs_map_.at(cell.material_id);
+    const auto& xs = matid_to_xs_map_.at(cell.block_id);
```
ragusa, line 712: there seems to be a `matid` here and I think this is perfectly fine as we are using a block id to pick a xs, which is a material property. so i like that we kept `matid` but if this was supposed to be changed and fell through the cracks, let's discuss

zhardy-lanl (approved):

modules/diffusion/mg_diffusion_solver.cc
andrsd: I was not sure if I should change `matid_to_xs_map_`. I can see both `matid_to_xs_map_` and `block_id_to_xs_map_` as valid names, but `block_id_to_xs_map_` is probably a little bit more correct...

ragusa: block_id_to_xs_map_

modules/diffusion/dfem_diffusion_solver.cc
wdhawkins: I think I prefer `block_id = cell.block_id`.

wdhawkins (approved):

modules/diffusion/dfem_diffusion_solver.cc
andrsd: This should be done

modules/diffusion/mg_diffusion_solver.cc
andrsd: This should be done now.

ragusa (approved):

PR #513, opened 2025-03-06
Implement Python interface (caught up with Lua interface)

quocdang1998:
This PR implements the Python binding for OpenSn.

In this pull request:

- Fully bind the `aquad` submodule.
- Fully bind the `logvol` submodule.
- Fully bind the `mesh` submodule.
- Fully bind the `src` submodule.
- Fully bind the `fieldfunc` submodule.
- Fully bind the `xs` submodule.
- Fully bind the `solver` submodule.
- Fully bind the `response` submodule (containing `ResponseEvaluator`).
- Fully bind the `diffusion` submodule.
- Fully bind the `post` submodule.
- Add bindings for log verbosity, Caliper and PETSc in the `settings` submodule.
- Remove bindings for `Material` and `MaterialProperty`.
- Add bindings for `Ylm`, `Vector3`, `ScalarMaterialFunction`, `ScalarSpatialMaterialFunction`, `VectorSpatialFunction` to the `math` submodule.
- Add supports for `Vector3` addition, subtraction, scalar multiplication/division, scalar product and norm.
- Add documentation of the Python interface to Sphinx.
- Add `MPIBarrier` to the console app.
- Bring back some features in the old Lua bindigns:
   * `ComputeFissionRate` -> `LBSSolver::ComputeFissionRate`.
   * `GetFieldFunctionList` -> `Solver::GetFieldFunctions`.

Not yet implemented features:
- `ClearResponseSources`, `AddResponseBuffers`, `AddResponseSources`: failed to wrap functions taking `InputParameters` as arguments, waiting for a version taking `ParameterBlock` or regular C++ arguments.
- `GetPowerFieldFunction`: cannot resolve the "handle". A version returning a pointer/reference is required.
- `GetProductQuadrature`: not sure exactly what the funtion returns.


python/lib/aquad.cc
```diff
@@ -2,151 +2,281 @@
...
+  );
+  simplified_ldfes_quadrature.def(
+    "LocallyRefine",
+    &SimplifiedLDFESQ::Quadrature::LocallyRefine,
+    R"(
+    Locally refines the cells.
+
+    Parameters
+    ----------
+    ref_dir: pyopensn.math.Vector3
```
wdhawkins, line 245: @ragusa Can probably profile descriptions for these parameters.

python/lib/aquad.cc
```diff
@@ -2,151 +2,281 @@
 // SPDX-License-Identifier: MIT
 
 #include "python/lib/py_wrappers.h"
```
wdhawkins, line 4: Includes should be modified to follow our coding style: https://open-sn.github.io/opensn/devguide/coding_standard.html#include-directives.

python/lib/aquad.cc
ragusa: Let's call N the number of refinements. N>=0. What is happening is that a cube, inscribed inside one octant of the unit sphere has its faces  refined into (N+1)^2 facets. Then the facets on the 3 out of 6 faces visible from the sphere are projected onto the unit sphere, resulting in spherical quadrangles. Each facets has 4 angular quadratures points. This makes to total number of directions = (N+1)^2 *3*4*8 (the last factor 8 is for the 8 octants).

The vector is used for local refinement of the angular quadrature and it represent the axis of a cone. The angle of the is given in radians. Spherical quadrangles of the existing SLDFE quadratures are then refined if they are either inside or outside of that cone. The inside versus outside choice is based on the boolean. 

check old code at: https://chi-tech.github.io/de/d7d/group___lua_s_l_d_f_e_s_q.html#gab03ca8a4174984e0978c65e367d44e50

python/lib/aquad.cc
```diff
@@ -2,151 +2,281 @@
...
+    R"(
+    Locally refines the cells.
+
+    Parameters
+    ----------
+    ref_dir: pyopensn.math.Vector3
+        ???
+    cone_size: float
+        ???
+    dir_as_plane_normal: bool, default=False
```
ragusa, line 249: maybe dir_as_plane_normal is not the best name, given the description i gave

python/lib/mesh.cc
```diff
@@ -2,126 +2,315 @@
...
-                     "Set boundary ID's using a logical volume.");
-  mesh_continuum.def("SetupOrthogonalBoundaries", &MeshContinuum::SetupOrthogonalBoundaries, "...");
+    Parameters
+    ----------
+    log_vol: pyopensn.logvol.LogicalVolume
+        ???
+    mat_id: int
+        ???
+    sense: bool
+        ???
```
ragusa, line 64: it's a good idea to use ??? so that i can quickly find these locations.
`log_vol`: logical volume that determines which mesh cells will be selected 
`mat_id`: material ID that will be assigned (I wonder if you have not a behind some PRs. Now, this may be called block_id and the function maybe `SetBlockIDFromLogivalVolume)`
`sense`: the name is bad, replace with `inside`. if true the selected mesh cells are the ones whose centroids are inside the logival volume. otherwise, the selected meshes are the ones whose centroids are outside of the logical volume

python/lib/mesh.cc
```diff
@@ -2,126 +2,315 @@
...
 
+    Parameters
+    ----------
+
+    log_vol: pyopensn.logvol.LogicalVolume
+        ???
+    boundary_name: str
+        ???
+    sense: bool
+        ???
```
ragusa, line 84: similar to my previous comments on the setmaterialIDfrom logical volume, except for boundary IDs. now it will be whether the face centers are inside or not of the logival volume to assign a boundaryID. also change `sense` to `inside`

python/lib/mesh.cc
```diff
@@ -2,126 +2,315 @@
...
+    )",
+    py::arg("log_vol"),
+    py::arg("boundary_name"),
+    py::arg("sense") = true
+  );
+  mesh_continuum.def(
+    "SetupOrthogonalBoundaries",
+    &MeshContinuum::SetupOrthogonalBoundaries,
+    "???"
```
ragusa, line 93: this simply setups boundary IDs for xmin/xmax ymin/ymax zmin/zmax for an domain that is a right parallelpiped

python/lib/mesh.cc
```diff
@@ -2,126 +2,315 @@
...
+    R"(
+    Loads a surface mesh from gmsh's file format.
+
+    Parameters
+    ----------
+    file_name: str
+        ???
+    as_poly: bool
+        ???
+    )",
```
ragusa, line 181: str: filename mesh for the surface mesh
as_poly: if the surface mesh is allowed to contain polygonal facets (as opposed to only triangular faces)
transform: mesh transformation (I think if should be changed to translation as this is the only transformation that is allowed)

create an issue for all these surface mesh stuff and assign it to me for further review

wdhawkins:
Instead of merging, you should `rebase` your changes on `main`. You'll then want to squash trivial commits to keep the git history clean. See here: https://open-sn.github.io/opensn/devguide/workflow.html#updating-your-branch

python/.clang-format
```diff
@@ -0,0 +1 @@
+DisableFormat: true
```
wdhawkins, line 1: Instead of turning `clang-format` off for everything, let's just disable it for the internals of each of the `Wrap` methods.

python/lib/fieldfunc.cc
```diff
@@ -2,44 +2,65 @@
 // SPDX-License-Identifier: MIT
 
 #include "python/lib/py_wrappers.h"
-#include "framework/field_functions/field_function_grid_based.h"
+#include "python/lib/functor.h"  // temporary!
```
wdhawkins, line 5: Please expand on this comment and explain why it's temporary and when it can be removed.

python/lib/functor.h
```diff
@@ -0,0 +1,67 @@
+// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+// This is only a temporary solution!
+// Once the Lua API is eradicated and opensn::Function is replace by std::function
+// this API is no longer needed.
+
+#pragma once
+
+#include <functional>
```
wdhawkins, line 10: Please correct the formatting of includes according to the style guild.

python/lib/math.cc
```diff
@@ -0,0 +1,294 @@
+// SPDX-FileCopyrightText: 2025 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#include "python/lib/py_wrappers.h"
+#include "python/lib/functor.h"  // temporary!!
```
wdhawkins, line 5: I see a more complete comment further down in the file, but please expand here as well on why this is temporary.

python/lib/py_wrappers.h
```diff
@@ -3,59 +3,73 @@
 
 #pragma once
 
-#include "framework/parameters/parameter_block.h"
-#include <pybind11/pybind11.h>
 #include <vector>
 
+#include <pybind11/pybind11.h>
```
wdhawkins, line 8: Please fix the include block according to conform to the style guide.

python/lib/py_wrappers.h
```diff
@@ -3,59 +3,73 @@
...
+
+#include "framework/parameters/parameter_block.h"
+
 namespace py = pybind11;
 
 namespace opensn
 {
 
+// Converter
+// ---------
```
wdhawkins, line 18: Please remove the underlines.

python/lib/py_wrappers.h
```diff
@@ -3,59 +3,73 @@
...
 {
   return py::memoryview::from_buffer(const_cast<T*>(vec.data()), {vec.size()}, {sizeof(T)}, true);
 }
 
 /// Translate a Python dictionary into a ParameterBlock.
 ParameterBlock kwargs_to_param_block(const py::kwargs& params);
 
-/// Wrap the angular quadrature components of OpenSn (unfinished).
+// Module wrappers
+// ---------------
```
wdhawkins, line 31: Remove underlines.

python/lib/solver.cc
wdhawkins, line 1: The clang format commands need to be moved inside `wrap_solver`. `wrap_solver` -> `WrapSolver` for consistency.

python/lib/source.cc
wdhawkins, line 1: There is an extraneous `clang-format on` at the end of the file.

wdhawkins (changes requested):
Just style changes. Everything else looks great.

doc/requirements.txt
```diff
@@ -1,6 +1,6 @@
 breathe
 myst-parser
-sphinx
+sphinx<8
```
ragusa, line 3: just curious. why this?

ragusa (changes requested):
I just mark it as request changes, so that you can add the description of vars, instead of ???, based on my previous comments. I'll change my review to approve right after.

python/lib/fieldfunc.cc
quocdang1998: I want to mark that the "include" of `python/lib/functor.h` will be removed in the future, after we have removed the Lua interface. Once Lua is no longer needed, classes in `framework/math/functions` can be replaced with `std::function`, and binder classes in this file will not be necessary and should be removed.

python/lib/math.cc
quocdang1998: The reason for which these are temporary is in the `python/lib/functor.h` itself.

python/lib/solver.cc
quocdang1998: The solver module is under reformatting. It will be committed in the next commit.

python/lib/fieldfunc.cc
wdhawkins: That's perfect. Just please add this comment or put a comment that says it's temporary and to see `python/lib/functor.h` for for information. Just so other developers understand.

doc/requirements.txt
quocdang1998: Yesterday David and I tried to find the reason why I cannot compile the doc. It was because the version of Sphinx. Somehow, when I downgraded Sphinx from 8 to 7, the error disappeared.

andrsd: I think some sphinx extensions we are using are sphinx7-only. I have seen warnings about features not working with sphinx8. Thus, we have to explicitly force the sphinx version. When the extensions are updated, we could bump the sphinx version as well...

wdhawkins:
@quocdang1998 Everything is looking pretty good to me. If you can squash all of your commits and resolve my clang-format issue, I'll approve.

ragusa (approved):

wdhawkins (approved):

ragusa:
#527 will address the `???` left in the python binding documentation

ragusa:
merged to use in class this week

Issue #448, opened 2024-12-17
extract python routines for LDFE plotting out of c++ code


ragusa:
This portion of code prints out a python code for plotting the LDFE quadrature.
https://github.com/Open-Sn/opensn/blob/114fb361f37428f2210a5bde460610307aaf65eb/framework/math/quadratures/angular/sldfe_sq_quadrature.cc#L805-L965

This portion of code should be removed from the c++ files and placed in a folder in `tools/` (maybe a decent name for the subfolder could be `ang_quad_plotting`. What the routine `SimplifiedLDFESQ::Quadrature::PrintQuadratureToFile` should now do is just export to a CSV files, the information needed for the reworked python file to do the plot.

PR #514, opened 2025-03-06
Removing python verbosity

Aggie04:
This is regarding Issue #448. 

I am going to be perfectly clear, this is the first time I have ever done something like this, so please correct me if I did anything wrong.


ragusa:
pretty good for a first try! I will try to do a quick update to the dev guide so that people know that they need to abide by some formating  standard for python, lua, and c++. currently, you are failing 2 QA tests: formating in c++ and formating in python. technically, your IDE (VS code, CLion,...) could take care of that for you if you let them know. in the meantime, the quickest way is to ping Jason, then me or David

ragusa:
you should had something to the readme found in https://github.com/Open-Sn/opensn/tree/main/tools

after that, you should create a short readme int he new folder you have created inside tools

tools/ang_quad_plotting/plot_quadrature.py
```diff
@@ -0,0 +1,96 @@
+import matplotlib.pyplot as plt
+from mpl_toolkits import mplot3d
+import mpl_toolkits.mplot3d.art3d as art3d
+import mpl_toolkits.mplot3d as ax3
+import matplotlib.transforms as mpltransform
+from pathlib import Path
+import numpy as np
+import math
+
+#====================================== Declare File Base to read (User Edit)
```
andrsd, line 10: I know you just took this from the C++ source file, but please remove the `=` here and below, so the comments are:
```
# My nice comment message here
```

Aggie04:
I completely screwed something up while trying to squash the 3 relevant commits. I have no idea how I managed to pull previous commits into this PR. I am currently in the process of trying to fix this mess.

PR #529, opened 2025-03-17
Add SetOptions to LBSSolver and fix doco bug with Python interface

quocdang1998:
In this PR:
 - `SetOptions` method is wrapped for the Python interface of `LBSSolver`.
 - Documentation bug with Sphinx due to import error is fixed in order to deploy the HTML with Python interface.
 - Add `GetPowerFieldFunction` to `LBSSolver`.
 - Add `ClearForwardSources` (replacing `ClearResponseSources`), `SetBufferOptions` (alternative version of `AddResponseBuffers`, in which user have to loop over the list by themselves) and `SetSourceOptions` (replacing `AddResponseSources`).


ragusa (approved):

.github/workflows/gh-pages.yml
```diff
@@ -42,8 +42,8 @@ jobs:
       - name: Build doco
         run: |
           cmake -S . -B ${{ github.workspace }}/build \
-            -DOPENSN_WITH_DOCS=ON
-          make -C ${{ github.workspace }}/build doc
+            -DOPENSN_WITH_DOCS=ON -DOPENSN_WITH_PYTHON_MODULE=ON
+          make -j64 -C ${{ github.workspace }}/build __init__ doc
```
andrsd, line 46: This workflow does not run on TAMU hardware, but on a github VM. They don't give you 64 cores.
If we are required to build the code for doco, we need to switch this to run on TAMU HW.
I was able to do a full OpenSn build on a github VM, but it takes quite a bit of time (they provide only 2 cores, IIRC).
If you are going to switch this to run on TAMU HW, I see that we use `make -j` which should pickup the right amount of cores

pyproject.toml
```diff
@@ -4,6 +4,7 @@ requires = [
     "wheel",
     "cmake>=3.12",
     "pybind11",
+    "numpy",
```
andrsd, line 7: Comment: If `numpy` is required now for building doco, add it to `doc/requirements.txt` as well.

andrsd (changes requested):

pyproject.toml
quocdang1998: `numpy` is not required to build doco. It is required only to install `pyopensn`.

.github/workflows/gh-pages.yml
quocdang1998: Can I reduce the number of core down to 2? It will take significantly more time to build the documentation< but it will not alter the deployment workflow.
Please tell me what do you think @ragusa?

ragusa: @wdhawkins what do you think? bring down the # of core to 2 on the github VM ? or switch to tamu's hardware ? the intricacy is that the code needs to be built for the doc to be generated (AFAIU)

ragusa (approved):

andrsd (approved):

PR #531, opened 2025-03-18
Solvers are `Solver`s, problems are `Problem`s

andrsd:
- Renaming `LBSSolver` to `LBSProblem`
- Removing `DiscreteOrdinatesSolver` to `DiscreteOrdinatesProblem`
- Renaming `DiscreteOrdinatesCurvilinearSolver` to `DiscreteOrdinatesCurvilinearProblem`
- Renaming `SteadyStateSolver` to `LBSSteadyStateSolver`
- Renaming `NonLinearKEigen` to `NonLinearKEigenSolver`
- Renaming `PowerIterationKEigen` to `PowerIterationKEigenSolver`
- Renaming `PowerIterationKEigenSCDSA` to `PowerIterationKEigenSCDSASolver`
- Renaming `PowerIterationKEigenSMM` to `PowerIterationKEigenSMMSolver`

Refs #471  (task 1)


andrsd:
some files might be also renamed to better match the classes they contain. Just let me know, if you want to use better file names - this would be the time to do that...

ragusa (approved):
great. i hope you wrote a parser in python! we are still doing a lot of renaming!

andrsd:
> i hope you wrote a parser in python

No. I have good refactoring tools that will do all this faster then spending time debugging parsers...

ragusa:
the file names inside sweep_chunk for curvilinear should be renamed.
same thing for the files on level up inside the curvilinear

ragusa:
rename executors --> solvers

zhardy-lanl:
Why does `SteadyStateSolver` get an `LBS` prefix while none of the other solvers do?

andrsd:
> Why does `SteadyStateSolver` get an `LBS` prefix while none of the other solvers do?

When I looked back at the code I think the `LBSSteadyStateSolver` name came from the file name (`lbs_steady_state`). So I undid the rename and renamed the file to match the class name, i.e. `SteadyStateSolver` in `steady_state_solver.{h|cc}` files.

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_problem/sweep_chunks/aah_sweep_chunk_rz.cc
```diff
@@ -1,17 +1,17 @@
...
+#include "modules/linear_boltzmann_solvers/lbs_problem/groupset/lbs_groupset.h"
 #include "framework/math/spatial_discretization/spatial_discretization.h"
 #include "framework/math/quadratures/angular/curvilinear_product_quadrature.h"
 #include "framework/mesh/mesh_continuum/mesh_continuum.h"
 
 namespace opensn
 {
 
-SweepChunkPwlrz::SweepChunkPwlrz(const std::shared_ptr<MeshContinuum> grid,
+AahSweepChunkRZ::AahSweepChunkRZ(const std::shared_ptr<MeshContinuum> grid,
```
ragusa, line 14: i guess this is the name convention, and we cannot use AAH (AdamsAdamsHawkins)

python/lib/py_wrappers.h
```diff
@@ -83,6 +83,7 @@ void py_settings(py::module& pyopensn);
 
 /// Wrap the solver components of OpenSn (unfinshed).
 void py_solver(py::module& pyopensn);
+void WrapProblem(py::module& slv);
 void WrapSolver(py::module& slv);
 void WrapLBS(py::module& slv);
```
ragusa, line 88: for my own comprehension, why are these 3 wrappers having the same arguments? @andrsd @quocdang1998

python/lib/solver.cc
```diff
@@ -388,24 +406,24 @@ WrapLBS(py::module& slv)
   );
 
   // discrete ordinate solver
-  auto do_solver = py::class_<DiscreteOrdinatesSolver, std::shared_ptr<DiscreteOrdinatesSolver>,
-                              LBSSolver>(
+  auto do_problem = py::class_<DiscreteOrdinatesProblem, std::shared_ptr<DiscreteOrdinatesProblem>,
+                              LBSProblem>(
```
quocdang1998, line 410: Pls add one space to line 410 to have indentations match.

ragusa (commented):
looks good to me but i am likely only the 3rd best person to approve, so i will let Danny and Daryl be the 2 minimum approvers for this to go through. if no mevement from them in 1-2days, i'll go back and approve.

python/lib/py_wrappers.h
andrsd: `py::module` is the C++ class that represents a python module that you add the wrapper to. It is not global, so it is passed as an argument. The whole wrapping code is designed around this pattern.
If you are asking why it is `slv`, then my guess is that these APIs are part of the `solver` submodule.

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_problem/sweep_chunks/aah_sweep_chunk_rz.cc
andrsd: This matches the naming scheme of the other sweep chunks `AahSweepChunk` and `CbcSweepChunk`.
However, SPDSs use `AAH_SPDS` and `CBC_SPDS` 🤔

andrsd:
> `DiffusionDFEMSolver` inherits from `LBSProblem`. Should `DiffusionDFEMSolver` be a `Problem` too? (I'm not sure. Pls correct me if I'm wrong)

I did not change anything regarding `DiffusionDFEMSolver`, since as I understood nobody likes this solver/problem class. It is the wrong way how to do diffusion and people want this code gone...

wdhawkins (approved):

quocdang1998 (approved):

zhardy-lanl (approved):

PR #532, opened 2025-03-19
Support for testing both the Lua and Python interfaces

wdhawkins:
This PR adds support for testing the Lua and Python interfaces. The `test` and `tutorials` directories now contain `lua` and `python` subdirectories containing inputs for the respective interfaces. Only a small subset of the regression tests have been converted to Python for this commit.  Python testing will be added to the nightly and regression workflows in an upcoming commit.

The `run_tests` script now attempts to guess which OpenSn binary to use based on the supplied path for the test directory.

To run the Lua tests and tutorials:
`test/run_tests -d test/lua` or `test/run_tests -d tutorials/lua`

To run the Python tests and tutorials:
`test/run_tests -d test/python` or `test/run_tests -d tutorials/python`

The `--exe` command line option can be used to specify a binary if necessary.

One final note. Because the Python interface does not currently bind the OpenSn `Logger`, some of the keys in `tests.json` have changed to correctly match the output.


test/python/modules/linear_boltzmann_solvers/transport_steady/transport_1d_1.py
quocdang1998, line 1: If you do not use Caliper, you don't need to set `use_caliper = True`.

In module mode, Caliper is enabled through `settings.EnableCaliper`.

wdhawkins: Fixed.

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_1_poly.py
quocdang1998, line 1: Module mode doesn't have `size` and `rank`. These are inteded to be retrieved by users with `mpi4py`. Can you add this in the import block:

```
if "opensn_console" not in globals():
    from mpi4py import MPI
    size = MPI.COMM_WORLD.size
    rank = MPI.COMM_WORLD.rank
    from pyopensn...
    ...
```

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_1_poly.py
```diff
@@ -0,0 +1,132 @@
...
+2D PWLD transport test with vacuum and Incident-isotropic bundary conditions
+Test: Max-value=0.50758 and 2.52527e-04
+"""
+
+import os
+import sys
+import math
+
+if "opensn_console" not in globals():
+    sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
quocdang1998, line 14: This line is meant to add the project directory to sys.path, so that Python can find `pyopensn`. Please make sure that it points to the correct folder.

For example, `os.path.join(os.path.dirname(__file__)` will point to `test/python/modules/linear_boltzmann_solvers/transport_steady`, the directory containing the test script. You will need to go up 5 levels (`../../../../..`) to find `pyopensn`, instead of just 1.

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_1_poly.py
wdhawkins: Fixed.

ragusa (approved):

CMakeLists.txt
```diff
@@ -223,6 +223,7 @@ set_target_properties(
 if(OPENSN_WITH_LUA)
     add_subdirectory(lua)
     add_subdirectory(test)
+    add_subdirectory(test/lua)
```
andrsd, line 226: This should be in `test/CMakeLists.txt` not here.

andrsd (changes requested):

CMakeLists.txt
wdhawkins: Fixed.

andrsd (approved):

ragusa (approved):

PR #539, opened 2025-03-19
Run gh-pages on nuen-class02

quocdang1998:


ragusa:
I am not sure what this does in details, so my approval on this one has little value. It looks like a certain number of explicit actions have been removed (or maybe this as still there, but implicitly through `- uses: actions/checkout@v3`. I'd say one approver is enough but it should be @wdarylhawkins or @andrsd

wdhawkins:
I think the `concurrency` and `defaults` sections can be removed.

.github/workflows/gh-pages.yml
```diff
@@ -13,41 +13,17 @@ jobs:
...
-            petsc==3.17.5 \
-            vtk \
-            lua \
-            caliper \
-            pybind11 \
-            boost
-          sudo pip install -r doc/requirements.txt
-
-      - name: Build doco
+      - uses: actions/checkout@v3
```
andrsd, line 18: `v3` is most likely incorrect. There were issues with node16 (IIRC), should be using `v4` (node20). This is where github actions moved, so we should be following the suit.

.github/workflows/gh-pages.yml
```diff
@@ -13,41 +13,17 @@ jobs:
...
+        shell: bash
         run: |
-          cmake -S . -B ${{ github.workspace }}/build \
-            -DOPENSN_WITH_DOCS=ON -DOPENSN_WITH_PYTHON_MODULE=ON
-          make -j2 -C ${{ github.workspace }}/build __init__ doc
+          module load opensn/gcc/13.3.0 python3/3.12.3 doxygen/1.13.2
+          mkdir build && cd build
+          cmake -DOPENSN_WITH_DOCS=ON -DOPENSN_WITH_PYTHON_MODULE=ON ..
+          make -j __init__ doc
+          cd ..
```
andrsd, line 26: The `cd ..` does not achieve anything here and should be removed. Explanation: each step is spawned in its own subprocess.

andrsd (changes requested):

.github/workflows/gh-pages.yml
quocdang1998: I removed ``cd ..`` in the latest (forced) commit.

.github/workflows/gh-pages.yml
quocdang1998: I changed it in the latest (forced) commit.

wdhawkins (approved):

ragusa (approved):
if you think you are resolved the comments, mark them as resolved after adding a short explanation in each discussion box. merging is blocked until conversations are resolved

ragusa:
before merging, renaming PR (it is not on class-002 anymore, is it?)

ragusa:
address #534

PR #541, opened 2025-03-21
Fix segfault bug

quocdang1998:
In this PR:
 - Segfault error when using `VectorSpatialFunction` with `ParameterBlock` in Python interface is fixed.
 - CSS file is added to enhance readability of the Python interface documentation.


doc/_static/FiraCode-Regular.woff2
ragusa, line 1: i suppose those binary files are required for the font. Are they legit? Are they not directly available from Google ?
https://fonts.google.com/
https://github.com/google/fonts

wdhawkins (approved):

doc/_static/FiraCode-Regular.woff2
quocdang1998: As far as I can tell, `woff2` is a legit font format for almost any browsers.

Of course, we can have users' browser download the fonts (automatically) from the web. But I think it would be better to have the font offline, so that user can copy the doc from one machine to another. Even when the destination machine doesn't have Internet connection, user can still open the html.

Issue #554, opened 2025-03-25
Gtest linking with a conda environment fails


kitamuraa24:
When linking gtest to the `opensn-unit`, the linking will fail. This error arises from the fact that if the user is building within a conda environment, the conda environment already has gtest. Currently, the user needs to deactivate their conda environment to successfully build in the current state of things. I propose that in the CMake script, we try to locate a prexisiting gtest and not build gtest and link against the pre-existing gtest (in the conda environment or elsewhere).

andrsd:
My understanding of this issue is that if some software package "by accident" installs a different version of gtest/gmock in the system tree, then linking of opensn-unit may fail, because the linking process might pull in incorrect library, see this [SO](https://stackoverflow.com/questions/39207940/undefined-reference-when-linking-with-googletest) for a similar scenario. So, users/developers don't even get to running tests...

And, yes, this happened to me before where IIRC yaml-cpp installed gtest/gmock (don't ask me why - it could have been incorrectly deployed...)

loganharbour:
@andrsd we could also search for system gtest and use it if it is available, otherwise install it with cmake.

kitamuraa24:
@loganharbour Yea the tests fail if i try to execute it that way. If I just run `make test` it is fine, but I still fail some tests due to lack of resources on my computer. This was something that I got around by running `./test/run_tests -j4 --exe build/test/opensn-test` in the past (where `opensn-unit` appears to have replaced `opensn-test` ) but this appears to not work anymore. In terms of the issue I presented in this IR, it is relevant to the fact that if the user has anaconda, they likely have a gtest installed in their environment, and then when the build for OpenSn occurs, it has trouble linking between a pre-existing system gtest and the gtest that was built for OpenSn. So as you and @andrsd were discussing, I think this can be averted by just looking for a system gtest and if it exists, skip building another one and link against that.

andrsd:
> [@andrsd](https://github.com/andrsd) we could also search for system gtest and use it if it is available, otherwise install it with cmake.

:+1: this is what I was thinking we could do. There is also this thing that gtest is a package in linux distros, so it could be legit installed on the system. So it would make sense to pick that over the opensn-suplied one...

andrsd:
> where `opensn-unit` appears to have replaced `opensn-test`

This is not correct:
-  `opensn-unit` are unit tests implemented using gtest.
[...]

PR #561, opened 2025-03-28
Modified the CMakeLists.txt file in the main dir and the test dir.

kitamuraa24:
We now look for a system gtest and link against that instead of building gtest locally and then linking against that. If a system gtest is not found, gtest is built locally and then linked against. This is in response to Issue #554.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,12 +125,16 @@ if(caliper_FOUND)
 endif()
 
 # google test
-FetchContent_Declare(
-    googletest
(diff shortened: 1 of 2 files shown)


kitamuraa24:
@andrsd Yea that's my bad. I thought I fixed those earlier before I pushed last night, but my git was acting funny on my personal machine last night. I updated accordingly to your comments.

andrsd (approved):

loganharbour (approved):

PR #570, opened 2025-04-01
Python testing updates

wdhawkins:
1. Updating Python post processor tests to ensure consistent output across platforms.
2. Resolving a stack-use-after-scope issue in the Python binding of MPIBarrier.
3. Updating LSAN suppressions for Python.
4. Increasing the job wait time in the regression testing script to improve reliability.
5. Updating large C5G7 inputs to correctly set number of sweeps for SCDSA and SMM methods.

### DIFF
--- a/python/lib/console.cc
+++ b/python/lib/console.cc
@@ -31,7 +31,7 @@ Console::BindBarrier(const mpi::Communicator& comm)
   py::module main = py::module::import("__main__");
   main.def(
     "MPIBarrier",
-    [&comm](void)
+    [comm](void)
     {
       comm.barrier();
     },
--- a/test/python/framework/post_processors/solver_info_01.py
+++ b/test/python/framework/post_processors/solver_info_01.py
@@ -42,10 +42,10 @@ if __name__ == "__main__":
     for t in range(1, 21):
         phys0.Step()
         time = phys0.GetTimeNew()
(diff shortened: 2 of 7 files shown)


ragusa (approved):
I am just surprised that you had to increase the sleep time in the job manager by a factor 100. Was that necessary ?

python/lib/console.cc
quocdang1998, line 1: I'm just curious. Why do we need a copy of the communicator rather than a reference?

quocdang1998 (approved):

python/lib/console.cc
ragusa: I am curious too

wdhawkins: `BindBarrier` just registers a callback in the Python environment that is stored and invoked when Python code calls `MPIBarrier`. The lambda captured `comm` by reference and it was stored as part of the callback. But, that reference has `BindBarrier` scope. When `BindBarrier` returns, `comm` goes out of scope, and the stored reference is invalid.

wdhawkins:
> I am just surprised that you had to increase the sleep time in the job manager by a factor 100. Was that necessary ?

I tried 50x to start, but, in my very limited testing, I saw a ~20% failure rate.  With 1s, I have yet to see a failure... but time will tell.

PR #572, opened 2025-04-02
Disabling Python interface if Lua interface is requested.

wdhawkins:


CMakeLists.txt
```diff
@@ -46,8 +46,8 @@ if(OPENSN_WITH_PYTHON_MODULE)
     set(OPENSN_WITH_PYTHON ON CACHE BOOL "Python support requested for building module" FORCE)
 endif()
 
-if(OPENSN_WITH_PYTHON)
-    set(OPENSN_WITH_LUA OFF CACHE BOOL "Python interface requested, disabling Lua" FORCE)
+if(OPENSN_WITH_LUA)
+    set(OPENSN_WITH_PYTHON OFF CACHE BOOL "Lua interface requested, disabling Python interface" FORCE)
```
ragusa, line 50: shouldn't OPENSN_WITH_PYTHON then be turned ON by default on line 42 (and OPENSN_WITH_LUA be OFF on line 41)?

also, shouldn't line 50 also set OPENSN_WITH_PYTHON_MODULE to be OFF ?

quocdang1998 (approved):

CMakeLists.txt
wdhawkins: Good catch! I forgot to rebase on main. Also, good point about the Python module. This should all be fixed now.

ragusa (approved):

PR #575, opened 2025-04-03
adding python mesh tutorials

ragusa:
- adding python mesh tutorials (console only). they will be converted to notebook soon but i do not want them to regress in the meantime
- some small change to the python API (comments, renaming of a method)


python/lib/mesh.cc
```diff
@@ -695,11 +695,11 @@ WrapGraphPartitioner(py::module& mesh)
...
     Construct a PETSc based graph partitioner.
 
-    Parameters
-    ----------
-    type: str, default='parmetis'
-        Type of PETSc partitioner.
-    )"
+  Parameters
+  ----------
+  type : {'parmetis', 'ptscotch'}, default='parmetis'
```
quocdang1998, line 700: Remove the space before the `:` for consistency.

quocdang1998 (changes requested):
PLs fix the small typo and I'll approve.

wdhawkins (approved):

python/lib/mesh.cc
ragusa: fixed

quocdang1998 (approved):

PR #583, opened 2025-04-05
Moving coordinate system info

zhardy-lanl:
This PR moves coordinate system information to the mesh from discretizations. Coordinate systems can now be set within `OrthogonalMeshGenerator` and `FromFileMeshGenerator` using the `coord_sys` argument in input parameters. Coordinate system inputs were also standardized to `coord_sys` across the input language and the curvilinear solver was cleaned up.

### DIFF
--- a/framework/math/spatial_discretization/finite_element/finite_element_base.h
+++ b/framework/math/spatial_discretization/finite_element/finite_element_base.h
@@ -22,10 +22,9 @@ public:
 
 protected:
   explicit FiniteElementBase(const std::shared_ptr<MeshContinuum> grid,
-                             CoordinateSystemType cs_type,
                              SpatialDiscretizationType sdm_type,
                              QuadratureOrder q_order)
-    : SpatialDiscretization(grid, cs_type, sdm_type), q_order_(q_order)
+    : SpatialDiscretization(grid, sdm_type), q_order_(q_order)
   {
   }
 
--- a/framework/field_functions/field_function_grid_based.cc
+++ b/framework/field_functions/field_function_grid_based.cc
@@ -30,7 +30,7 @@ FieldFunctionGridBased::GetInputParameters()
 
   params.AddOptionalParameter("discretization", "FV", "The spatial discretization type to be used");
   params.AddOptionalParameter(
-    "coordinate_system", "cartesian", "Coordinate system to apply to element mappings");
+    "coord_sys", "cartesian", "Coordinate system to apply to element mappings");
   params.AddOptionalParameter("quadrature_order",
                               0,
                               "If supplied, will overwrite the default for the "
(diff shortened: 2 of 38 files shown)


ragusa (approved):

framework/math/spatial_discretization/finite_element/finite_element_base.h
```diff
@@ -21,11 +21,10 @@ class FiniteElementBase : public SpatialDiscretization
   QuadratureOrder GetQuadratureOrder() const;
 
 protected:
-  explicit FiniteElementBase(const std::shared_ptr<MeshContinuum> grid,
-                             CoordinateSystemType cs_type,
+  explicit FiniteElementBase(const std::shared_ptr<MeshContinuum>& grid,
```
andrsd, line 24: I see you are doing this consistently, so just that I understand: In terms of native pointers this would be equivalent for doing `const MeshContinuum *& grid`. Do you really need that? Or, plain `const *` is enough as it was before?

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/lbs_curvilinear_solver.cc
```diff
@@ -244,87 +206,71 @@ DiscreteOrdinatesCurvilinearSolver::PerformInputChecks()
...
 void
 DiscreteOrdinatesCurvilinearSolver::InitializeSpatialDiscretization()
 {
   log.Log() << "Initializing spatial discretization_.\n";
 
-  auto qorder = QuadratureOrder::INVALID_ORDER;
-  auto system = CoordinateSystemType::UNDEFINED;
-
-  //  primary discretisation
+  //  primary discretization
```
andrsd, line 223: You can fix the double space here as well (I saw you fixed it above). In case you care...

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_solver/lbs_curvilinear_solver.cc
```diff
@@ -244,87 +206,71 @@ DiscreteOrdinatesCurvilinearSolver::PerformInputChecks()
...
 
   ComputeUnitIntegrals();
 
-  //  secondary discretisation
-  //  system - manipulated such that the spatial discretisation returns
+  //  secondary discretization
+  //  system - manipulated such that the spatial discretization returns
   //  a cell view of the same type but with weighting of degree one less
-  //  than the primary discretisation
+  //  than the primary discretization
```
andrsd, line 252: The same deal with double space here...

framework/math/spatial_discretization/finite_element/finite_element_base.h
zhardy-lanl: I can revert this. Changed because of a clang-tidy suggestion.

andrsd: I guess it was suggesting that `const shared_ptr<> &` would avoid the unneeded copy? So, maybe keep this. Your call really...

zhardy-lanl: I already reverted it. We'll leave it as is. And yes, it was to avoid the unneeded copy.

wdhawkins (approved):

PR #587, opened 2025-04-05
Falling back to module mode if find_package(Boost CONFIG) fails

wdhawkins:
Header only installs of Boost don't include `BoostConfig.cmake`. Since many installs are header-only, fall back to using the `FindBoost` module if `find_package(Boost CONFIG)` fails.


ragusa (approved):
what is "module mode"?

wdhawkins:
> what is "module mode"?

module mode falls back to having cmake call the `FindBoost` module.

andrsd (approved):

PR #590, opened 2025-04-08
Adding a CMake script to configure and install dependencies

### LINKED ISSUE #87: Update installation instructions

### LINKED ISSUE #546: pybind11 is required and this is not mentioned
we need to update the doc so that users know that they need to pip install pybind11. this is missing int he install instructions

### LINKED ISSUE #592: deprecate configure_dependencies.py

wdhawkins:
This PR replaces `configure_dependencies.py` with `tools/dependencies/CMakeLists.txt`. This has a number of benefits:
1. The `CMake` version is easier to maintain and less complicated
2. `CMake` allows us to check for existing packages before installing dependencies
3. Finding compilers, downloads, etc. are automatically handled by `CMake`

To install dependencies with the new script, do the following (from the top-level OpenSn directory):
```
mkdir build_deps
cd build_deps
cmake -DCMAKE_INSTALL_PREFIX=../../opensn_dependencies ../tools/dependencies
make -j
cd ..
rm -rf build_deps
```

I'm leaving `configure_dependencies.py` as is until I finish updating the documentation. Once the documentation is finished, `configure_dependencies.py` will be removed.

Finally, the `CMake` dependency install includes the following improvements:
1. Updates PETSc to 3.23.0 and includes SLEPc
2. Updates HDF5 to 1.14.6
3. Updates Caliper to 2.13.0
4. Removes the need for a Fortran compiler to install PETSc

I've updated the documentation and removed the manual install instructions. Installing the dependencies should be done using the CMake install script.

Closes #87
Closes #592
Closes #546


tools/dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+endif()
+
+# Lua
+if(ENABLE_LUA)
+  ExternalProject_Add(lua
+    URL https://www.lua.org/ftp/lua-5.4.6.tar.gz
+    SOURCE_DIR ${CMAKE_BINARY_DIR}/lua-src
+    CONFIGURE_COMMAND ""
+    #BUILD_COMMAND cd <SOURCE_DIR> && make ${OS_TAG} MYCFLAGS=-fPIC MYLIBS=-lncurses -j${JOBS}
+    #INSTALL_COMMAND cd <SOURCE_DIR> && make install INSTALL_TOP=${CMAKE_INSTALL_PREFIX}
```
andrsd, line 67: is this left-over from debugin'?

quocdang1998 (approved):

tools/dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+endif()
+message(STATUS "CMAKE_INSTALL_PREFIX is set to: ${CMAKE_INSTALL_PREFIX}")
+
+set(OPENSN_ENV_DIR "${CMAKE_INSTALL_PREFIX}/bin")
+set(ENV_SCRIPT "${OPENSN_ENV_DIR}/set_opensn_env.sh")
+
+find_package(MPI REQUIRED)
+
+option(ENABLE_BOOST   "Install Boost"   ON)
+option(ENABLE_LUA     "Install Lua"     ON)
```
ragusa, line 23: Should lua be on?

wdhawkins: For now, I think it should be. I also need to add Python. I'll revise to check for all of the required packages and install them if not found.

ragusa (approved):

tools/dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,200 @@
...
+  endif()
+endif()
+
+if(INSTALLED_DEP_COUNT EQUAL 0)
+  message(STATUS "All dependencies satisfied; no package installations necessary.")
+else()
+  add_custom_target(finalize ALL
+    COMMAND ${CMAKE_COMMAND} -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -P "${CMAKE_SOURCE_DIR}/env_script.cmake"
+  )
+endif()
```
quocdang1998, line 200: I've just realized that `pandoc` is needed to build documentation with `nbsphinx`. Should we include it here or let user install it on their own if they want to build the documentation?

wdhawkins: The easiest methods for installing Pandoc are using the Pandoc installer, or using the OS package installer. I've opted to let the user install Pandoc instead of trying to install it as part of the install script.

doc/install/install.rst
```diff
@@ -1,224 +1,37 @@
...
-2. ``gfortran`` (required by the BLAS component of PETSc)
-3. ``flex`` and ``bison`` (required by the PTSCOTCH component of PETSc)
-4. Python3 v3.9+ and ``pip`` (required by PETSc and OpenSn)
-5. Git version control system
-6. CMake v3.12+
-7. MPI (OpenMPI, MPICH, and MVAPICH have been tested)
-8. HDF5
-9. Doxygen (required for generating the OpenSn documentation)
+1. A recent version of ``clang++``/``g++`` that supports C++20
+2. ``flex`` (required by the PTSCOTCH component of PETSc)
```
andrsd, line 10: `bison` is still needed, see [this](https://gitlab.com/petsc/petsc/-/blob/main/config/BuildSystem/config/packages/PTScotch.py#L40-41)

andrsd (dismissed):

doc/install/install.rst
wdhawkins: Instead of adding `bison` as a system dependency, it's installed with `PETSc`.

andrsd: Yeah. I missed that. 👍

andrsd (approved):

quocdang1998 (approved):

PR #598, opened 2025-04-11
Fixing install target for lua

andrsd:
- Lua is a dependency, so needs to be included and linked publicly, so the depending apps do not have to bother with discovering, including, and linking the libs. We really want to use the header location and libs we built OpenSn against
- Also, install LuaBridge so that OpenSn-based apps can register/bind their new classes


andrsd:
Tagging @quincy-huhn98 who needs this.

wdhawkins (approved):

ragusa (approved):

PR #609, opened 2025-04-15
Addition of Lebedev Quadrature into OpenSN.

Aggie04:
The Lebedev Quadrature class was added as an additional option within the angular quadrature suite. Shortened to "Lebedev" for the sake of brevity, although the quadrature is also occasionally called Lebedev-Laikov Quadrature. The class does not generate the necessary points every time it is called, instead, because of the limitted variety, it reads from pre-computed CSV files that contain the various cartesian points and weights for a given lebedev rule. All associated code was primarily based off of the code that forms the SLDFES Quadrature.


Aggie04:
@ragusa can you add reviewers?

wdhawkins:
Instead of reading in the `.csv` files at run time, we should leverage cmake to generate a c++ header at compile time that includes the contents of the csv files as regular c++ double arrays.  This is relatively clean and prevents us from having to carry around the csv files. Let me see if I can come up with an example.

wdhawkins:
Alternatively, if the number of csv files is relatively static, we can just encode the data into arrays in a header file.

Aggie04:
Lebedev isn't very customizable as a quadrature. If the user wants to use Lebedev, they supply a single arguement (Lebedev Order), which corresponds to a predetermined list points and weights. The list of orders is finite and the points for all orders are all tabulated in literature.

framework/math/quadratures/angular/lebedev_orders/lebedev_order003.csv
```diff
@@ -0,0 +1,6 @@
+1.0,0.0,6.123233995736766e-17,0.166666666666667
```
andrsd, line 1: I assume you computed these somehow and you got these machine zero numbers. You probably want to filter those and have exact `0.` instead.

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,169 @@
...
+    const double scale_factor = expected_sum / weight_sum;
+    for (auto& w : weights)
+      w *= scale_factor;
+
+    if (verbose)
+      log.Log() << "Weights have been normalized to sum to 1.";
+  }
+}
+
+// Method to get octant optimized quadrature
```
andrsd, line 123: You already have a much better doco in the header file, remove this...

framework/math/quadratures/angular/lebedev_quadrature.h
```diff
@@ -0,0 +1,63 @@
+// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#ifndef LEBEDEV_QUADRATURE_H
+#define LEBEDEV_QUADRATURE_H
```
andrsd, line 5: Use `#pragma once`

andrsd (changes requested):
The CSV tables should be hard-coded in the .cc file. At some point we will be deploying the code (we are already starting to do that) and you really don't want user to deal with file path when they want to use quadrature

Aggie04:
@wdhawkins 
I talked with Dr. Ragusa and we decided to go forward with creating a Header that contains all the data instead of the Cmake method. Please let me know if you disagree and should move forward with the Cmake process instead.

wdhawkins:
> @wdhawkins I talked with Dr. Ragusa and we decided to go forward with creating a Header that contains all the data instead of the Cmake method. Please let me know if you disagree and should move forward with the Cmake process instead.

 I think that's a good solution.

python/lib/aquad.cc
```diff
@@ -330,6 +331,76 @@ WrapSLDFESQuadrature(py::module& aquad)
...
+    Wrapper of :cpp:class:`opensn::LebedevQuadrature`.
+    )"
+  );
+  
+  lebedev_quadrature.def(
+    py::init(
+      [](int order, bool verbose)
+      {
+        try {
+          // Use the 3-parameter constructor directly
```
ragusa, line 359: i thought it was a 2-parameter constructor? `order` and `verbose`

ragusa (changes requested):

python/lib/aquad.cc
AndrewNJ03: It is, the "3-parameter" was a typo left behind when I had the file path specified. 

This has been removed

ragusa (approved):

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
...
+    {
+      double x = point.x;
+      double y = point.y;
+      double z = point.z;
+      double w = point.weight;
+
+      // Calculate phi and theta from x, y, z
+      double r = sqrt(x * x + y * y + z * z);
+      double theta = acos(z / r);
+      double phi = atan2(y, x);
```
andrsd, line 50: Should be `std::sqrt`, `std::acos`, and `std::atan2`.

This is not being done consistently in the code, but it is the preferred way in modern C++.

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
...
+    if (verbose)
+    {
+      log.Log() << "Loaded " << point_count << " Lebedev quadrature points from order " << order;
+      log.Log() << ostr.str() << "\n"
+                << "Weight sum=" << weight_sum;
+    }
+
+    // Check weight sum (should be 1 for 3D)
+    const double expected_sum = 1.0;
+    if (fabs(weight_sum - expected_sum) > 1.0e-10)
```
andrsd, line 95: `std::fabs`

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
+// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
```
andrsd, line 1: This should be the year the file was created, i.e. 2025 (see [TLDR](https://liferay.dev/blogs/-/blogs/how-and-why-to-properly-write-copyright-statements-in-your-code#tldr), and more [details](https://liferay.dev/blogs/-/blogs/how-and-why-to-properly-write-copyright-statements-in-your-code#why-keep-the-year) )

Also, fix in the other new files.

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
...
+        ostr << buf;
+      }
+
+      point_count++;
+    }
+
+    // If no points found
+    if (abscissae.empty())
+      throw std::invalid_argument("LebedevQuadrature: No quadrature points found for order: " +
+                                  std::to_string(order));
```
andrsd, line 84: The search is done inside `LebedevOrders::GetOrderPoints`. If it fails, we would not get here at all, because the function throws, so this seems like an overzealous check.

If the point of this code is to prevent empty `abscissae`, how could this happen? Is it expected that `GetOrderPoints` could return an empty `vector`? If so, you want to rephrase this to "if no points were created" and change the exception text similarly.

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
...
+
+    if (verbose)
+    {
+      log.Log() << "Loaded " << point_count << " Lebedev quadrature points from order " << order;
```
andrsd, line 88: Why not use `points.size()` instead of `point_count`?

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
...
+      log.Log() << ostr.str() << "\n"
+                << "Weight sum=" << weight_sum;
+    }
+
+    // Check weight sum (should be 1 for 3D)
```
andrsd, line 93: Misleading comment. 1) The  check is done regardless of spatial dimension. 2) code really normalizes the weights if they don't add up to 1. But an average coder can see that just by looking at the code, so why comment at all? Remember: comments should be used when it is not obvious what is going on... And this code is pretty self-explanatory...

framework/math/quadratures/angular/lebedev_quadrature.cc
```diff
@@ -0,0 +1,119 @@
...
+        w *= scale_factor;
+
+      if (verbose)
+        log.Log() << "Weights have been normalized to sum to 1.";
+    }
+  }
+  catch (const std::invalid_argument& e)
+  {
+    throw std::invalid_argument("LebedevQuadrature: Failed to load order " + std::to_string(order) +
+                                ": " + e.what());
```
andrsd, line 115: Is the "load" a leftover from the previous implementation that was using the CSV file? We are not loading anything any more, right?

More over, this feels weird, the error states are already throwing with good error messages and this just masks them and re-throwing. You probably do not need the try-catch at all...

andrsd (changes requested):

quocdang1998 (changes requested):

python/lib/aquad.cc
```diff
@@ -330,6 +331,74 @@ WrapSLDFESQuadrature(py::module& aquad)
...
+        try {
+          // Use the 2-parameter constructor directly
+          return std::make_shared<LebedevQuadrature>(order, verbose);
+        }
+        catch (const std::exception& e) {
+          // Convert C++ exceptions to Python exceptions
+          PyErr_SetString(PyExc_RuntimeError, e.what());
+          throw py::error_already_set();
+        }
+      }
```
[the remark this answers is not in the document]
wdhawkins: This quadrature will need to be normalized to 1.0.

framework/math/quadratures/angular/lebedev_quadrature.cc
AndrewNJ03: 👍 Updated

Additionally a handful of variables were updated to be const if they weren't altered directly. I am not sure if this is a common practice or used different across other parts of OpenSN, but I was told that it is good practice to declare unaltered values as const.

framework/math/quadratures/angular/lebedev_quadrature.cc
AndrewNJ03: Fixed

framework/math/quadratures/angular/lebedev_quadrature.cc
AndrewNJ03: Removed. 
I think this was an artifact from when I was bug testing and passing through wrong orders for the quadrature.

framework/math/quadratures/angular/lebedev_quadrature.cc
AndrewNJ03: Updated to points.size() so it's more explicitly clear what is being addressed in the verbose statement.

framework/math/quadratures/angular/lebedev_quadrature.cc
AndrewNJ03: The comment was initially in place to help create "sections" in the code so that a developer could glance at a block and get an understanding of what is being intended without having to make an assumption. The parenthesis were in place due to Lebedev being explicitly a 3D quadrature, and for added clarity incase someone is unfamiliar with the normalization in OpenSN (1 opposed to 4pi). 

I removed the parenthesis, and can remove the comment altogether if I should just comment for the sake of specific expression clarity.

AndrewNJ03:
@andrsd @quocdang1998 
I don't have permission to re-request or request reviews for some reason, but this PR is good to go for y'all to review the changes I made.

wdhawkins:
> @andrsd @quocdang1998 I don't have permission to re-request or request reviews for some reason, but this PR is good to go for y'all to review the changes I made.

I re-requested reviews by everyone. Thanks for the contribution!

framework/math/quadratures/angular/lebedev_quadrature.cc
andrsd: It is in the code quite a bit (probably not everywhere) and a good thing to do in general :+1:

andrsd (approved):

ragusa (approved):

quocdang1998 (approved):

AndrewNJ03:
The force pushes after the last review were my attempt at both squashing and rebasing the code. I kept sending bad commands to git which caused me to need to repeat the process a multitude of times in order to get a clean branch that can merge without conflicts. There weren't any changes made. 

@ragusa 
I don't think I have permission (on either of my accounts, although this one would be more suitable) to complete the Pull Request, at least, there isn't a button prompting me to do so. Would you mind completing this Pull Request for me?

wdhawkins (approved):

PR #612, opened 2025-04-17
Adding checks to ensure that all mesh block ids have entries in the user-supplied xs map

wdhawkins:


ragusa (approved):
Should negative block ids be allowed now?

wdhawkins:
> Should negative block ids be allowed now?

Good question.  There's no reason why we couldn't do that.

ragusa:
> > Should negative block ids be allowed now?
> 
> Good question. There's no reason why we couldn't do that.

if they are not `unsigned int`, what we only need to do (which we added) is a check that the block id is found in the map. but, for now, let's not rattle this boat, I guess.

andrsd (approved):

PR #615, opened 2025-04-18
returning the volume per blockID in python

ragusa:
returning the volume per blockID in python

usage is:
```
volumes_per_block = grid.ComputeVolumePerBlockID()

# iterate through block → volume
for block_id, volume in volumes_per_block.items():
    print(f"Block {block_id}: volume = {volume:.6e}")
```

### DIFF
--- a/framework/mesh/mesh_continuum/mesh_continuum.cc
+++ b/framework/mesh/mesh_continuum/mesh_continuum.cc
@@ -767,8 +767,8 @@ MeshContinuum::MapCellFace(const Cell& cur_cell, const Cell& adj_cell, const uns
   throw std::logic_error("MeshContinuum::MapCellFace: Mapping failure.");
 }
 
-void
-MeshContinuum::ComputeVolumePerBlockID()
+std::map<int, double>
+MeshContinuum::ComputeVolumePerBlockID() const
 {
   // Create a map to hold local volume with local block as key
   std::map<int, double> block_volumes;
@@ -823,13 +823,7 @@ MeshContinuum::ComputeVolumePerBlockID()
     global_block_volumes[global_block_ids[i]] = global_volumes[i];
   }
 
-  // Output the volumes per material_id on the root process
-  if (mpi_comm.rank() == 0)
-  {
-    for (const auto& [id, vol] : global_block_volumes)
-      log.Log() << "Block ID: " << id << " Volume: " << std::setprecision(12) << vol << std::endl;
-    log.Log() << std::endl;
-  }
+  return global_block_volumes;
 }
 
 } // namespace opensn
--- a/python/lib/mesh.cc
+++ b/python/lib/mesh.cc
@@ -18,6 +18,7 @@
 #include "framework/mesh/surface_mesh/surface_mesh.h"
 #include "framework/utils/timer.h"
 #include <pybind11/functional.h>
+#include <pybind11/stl.h>
 #include <cstdint>
 #include <functional>
 #include <memory>
@@ -120,7 +121,14 @@ WrapMesh(py::module& mesh)
   mesh_continuum.def(
     "ComputeVolumePerBlockID",
     &MeshContinuum::ComputeVolumePerBlockID,
-    "Compute volume per block ID."
+    R"(
+    Compute volume per block ID
+
+    Returns
+    -------
+    Dict[int, float]
+        Key is the block ID and the value is the computed volume
+    )"
   );
   mesh_continuum.def(
     "SetBlockIDFromFunction",
--- a/framework/mesh/mesh_continuum/mesh_continuum.h
+++ b/framework/mesh/mesh_continuum/mesh_continuum.h
@@ -145,7 +145,7 @@ public:
   GlobalCellHandler cells;
 
   /// Compute volume per block IDs
-  void ComputeVolumePerBlockID();
+  std::map<int, double> ComputeVolumePerBlockID() const;
 
 private:
   /// Spatial dimension
--- a/test/python/framework/mesh/tests.json
+++ b/test/python/framework/mesh/tests.json
@@ -25,13 +25,13 @@
     "checks": [
       {
         "type": "KeyValuePair",
-        "key": "[0]  Block ID: 0 Volume:",
+        "key": "Block 0: volume =",
         "goldvalue": 11.25,
         "abs_tol": 1.0e-6
       },
       {
         "type": "KeyValuePair",
-        "key": "[0]  Block ID: 1 Volume:",
+        "key": "Block 1: volume =",
         "goldvalue": 13.75,
         "abs_tol": 1.0e-6
       }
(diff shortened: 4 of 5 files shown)


python/lib/mesh.cc
```diff
@@ -117,10 +118,26 @@ WrapMesh(py::module& mesh)
     )",
     py::arg("file_base_name")
   );
+//  mesh_continuum.def(
+//    "ComputeVolumePerBlockID",
+//    &MeshContinuum::ComputeVolumePerBlockID,
+//    "Compute volume per block ID."
+//  );
```
quocdang1998, line 125: Remove if not needed.

python/lib/mesh.cc
```diff
@@ -117,10 +118,26 @@ WrapMesh(py::module& mesh)
...
   mesh_continuum.def(
     "ComputeVolumePerBlockID",
-    &MeshContinuum::ComputeVolumePerBlockID,
-    "Compute volume per block ID."
+    [](MeshContinuum & self) {
```
quocdang1998, line 128: `MeshContinuum& self`

python/lib/mesh.cc
```diff
@@ -117,10 +118,26 @@ WrapMesh(py::module& mesh)
...
+      for (auto const & kv : cpp_map) {
+        // kv.first is int, kv.second is double
+        py_map[ py::cast(kv.first) ] = py::cast(kv.second);
+      }
+      return py_map;
+    },
+    R"(
+  Compute volume per block ID and return a Python dict mapping
+  each block_id (int) to its total volume (float).
+  )"
```
quocdang1998, line 140: Double-check indentation. It should be:
```
    R"(
    Compute volume per block ID and return a Python dict mapping
    each block ID to its total volume.
    )"
```

quocdang1998 (changes requested):

python/lib/mesh.cc
```diff
@@ -117,10 +118,26 @@ WrapMesh(py::module& mesh)
...
     "ComputeVolumePerBlockID",
-    &MeshContinuum::ComputeVolumePerBlockID,
-    "Compute volume per block ID."
+    [](MeshContinuum & self) {
+      auto cpp_map = self.ComputeVolumePerBlockID();
+      py::dict py_map;
+      for (auto const & kv : cpp_map) {
+        // kv.first is int, kv.second is double
+        py_map[ py::cast(kv.first) ] = py::cast(kv.second);
+      }
```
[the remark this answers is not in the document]
ragusa: now I get this:

/Users/jean.ragusa/repo/opensn/python/lib/mesh.cc:127:15: error: no viable overloaded operator[] for type 'py::dict'
        py_map[block_id] = volume;

framework/mesh/mesh_continuum/mesh_continuum.cc
```diff
@@ -830,6 +830,7 @@ MeshContinuum::ComputeVolumePerBlockID()
       log.Log() << "Block ID: " << id << " Volume: " << std::setprecision(12) << vol << std::endl;
```
quocdang1998, line 830: This line already prints the volume per block ID to the terminal. Should we remove this? @ragusa @wdhawkins

python/lib/mesh.cc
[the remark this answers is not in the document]
quocdang1998: I've forgotten that pybind11 dict cannot take an integer as a key. Can you try:
`py_map[py::cast(block_id)] = volume;`
?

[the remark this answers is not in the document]
ragusa: needed to cast the key. resolved now

quocdang1998 (approved):

framework/mesh/mesh_continuum/mesh_continuum.cc
ragusa: we can keep it for now but i am ok removing it right now. @wdhawkins ?

ragusa: good point. it can be printed in python now, so i removed it.

python/lib/mesh.cc
```diff
@@ -119,8 +119,19 @@ WrapMesh(py::module& mesh)
...
+      for (const auto& [block_id, volume] : cpp_map)
+      {
+        py_map[py::cast(block_id)] = volume;
+      }
+      return py_map;
+    },
+    R"(
+    Compute volume per block ID and return a Python dict mapping
+    each block_id (int) to its total volume (float).
+    )"
```
andrsd, line 134: What is the reason for not using `#include <pybind11/stl.h>`? 

Doco says if you use it, then conversion between STL containers and python types are done automatically [see here](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#automatic-conversion), i.e. in this case `std::map` would be automatically converted into `dict`. Then, you would not need the lambda and you would just bind the member function...

ragusa: I suppose then the answer is simply (aftyer including `<pybind11/stl.h>`):

```
    .def("ComputeVolumePerBlockID",
         &MeshContinuum::ComputeVolumePerBlockID,
         R"(
           Compute volume per block ID and return a std::map<int,double>,
           which will automatically be converted to a Python dict.
         )");

```

@quocdang1998 can you comment on this and on the use of `<pybind11/stl.h>`?

andrsd: Note that the `R"(...)"` is the doco for python, so `std::map` in that context makes no sense. You probably want to say something like:
```
Compute volume per block ID. Returns dict where key is the block ID and 
the value is the computed volume.
```

Or you can take it one step forward and use proper python doco:
```
Compute volume per block ID

Returns
-------
dict
        Key is the block ID and the value is the computed volume
```

ragusa: I have tried this and it works. I have left the older version in comments for now. I will fix this later, after I get feedback from @quocdang1998 . it looks a lot easier with pybind11/stl.h

ragusa: is this good? [should dict be alone on the line?]
```
  mesh_continuum.def(
    "ComputeVolumePerBlockID",
    &MeshContinuum::ComputeVolumePerBlockID,
    R"(
    Compute volume per block ID

    Returns
    -------
    dict
            Key is the block ID and the value is the computed volume
    )"
  );
```

andrsd: I think it should. The code is using the numpy style, see [code sample](https://github.com/Open-Sn/opensn/blob/acb16041898deccf0795c4bbda6f20bbeec7870e/python/lib/post.cc#L55-L58).

See [here](https://www.datacamp.com/tutorial/docstrings-python) for different styles. I am not sure why they do not provide types for return values. Maybe, you want to change `dict` to `volumes`, though. IDK.

andrsd (approved):

quocdang1998 (approved):

PR #616, opened 2025-04-21
Removing LBS MIP DFEM diffusion solver

wdhawkins:


ragusa (approved):
I will have to remember PR 616, for when we re-introduce diffusion as a stadalone solver. I want to make sure that, with theh same mesh + xs, the same results are obtained

PR #591, opened 2025-04-09
Update installation guide in the doc

quocdang1998:
In this PR:

- Installation guide of Linux and MacOs are factored.
- Remove dependencies of Lua.
- Update easy install according to [PR590](https://github.com/Open-Sn/opensn/pull/590).

PR #618, opened 2025-04-22
Remove Lua support

wdhawkins:
~I'll update the install instructions in this PR once #591 is merged.~


tools/lua-input-style.md
ragusa, line 1: the file .stylua.toml in the root should be deleted too

wdhawkins: Fixed.

ragusa (approved):

CMakeLists.txt
```diff
@@ -52,8 +51,7 @@ if(OPENSN_WITH_PYTHON_MODULE)
 endif()
 
 if(OPENSN_WITH_LUA)
-    set(OPENSN_WITH_PYTHON OFF CACHE BOOL "Lua support requested, disabling Python interface" FORCE)
-    set(OPENSN_WITH_PYTHON_MODULE OFF CACHE BOOL "Lua support requested, disabling Python module" FORCE)
+  message(FATAL_ERROR "The option OPENSN_WITH_LUA has been removed and is no longer supported.")
 endif()
```
andrsd, line 55: I don't think you want to keep this. The `OPENSN_WITH_LUA` has been removed, so all the if-statements with `OPENSN_WITH_LUA` should be removed as well. We have the advantage that there was no release yet, so you don't have to go through the deprecation circle and you can just remove stuff.

andrsd (changes requested):

andrsd (approved):

andrsd:
Just noticed that this is removing 1.3M LOCs. Nice :-)

quocdang1998 (approved):

PR #622, opened 2025-04-24
Removing Exit() and replacing calls to it with throw

### LINKED ISSUE #437: error handling
replace occurrence of `Exit(EXIT_FAILURE)` with `throw std::_logic_error(...)`

wdhawkins:
Closes #437

### DIFF
--- a/framework/math/sparse_matrix/sparse_matrix.cc
+++ b/framework/math/sparse_matrix/sparse_matrix.cc
@@ -6,6 +6,7 @@
 #include "framework/logging/log.h"
 #include <iomanip>
 #include <algorithm>
+#include <sstream>
 
 namespace opensn
 {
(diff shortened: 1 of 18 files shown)


framework/math/sparse_matrix/sparse_matrix.cc
```diff
@@ -218,10 +215,7 @@ void
 SparseMatrix::CheckInitialized() const
 {
   if (rowI_values.empty())
-  {
-    log.LogAllError() << "Illegal call to unitialized SparseMatrix matrix.";
-    Exit(EXIT_FAILURE);
-  }
+    throw std::runtime_error("SparseMatrix: Illegal call to unitialized matrix");
```
andrsd, line 218: Can you also fix the typo here?
"unitialized" -> "uninitialized"

framework/mesh/surface_mesh/surface_mesh.cc
```diff
@@ -197,7 +198,7 @@ SurfaceMesh::ImportFromOBJFile(const std::string& fileName, bool as_poly, const
   if (not file.is_open())
   {
     std::ostringstream oss;
-    oss << fname << ": Failed to open file: " << fileName;
+    oss << "SurfaceMesh: Failed to open file " << fileName;
```
andrsd, line 201: If yo do this, can you check that you are not getting a warning on `fname` that would be "assigned into but never used"

andrsd (changes requested):

ragusa (approved):

framework/math/sparse_matrix/sparse_matrix.cc
wdhawkins: Fixed.

framework/mesh/surface_mesh/surface_mesh.cc
wdhawkins: Fixed. I removed the unused `fname` variables.

Issue #506, opened 2025-03-01
making sure old lua functions can still be accessed with the new interface


ragusa:
here is a potential list of old lua functions that are not in the new code. I have some doubts that the list is correct, but I kept some names in it because I'd rather be safe than sorry.
    "ComputeFissionRate",
    "GetPowerFieldFunction",
    "InitializeMaterials",
    "WriteGroupsetAngularFlux",
    "ReadGroupsetAngularFlux",
    "SetPhiFromFieldFunction",
    "AddPointSource",
    "ClearPointSources",
    "AddVolumetricSource",
    "ClearVolumetricSources",
    "ClearResponseSources",
    "AddResponseBuffers",
    "AddResponseSources",
    "GetFieldFunctionList",
    "CreateCustomAngularQuadrature",
    "CreateSphericalProductQuadrature",
    "CreateLineQuadrature",
    "Get1DQuadratureData",
    "GetProductQuadrature",
    "LegendreDerivative",
    "MakeCombined",
    "SetCombined",
    "MakeScaled",
    "SetScalingFactor",
    "ExportToOpenSnFormat",

    "ExportToOBJ",
    "ExportToExodusII",
    "FFInterpolationCreate",
    "PointSense",
    "UnpartitionedMeshFromVTU",
    "UnpartitionedMeshFromPVTU",
    "UnpartitionedMeshFromEnsightGold",
    "UnpartitionedMeshFromWavefrontOBJ",
    "UnpartitionedMeshFromMshFormat",
    "UnpartitionedMeshFromExodusII",

ragusa:
if incorrect, mark as done. if correct, mark as done when done. if unclear, let's use the discussion here.

wdhawkins:
Currently used in regression tests.

ragusa:
> `FFInterpolationCreate` is currently used in our regression tests. `CreateCustomAngularQuadrature` will be deprecated in the new angular quadrature PR.

regarding `FFInterpolationCreate`, I apologixe if I marked it as not in the current lua interface. Honestly, doing that code spelunking, writing a bunch of re and parsing like crazy the old and new code folders was a pain and likely error prone. but, again, better safe then sorry, i hope

wdhawkins:
I've marked a few more that Ifound in the test suite.

ragusa:
I collected the data from commit 95b97902da68d206d0fef47b88b0ad64cd988c3a

here it is, for the record:
Results for variable 'GetPowerFieldFunction':
[...]

PR #631, opened 2025-06-02
Binding `xs.Combine` to `MultiGroupXS::Initialize()`

andrsd:
This binds the MultiGroupXS combining API into python. Test added.

Refs #506

### DIFF
--- a/python/lib/xs.cc
+++ b/python/lib/xs.cc
@@ -3,6 +3,7 @@
 
 #include "python/lib/py_wrappers.h"
 #include "framework/materials/multi_group_xs/multi_group_xs.h"
+#include <pybind11/stl.h>
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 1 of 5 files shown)


test/python/framework/materials/xs_combined.py
```diff
@@ -0,0 +1,26 @@
...
+    xs_1 = MultiGroupXS()
+    xs_1.CreateSimpleOneGroup(sigma_t=1, c=0.5)
+
+    xs_2 = MultiGroupXS()
+    xs_2.CreateSimpleOneGroup(sigma_t=2, c=1./3.)
+
+    xs_combined = MultiGroupXS()
+    combo = [
+        ( xs_1, 0.5 ),
+        ( xs_2, 3.0 )
```
andrsd, line 21: @ragusa I am doing this, because it directly maps onto `vector<pair<XS, float>>`, but if you want a dict here `{ "xs" : xs, "factor": <float> }`, I can do that...

ragusa: no. I am good. I was just provided a suggestion when you asked how the user-facing stuff should look like

python/lib/xs.cc
```diff
@@ -97,6 +98,39 @@ WrapMultiGroupXS(py::module& xs)
...
+        xs_1 = MultiGroupXS()
+        xs_1.CreateSimpleOneGroup(sigma_t=1, c=0.5)
+        xs_2 = MultiGroupXS()
+        xs_2.CreateSimpleOneGroup(sigma_t=2, c=1./3.)
+        xs_combined = MultiGroupXS()
+        combo = [
+            ( xs_1, 0.5 ),
+            ( xs_2, 3.0 )
+        ]
+        xs_combined.Combine(combo)
```
quocdang1998, line 130: In Examples section, you don't need the code-block. Instead, you write it as if you are in a Python console, with each entry start with ``>>>`` (see https://numpydoc.readthedocs.io/en/latest/example.html).

Each command is started with ``>>>``, and continuation line is started with ``...``.

The Example section should be:
```
    Examples
    --------

    >>> xs_1 = MultiGroupXS()
    >>> xs_1.CreateSimpleOneGroup(sigma_t=1, c=0.5)
    >>> xs_2 = MultiGroupXS()
    >>> xs_2.CreateSimpleOneGroup(sigma_t=2, c=1./3.)
    >>> xs_combined = MultiGroupXS()
    >>> combo = [
    ...     ( xs_1, 0.5 ),
    ...     ( xs_2, 3.0 )
    ... ]
    >>> xs_combined.Combine(combo)
```

python/lib/xs.cc
```diff
@@ -97,6 +98,39 @@ WrapMultiGroupXS(py::module& xs)
...
+    {
+      self.Initialize(combinations);
+    },
+    R"(
+    Combine cross-section
+
+    Parameters
+    ----------
+
+    combinations: list[tuple, ...]
```
quocdang1998, line 113: Pls change the type to ``List[Tuple[pyopensn.xs.MultiGroupXS, float]]``

quocdang1998 (changes requested):

quocdang1998 (changes requested):

PR #636, opened 2025-06-04
Add Doxygen comments to MultiGroupXS header

ragusa:
doxygen style comments


ragusa:
@andrs @quocdang1998 : is this what we need for doxygen later on?

PR #642, opened 2025-06-09
Fixing garbage initial k_eff estimates for PI algorithms

wdhawkins:


test/python/modules/linear_boltzmann_solvers/transport_keigen/tests.json
```diff
@@ -207,7 +207,7 @@
         "type": "KeyValuePair",
         "key": "Python k-eigenvalue:",
         "goldvalue": 0.9293377,
-        "abs_tol": 1e-07
```
ragusa, line 210: If the gold value has seven digits, shouldn't the tolerance be 10 to the -7?

wdhawkins: I'll fix the gold value. The gold value has 7 digits, but we only converged to 6, and we fail the seventh with this fix.

wdhawkins: Fixed.

PR #659, opened 2025-06-17
Major directory restructuring. Removing PRK solver.

wdhawkins:
This PR includes a major directory restructuring with the following changes:

1. Rename `modules` to `physics`
2. `physics` has two directories `problems` and `solvers`
3. `problems` contains `diffusion` and `linear_boltzmann` as well as `problem.h` and `problem.cc` from `framework/physics`
4. `solvers` contains the contents of what used to be `solvers`, `acceleration`, and `iterative_methods` from `modules/linear_boltzmann_solvers/lbs_problem` as well as solver and time-stepper sources from `framework/physics`
5. `framework/physics` has been removed
6. The PRK solver has been removed

This is marked as a draft to get everyone opinions. This structure makes sense to me, but may not to anyone else. Making these changes was pretty trivial, so no time lost if we don't like this. Please check out the new directory structure and let me know what you think... good or bad.


loganharbour (approved):
Beautiful find and replace.

Not sure if you guys auto-dismiss reviews on new approvals (given the conflicts)... but approving otherwise

andrsd:
Would it make sense to do:
```
+ framework
|  +- all stuff that physics would build upon
+ physics
|  +- diffusion
|  |  +- all stuff for diffusion
|  +- linear_boltzmann
|    +- all stuff for LBS
```

where diffusion is not diffusion for acceleration, but the MG diffusion solver, et al.

wdhawkins:
> Would it make sense to do:
> 
> ```
> + framework
> |  +- all stuff that physics would build upon
> + physics
> |  +- diffusion
> |  |  +- all stuff for diffusion
> |  +- linear_boltzmann
> |    +- all stuff for LBS
> ```
> 
> where diffusion is not diffusion for acceleration, but the MG diffusion solver, et al.

I'd be ok with that structure as well. My thoughts behind the original structure were along the lines of solvers consuming problems (be it diffusion or lbs).  Now, the issue with this is that some solvers may be specific to lbs or diffusion.

PR #663, opened 2025-06-20
Updating C5G7 tests to use new quadrature API.

wdhawkins:


ragusa (commented):
I do not follow why scattering order 1 is needed in the solver.

PR #666, opened 2025-06-20
Split solution vectors per groupset

andrsd:


wdhawkins:
I haven't had a chance to look at this very closely yet, but I see that c5g7_restart is taking a very long time to run. I suspect that the restart data file needs to be regenerated.

PR #670, opened 2025-06-25
GPU implementation for AAH sweep

quocdang1998:
In this PR, GPU offloading for the sweep is implemented with CUDA 12. Compile the code with CUDA using:

`cmake -DOPENSN_WITH_CUDA=ON ..`

or

`CMAKE_ARGS="-DOPENSN_WITH_CUDA=ON" python setup.py build_ext -i`

Major notes:

- C++ standard migrated to C+20 to accommodate for concept, constexpr algorithm and std::format.
- An header-only external library named `caribou` is added to interface both CUDA and HIP (in the future).
- New classes:
   * `Carrier`: manages memory allocation and transports data to GPU.
   * `AAHD_FLUD`: translates between the contiguous data structure for the FLUD on GPU with the legacy `AHH_FLUD`.
   * `Views`: retrieves information from contiguous data on GPU.
   * Some debug tools for printing the data on GPU, making sure that the data gathering and transferring step are properly executed.
- An option named `use_gpus` is added to `LBSProblem`, allowing user to switch GPU acceleration.

Caveats:

- GPU acceleration is only supported for Ubuntu users with gcc compilers. Clang does not work due to compiler compatibility issues with C++20. MacOS user is not eligible because CUDA has dropped their support for MacOS since CUDA 11.
- Workflows have not yet been restricted to only compilers supporting C++20.
- Not all regressions tests were passed. This implementation requires all cells in the mesh to have at most 8 DOFs. Transport steady tests involving "cycles" use up to 40 DOFs. I'm not sure if this is a bug or an intention.
- Tests were conducted on single-GPU machine. Thus, more MPI ranks means the GPU is over-charged, hampering the performance. I recommend that GPU tests should be conducted with 1 (distinct) GPU per MPI rank.
- In the future, a function allowing user to choose GPU for each MPI rank and get GPU information will be added to the Python interface.

List of failed tests:

- `modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2.py`
- `modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_restart.py`
- `modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_crichardson.py`
- `modules/linear_boltzmann_solvers/transport_steady/transport_3d_5_cycles_2_bicgstab.py`
- `modules/linear_boltzmann_solvers/transport_steady/transport_3d_4_cycles_1.py`


modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_problem/discrete_ordinates_curvilinear_problem.cc
```diff
@@ -357,7 +357,10 @@ DiscreteOrdinatesCurvilinearProblem::ComputeSecondaryUnitIntegrals()
 }
 
 std::shared_ptr<SweepChunk>
-DiscreteOrdinatesCurvilinearProblem::SetSweepChunk(LBSGroupset& groupset)
+DiscreteOrdinatesCurvilinearProblem::SetSweepChunk(LBSGroupset& groupset,
+                                                   size_t max_level_size,
```
wdhawkins, line 361: I think I intended these parameters to be temporary. It might be best to add these as members of `LBSProblem`.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
```diff
@@ -74,6 +74,12 @@ DiscreteOrdinatesProblem::DiscreteOrdinatesProblem(const InputParameters& params
     verbose_sweep_angles_(params.GetParamVectorValue<size_t>("directions_sweep_order_to_print")),
     sweep_type_(params.GetParamValue<std::string>("sweep_type"))
 {
+  if (use_gpus_ && sweep_type_ == "CBC")
+  {
+    std::clog << "Sweep computation on GPUs has not yet been supported for CBC. "
```
wdhawkins, line 79: Instead of `clog`, this should be `Log.Log0Warning`.

external/caribou/.clang-format
ragusa, line 1: @wdhawkins @quocdang1998 has this file moved location or is it a new one? what is caribou

framework/field_functions/field_function_grid_based.cc
```diff
@@ -315,7 +315,7 @@ FieldFunctionGridBased::ExportMultipleToVTK(
       point_data->AddArray(point_array);
       cell_data->AddArray(cell_array);
     } // for component
-  }   // for ff_ptr
+  } // for ff_ptr
```
ragusa, line 318: @wdhawkins @andrsd it is not hte first time that we come cross the issue of comment alignments in nested loops. It seems that a clang default is to not do what we do. Should we abandon our alignment request. I personally prefer when the comments are not all on the same column

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_problem/sweep_chunks/aah_sweep_chunk_rz.cc
```diff
@@ -117,7 +117,7 @@ AAHSweepChunkRZ::Sweep(AngleSet& angle_set)
     // Loop over angles in set (as = angleset, ss = subset)
     const int ni_deploc_face_counter = deploc_face_counter;
     const int ni_preloc_face_counter = preloc_face_counter;
-    const std::vector<size_t>& as_angle_indices = angle_set.GetAngleIndices();
+    const std::vector<std::uint32_t>& as_angle_indices = angle_set.GetAngleIndices();
```
ragusa, line 120: why not `size_t` anymore ?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cc
```diff
@@ -83,7 +101,7 @@ AAHSweepChunk::Sweep(AngleSet& angle_set)
     // Loop over angles in set (as = angleset, ss = subset)
     const int ni_deploc_face_counter = deploc_face_counter;
     const int ni_preloc_face_counter = preloc_face_counter;
-    const std::vector<size_t>& as_angle_indices = angle_set.GetAngleIndices();
+    const std::vector<std::uint32_t>& as_angle_indices = angle_set.GetAngleIndices();
```
ragusa, line 104: `size_t` ?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/cbc_sweep_chunk.cc
```diff
@@ -102,7 +102,7 @@ CBCSweepChunk::Sweep(AngleSet& angle_set)
 
   // as = angle set
   // ss = subset
-  const std::vector<size_t>& as_angle_indices = angle_set.GetAngleIndices();
+  const std::vector<std::uint32_t>& as_angle_indices = angle_set.GetAngleIndices();
```
ragusa, line 105: `size_t` ?

test/python/modules/cfem_diffusion/c_diffusion_2d_1a_linear.py
```diff
@@ -17,7 +17,7 @@
     from pyopensn.solver import DiscreteOrdinatesProblem, SteadyStateSolver
     from pyopensn.fieldfunc import FieldFunctionGridBased
     from pyopensn.fieldfunc import FieldFunctionInterpolationLine, FieldFunctionInterpolationVolume
-    from pyopensn.settings import EnableCaliper
+    from pyopensn.context import EnableCaliper
```
ragusa, line 20: @quincy-huhn98 : FYI, since you are moving them to the App. Look at this and advise.

external/caribou/.clang-format
quocdang1998: The content was the same. Only the namespace has changed. I think that naming the external library as something specific is more memorable than a generic name like "gpu".

framework/field_functions/field_function_grid_based.cc
andrsd: IIRC, the problem is "caused" by clang-format-14. The newer versions may have more flags to control these cases.

I, personally, don't like these comments at all. Any modern editor can find matching curly brackets, but some people find those useful. This is one of the downsides of using automatic code formatters...

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -0,0 +1,356 @@
...
+#include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aah_fluds.h"
+
+namespace opensn
+{
+
+template <typename T>
+inline static char*
+write_to(char* dest, const T& src)
+{
+  T* dest_ = reinterpret_cast<T*>(dest);
```
andrsd, line 14: Could you use something like `dest_typed`, the trailing `_` is reserved for private member variables.
But it is not a big deal...

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -0,0 +1,356 @@
...
+  bool is_boundary_face = not cell.faces[f].has_neighbor;
+  if (is_local_face)
+  {
+    ++in_face_counter;
+  }
+  else if (not is_boundary_face)
+  {
+    ++preloc_face_counter;
+  }
+  // record pointer to FLUDs of a given asngle index and a given face node
```
andrsd, line 48: typo: "asngle" -> "angle"

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -0,0 +1,356 @@
...
+  out_face_counter++;
+  bool is_local_face = cell_transport_view.IsFaceLocal(f);
+  bool is_boundary_face = not cell.faces[f].has_neighbor;
+  bool is_reflecting_boundary_face =
+    (is_boundary_face and angle_set.GetBoundaries()[cell.faces[f].neighbor_id]->IsReflecting());
+  if (not is_boundary_face and not is_local_face)
+  {
+    ++deploc_face_counter;
+  }
+  // record pointer to FLUDs of a given asngle index and a given face node
```
andrsd, line 106: typo: "asngle" -> "angle"

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -0,0 +1,356 @@
...
+  for (std::size_t as_ss_idx = 0; as_ss_idx < nangles; ++as_ss_idx)
+  {
+    std::size_t direction_num = angle_set.GetAngleIndices()[as_ss_idx];
```
andrsd, line 109: `GetAngleIndices()` returns now `vector<uint32_t>`, this is stored as `std::size_t`, but then passed into `PsiBoundary` as `unsigned int`. Not that it was better before, but will be unified into `uint32_t` based on the [discussion](https://github.com/Open-Sn/opensn/pull/670/files#r2173739858)?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -0,0 +1,356 @@
...
+
+AAHD_FLUDS::AAHD_FLUDS(LBSProblem& lbs_problem,
+                       const LBSGroupset& group_set,
+                       AngleSet& angle_set,
+                       bool is_surface_source_active)
+{
+  // get SPDS
+  const SPDS& spds = angle_set.GetSPDS();
+  const std::vector<std::vector<int>>& levelized_spls = spds.GetLevelizedLocalSubgrid();
+  // get FLUD
```
andrsd, line 143: typo: "FLUD" -> "FLUDS" - IIRC, DS in FLUDS stands for Data Structure (S is not a denotation of plural).

modules/linear_boltzmann_solvers/lbs_problem/device/debug_tool/print_xs.h
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2025 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/carrier/total_xs_carrier.h"
+
+namespace opensn::debug
```
andrsd, line 8: This is a nested namespace. Earlier in the project, it was discussed to remove nested namespaces (see #44), but we did not codified it in the code standard. This may be a leftover from some other work, so just bringing this up...

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.h
```diff
@@ -33,13 +34,13 @@ class LBSProblem : public Problem
   explicit LBSProblem(const std::string& name, std::shared_ptr<MeshContinuum> grid);
 
   /// Input parameters based construction.
-  explicit LBSProblem(const InputParameters& params);
+  explicit LBSProblem(const InputParameters& pManagerarams);
```
andrsd, line 37: change `pManagerarams` back into `params`

andrsd (changes requested):
I suggest that the formatting part of this PR will go in its own PR. It will make it easier to review the GPU-related code.

modules/linear_boltzmann_solvers/lbs_problem/device/debug_tool/print_mesh.h
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2025 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/carrier/mesh_carrier.h"
+
+namespace opensn::debug
```
ragusa, line 8: Ideas of namespaces that are a little more explicit:

- gpudebug
- devdebug
- debuggpu
- debugdev
- device_dbg
- gpu_dbg

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
```diff
@@ -0,0 +1,533 @@
...
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/carrier/outflow_carrier.h"
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/carrier/quadrature_carrier.h"
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/storage.h"
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/view/mesh_view.h"
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/view/quadrature_view.h"
+#include "modules/linear_boltzmann_solvers/lbs_problem/device/view/xs_view.h"
+#include "framework/mesh/mesh_continuum/mesh_continuum.h"
+#include "caliper/cali.h"
+#include "caribou/caribou.h"
+#include <format>
```
wdhawkins, line 19: Since we only use this here - and only for one message. Let's remove it so that we can support a wider range of compilers (such as those on the LC machines).

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
```diff
@@ -0,0 +1,533 @@
...
+  }
+}
+
+void
+AAHSweepChunk::GPUSweep(AngleSet& angle_set)
+{
+  // check arguments
+  if (max_num_cell_dofs_ > max_dof)
+  {
+    std::clog << std::format("GPU sweep expects max number of DOF to be {}, but got a mesh with "
```
wdhawkins, line 405: We don't use `clog`. This should either use `Log0Warning` or just `throw`.

modules/linear_boltzmann_solvers/lbs_problem/device/carrier/mesh_carrier.cu
```diff
@@ -0,0 +1,201 @@
...
+      alloc_size += sizeof(std::uintptr_t);
+      // normal vector
+      alloc_size += 3 * sizeof(double);
+      // M_surf matrix (num_face_nodes x num_face_nodes)
+      std::size_t num_face_nodes = face_node_mappings[f].size();
+      alloc_size += num_face_nodes * num_face_nodes * sizeof(double);
+      // IntF_shape_I
+      alloc_size += num_face_nodes * sizeof(double);
+      // cell mapping (allocate up to the smallest even number to ensure 64-bit alignment)
+      num_face_nodes = (num_face_nodes + 1) & ~1zu;
```
wdhawkins, line 65: For portability, we should change `~1zu` to something like `~static_cast<size_t>(1)`.

external/caribou/cuda/atomic.hpp
```diff
@@ -0,0 +1,26 @@
...
+    unsigned long long int old = *address_as_ull, assumed;
+    do {
+        assumed = old;
+        old = ::atomicCAS(address_as_ull, assumed, ::__double_as_longlong(val + ::__longlong_as_double(assumed)));
+    } while (assumed != old);
+    return ::__longlong_as_double(old);
+}
+#else
+__device__ inline double atomic_add(double* address, double val) {
+    return ::atomicAdd(address, val);
```
wdhawkins, line 22: For better compiler support, we should change this to:
```
#else 
  return atomicAdd(address, val);
#endif
```

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
quocdang1998: Fixed.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
quocdang1998: Fixed. I changed `dest` to `buffer` and `dest_` to `dest` (for destination data).

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
quocdang1998: Thanks for this remark. I fixed it in the latest commit.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
quocdang1998: Thanks a lot for this remark. I forgot to change this to `std::uint32_t`. `direction_num` is now declared as `const std::uint32_t&` in the latest commit.

As far as I can tell, on Ubuntu 24.04 with gcc 13.3, `unsigned int` is 4 bytes. Thus, both `unsigned int` and `std::uint32_t` are the same, and we don't have to worry about incompatibility.

I think that the unification of integer type deserve its own PR. We should fix it in the future.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
quocdang1998: Fixed. Thanks for the clarification.

I also renamed `flud_(level|cell|face)` to `fluds_$1`.

modules/linear_boltzmann_solvers/lbs_problem/device/debug_tool/print_xs.h
quocdang1998: I changed the namespace to `device_dbg`.

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.h
quocdang1998: Fixed.

modules/linear_boltzmann_solvers/lbs_problem/device/debug_tool/print_mesh.h
quocdang1998: I chose `device_dbg`.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
quocdang1998: Fixed. `std::format` usage was removed.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
quocdang1998: Fixed. `std::clog` was replaced with `log.Log0Warning()`.

modules/linear_boltzmann_solvers/lbs_problem/device/carrier/mesh_carrier.cu
quocdang1998: Fixed.

external/caribou/cuda/atomic.hpp
quocdang1998: Fixed.

wdhawkins (approved):

PR #679, opened 2025-07-06
Convert CMake-doc to Make-doc

quocdang1998:
In this PR, the documentation build system is transitioned from CMake-based to Make-based, which:

- removes the need to manually add image file copying commands to the build script.
- improves compatibility with Sphinx 8.
- simplifies the document debugging process.
- enable the generation of the PDF version of the documentation (using `make latexpdf`) alongside the HTML version (`make html`).

Note: the generation of the PDF version was tested on `Ubuntu 24.04`, using TexLive version 2025.06.01.


ragusa (commented):
I wonder if some rst files that show and deleted and the. New in another folder should actually show up as moved?

ragusa (approved):

.github/workflows/gh-pages.yml
```diff
@@ -17,10 +17,9 @@ jobs:
         run: |
           module load python3/3.12.3 opensn/gcc/14 doxygen/1.13.2 pandoc/3.6.4
           mkdir build && cd build
-          cmake -DOPENSN_WITH_DOCS=ON -DOPENSN_WITH_PYTHON_MODULE=ON ..
-          make -j __init__ doc
+          python setup.py build_ext --inplace
```
andrsd, line 20: There might be a problem here. The line above does: `cd build`, so when `python setup.py ...` gets to be executed it would fail. I think you assume to build from the repo root?

Another problem might be: Does `python setup.py build_ext --inplace` build the doco? It seems you changed that to roughly `cd doc; make html`. The original command was doing `make doc` after cmake, which build the html pages. This seems to be building just the module. Unless I am horribly missing something...

andrsd (commented):
So no more doxygen doco generated via breathe?

.github/workflows/gh-pages.yml
quocdang1998: You were absolutely correct. I forgot to add a `cd doc` and `make html`. Thanks for pointing this out.

quocdang1998: @andrsd We don't include the C++ documention in the Sphinx code (for now).

@ragusa Maybe when we finished the K-eigen solver, we can go back and add developper documentation for the C++ source (using Doxygen) and then add the C++ documention to the current Sphinx documentation later.

quocdang1998: Thanks for the comment. I would refer to the second option (generate the doxygen doco inside repo's doco).

I made a prototype 3 years ago. I added `doxygen ./Doxyfile` to `Makefile` to generate a `doc/source/xml` folder (ignored by Git). `breathe` is configured to automatically read this folder and generate Sphinx documentations for C++ code. I can apply the same implementation for the current doc once the C++ doxygen doc is completed.

wdhawkins (approved):

PR #688, opened 2025-07-17
Updating CMake CUDA target

wdhawkins:


CMakeLists.txt
```diff
@@ -173,7 +174,7 @@ endif()
 add_library(libopensn SHARED ${LIBOPENSN_SRCS})
 
 if (OPENSN_WITH_CUDA)
-    target_compile_options(libopensncuda PRIVATE --expt-extended-lambda --expt-relaxed-constexpr)
```
quocdang1998, line 176: Please use
`set_property(TARGET libopensncuda PROPERTY CUDA_STANDARD 20)`
or
`target_compile_features(libopensncuda PUBLIC cuda_std_20)`
instead.

quocdang1998 (changes requested):

CMakeLists.txt
wdhawkins: I tried `set_property(TARGET libopensncuda PROPERTY CUDA_STANDARD 20)`, but wasn't successful. Let me try `cuda_std_20` and see if that works.

wdhawkins: `target_compile_features(libopensncuda PUBLIC cuda_std_20)` works if we bump our minimum CMake version to 3.25.2. Fixed.

ragusa (approved):

andrsd (approved):

PR #693, opened 2025-07-21
Moving sweep boundaries into `DiscreteOrdinatesProblem`

andrsd:


test/python/modules/linear_boltzmann_solvers/transport_keigen/keigenvalue_transport_2d_1b_qblock_cbc.py
```diff
@@ -57,6 +57,6 @@
         },
         sweep_type="CBC",
     )
-    k_solver = NonLinearKEigenSolver(lbs_problem=phys)
+    k_solver = NonLinearKEigenSolver(do_problem=phys)
```
ragusa, line 60: From the user's perspective, it feels strange to have to supply a value to a keyword that reads as "do (a) problem". I know we mean discrete ordinates, but at the user level, I am not certain this is how we should call it/expose it. @wdhawkins thoughts?

wdhawkins: I agree. We could just drop the prefix entirely and go with `problem=`.

modules/linear_boltzmann_solvers/lbs_problem/acceleration/nl_keigen_acc_residual_func.cc
```diff
@@ -18,7 +18,7 @@ NLKEigenAccResidualFunction(SNES snes, Vec phi, Vec r, void* ctx)
 
   auto& diff_solver = nl_context_ptr->diff_solver;
 
-  auto& lbs_problem = nl_context_ptr->lbs_problem;
+  auto& lbs_problem = nl_context_ptr->do_problem;
```
ragusa, line 21: Why is the result still called `lbs_problem`?

andrsd: @ragusa, @wdhawkins  When fixing this, I noticed that there are still a few `lbs_problem` user params. I renamed the `do_problem` to `problem`, so should I also rename `lbs_problem` input parameter to `problem`?

PR #695, opened 2025-07-26
Fix bug in github workflow for building documentation with make

quocdang1998:


ragusa (approved):
@wdhawkins i suppose this choice of alias/name on the CI machine is dependent on how you set it up.

PR #701, opened 2025-07-31
Remove event system

andrsd:
This removes:
- event systems
- post processors


andrsd:
This is built on top of `remove-object` branch (#698).

4 tests are failing. This is expected, since they are using post processors and those were removed. Would should the tests check instead?

andrsd:
Tests are changed and this is now ready for review.

ragusa (approved):

PR #705, opened 2025-08-06
Moving power field function into KEigenSolver

andrsd:
KEgienSolver is a new intermediate class to house common functionality for k-eigenvalue solvers.


wdhawkins:
> I do not think it makes 100% sense to have power inside the Keigensolver. A subcritical system can be computed using a source driven solver and may need the power to be computed too.

That's a good point. I hadn't considered that when I asked David to move this to keigensolver. Maybe what we really need to do is collect all the field function routines in one place.

PR #723, opened 2025-08-15
Adding boundary conditions as an optional solver input parameter

wdhawkins:
Boundary conditions are now passed to the solver via the lbs_solver block and not the lbs_options block. BCs are still an optional parameter. If a BC block is not supplied, vacuum boundary conditions are used. This PR leaves the `clear_boundary_conditions` parameter in lbs_options (although it's unclear to me how useful this is).

### DIFF
--- a/doc/source/tutorials/lbs/mg_keigen/pincell_example.ipynb
+++ b/doc/source/tutorials/lbs/mg_keigen/pincell_example.ipynb
@@ -251,7 +251,7 @@
     "    {\"block_ids\": [3], \"xs\": xs_list[3]}\n",
     "]\n",
     "\n",
-    "phys = DiscreteOrdinatesProblem(mesh=grid, num_groups=num_groups, groupsets=group_sets, xs_map=xs_mapping, scattering_order=scat_order)\n",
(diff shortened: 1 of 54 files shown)


andrsd (approved):

wdhawkins:
> Need to check the openSn example repo, jupyter notebook, too. For the clear bc method, it could be used in the adjoint mode and we should ping @zhardy-lanl

I'll make sure all notebooks and examples are update if we decide to merge this PR. But...

I think the bigger question is: Do we want to do this? Having looked late last night at refactoring sources in the same way, I now understand why these are in options. It is all adjoint related, and we may opt to leave both BCs and source in options for now. With this commit (and with the upcoming analogous source commit), we lose the ability to clear boundary conditions (and sources) and then reset them in the same solver instance to something new using the `SetOptions` functionality. I think it's too messy to have them as both solver parameters and options. Right now, I think it's probably best to close this PR and mark the issue as `won't fix`. In the future, maybe we do move them out of lbs_options and also provide user-facing methods to reset them on an existing solver instance. Something like `SetBoundaryConditions` and `SetSources`. This seems like a v2.0 change to me.

wdhawkins:
If we are ok with the `SetBoundaryConditions` and `SetSources` user-facing options, we could delay 1.0 for a few days while we implement this. @quocdang1998 could probably do it fairly quickly.  Thoughts?

PR #724, opened 2025-08-15
Removing unused spatial_discretization LBS option

wdhawkins:


ragusa (commented):
Shouldn't things like this be removed from inputs then?

`  "spatial_discretization": "pwld",`

wdhawkins:
> Shouldn't things like this be removed from inputs then?
> 
> ` "spatial_discretization": "pwld",`

Yes. I intended to do that. Fixed.

ragusa:
> > Shouldn't things like this be removed from inputs then?
> 
> > 
> 
> > ` "spatial_discretization": "pwld",`
> 
> 
> 
> Yes. I intended to do that. Fixed.

Why did the regression tests pass early with what should have been an unknown option then?

wdhawkins:
> Why did the regression tests pass early with what should have been an unknown option then?

We don't check for valid options in the option block. We loop through the options and perform some function if they are present, but ignore them otherwise.

ragusa:
> > Why did the regression tests pass early with what should have been an unknown option then?
> 
> 
> 
> We don't check for valid options in the option block. We loop through the options and perform some function if they are present, but ignore them otherwise.

Wouldn't it be good at add an error for unrecognized options? Maybe you can create an issue for that.

ragusa (approved):

PR #738, opened 2025-08-21
Updating quadrature documentation

wdhawkins:


ragusa (commented):
Rather than simply "Scattering order.", do we want to say "Maximum scattering order supported by the angular quadrature." or "Maximum scattering order that the angular quadrature will enable."

quocdang1998 (approved):

andrsd (approved):

PR #745, opened 2025-08-27
Add set/clear sources and boundary conditions methods to Python interface

quocdang1998:
This PR adds the methods setting/clearing sources and boundary conditions methods to Python interface.

Note: boundary conditions and sources arguments in ``OptionBlock`` to be removed (with regression tests to be updated).

### DIFF
--- a/python/lib/solver.cc
+++ b/python/lib/solver.cc
@@ -418,6 +418,94 @@ WrapLBS(py::module& slv)
     )",
     py::arg("file_base")
   );
+  lbs_problem.def(
+    "SetPointSources",
+    [](LBSProblem& self, py::kwargs& params)
+    {
+      for (auto [key, value] : params)
+      {
+        auto c_key = key.cast<std::string>();
+        if (c_key == "clear_point_sources")
+          self.ClearPointSources();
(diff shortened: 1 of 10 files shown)


python/lib/solver.cc
```diff
@@ -418,6 +418,67 @@ WrapLBS(py::module& slv)
...
+    py::arg("volumetric_source")
+  );
+  lbs_problem.def(
+    "ClearBoundaries",
+    &LBSProblem::ClearBoundaries,
+    R"(
+    Clears all the boundary conditions from the solver.
+    )"
+  );
+  lbs_problem.def(
```
wdhawkins, line 458: Instead of `SetBoundaryOptions`, can we call this `SetBoundaries`?

wdhawkins:
Before this is committed, the current processing that is done in `SetOptions` needs to be removed, and the regression tests using the functionality need to be updates to use the new methods.

wdhawkins:
It looks like the boundary/source processing that is done in `SetOptions` still needs to be removed. Users should have to use the new `Set` routines instead of `SetOptions` (this applies to `lbs_problem.cc` and `python/lib/solver.cc`).

PR #746, opened 2025-08-27
Implemented OpenFOAM polyMesh reader with regression test for Hex, Te…

kitamuraa24:
…ts, Wedge, and Pyr elements.


test/python/framework/mesh/read_openfoam_meshes.py
```diff
@@ -0,0 +1,46 @@
...
+    grid = meshgen.Execute()
+    grid.ExportToPVTU("Pyr")
+
+    # Test Wedge
+    meshgen = FromFileMeshGenerator(
+        filename="openfoam_meshes/Wedge"
+    )
+    grid = meshgen.Execute()
+    grid.ExportToPVTU("Wedge")
+
```
andrsd, line 43: General notes on the test. 
- You have a log statement in the OpenFOAM reader, so you may want to match that rather then the `Global cell count`
- You should probably match both the cell and node count
- It would probably make sense to split this into 4 tests and each one would be matching the cell/node count
- Why is the export to VTU needed?

framework/mesh/io/foam_io.cc
```diff
@@ -0,0 +1,655 @@
...
+        return false;
+      }
+      else
+      {
+        face_verts.emplace_back(std::move(fv));
+        return true;
+      }
+    };
+    ReadFoamList(in, nfaces, read_face, fname, "faces");
+    if (face_verts.size() != nfaces)
```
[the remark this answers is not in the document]
kitamuraa24: Yes you are right; this was me being sloppy sorry about that.

ragusa:
redo the original text of the PR. Currently, it reads "…ts, Wedge, and Pyr elements."
something got cropped

framework/mesh/io/foam_io.cc
```diff
@@ -0,0 +1,629 @@
...
+    { // // ...
+      s.get();
+      s.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+      continue;
+    }
+    if (c2 == '*')
+    { // /* ... */
+      s.get();
+      int prev = 0, cur = 0;
+      while (s and not(prev == '*' and (cur = s.get()) == '/'))
```
wdhawkins, line 39: This looks like it might result in a hang if prev != '*'.

kitamuraa24:
Give me a moment; I found another bug that I am currently working on fixing that was not present earlier.

kitamuraa24:
Fixed the issue. Redid the ReadCellZones function

kitamuraa24:
Hi all, just wondering if everything is good to merge?

wdhawkins:
Can you rebase this on main and extend one of your tests to run a very simple problem? Just to make sure that boundary conditions, etc. work?

PR #747, opened 2025-08-28
Don't store node-averaged values as elemental field for discontinuous solutions

andrsd:


ragusa (approved):
This change may affect the OpenSnExamples repo as well

wdhawkins (approved):

PR #762, opened 2025-09-19
Implementation of Galerkin Quadrature methods and additional tutorials

Aggie04:
This PR will add the Galerkin Quadrature methods 1 and 3 into OpenSn to offer another option when forming the scattering source. These scattering operators are formally expressed in literature. 

In addition to this, 5 tutorials have been added --
Triangular quadrature sets and Lebedev quadrature sets have been given tutorials that are inline with the Product and SLDFE tutorials, explaining the requirements for them and an example of their angular grid for a given order. 
```operator_methods.ipynb``` exists as an introduction to the different ways the discrete-to-moment flux maps can be constructed. A quick aside is also mentioned here that the D2M matrix is stored as the transpose of the literature-defined D2M. 
Two tutorials (one with highly forward peaked Henyey-Greenstein scattering, the other with pure mu=1 delta scattering) were added. The first one shows that a large cross section expansion is needed to converge the legendre moments, but even a truncated expansion is capable of yeilding good results if the solution is well represented by harmonics. The second tutorial exists to showcase the major differences between the Standard method and the Galerkin-Quadrature methods, which differences are exacerbated in highly-forward peaked problems with the full harmonic expansion. 

All of these tutorials I added under the angular quadrature tutorials, since they are designed to showcase certain aspects directly related to the angular quadrature class.

### CODE REVIEW

File: python/lib/aquad.cc [hunk 1]
```diff
@@ -79,6 +79,79 @@ WrapQuadrature(py::module& aquad)
...
+    &AngularQuadrature::GetOperatorConstructionMethod,
+    R"(
+    Get the current operator construction method.
+
+    Returns
+    -------
+    int
+        Current construction method: 0=Standard, 1=Galerkin Method 1, 2=Galerkin Method 2
+    )"
+  );
```
quocdang1998, line 106: Is this method necessary?

python/lib/aquad.cc
```diff
@@ -79,6 +79,79 @@ WrapQuadrature(py::module& aquad)
...
+    R"(
+    Set the method used to construct the D2M and M2D operators.
+    
+    Parameters
+    ----------
+    method : int
+        Construction method: 0=Standard, 1=Galerkin Method 1, 2=Galerkin Method 2
+    )",
+    py::arg("method")
+  );
```
quocdang1998, line 94: Do not use integer for enum. Classic Python uses strings for options (like [this](https://numpy.org/doc/2.3/reference/generated/numpy.ndarray.copy.html)).

Instead, you should create an ``std::map<std::string, OperatorConstructionMethod>`` outside of this function (see [this example](https://github.com/Open-Sn/opensn/blob/main/python/lib/fieldfunc.cc#L22)) and use ``std::map::at``. It will be more explicit for users and more type-safe as well.

python/lib/aquad.cc
```diff
@@ -79,6 +79,79 @@ WrapQuadrature(py::module& aquad)
...
+        py::list py_row;
+        for (double val : row) {
+          py_row.append(val);
+        }
+        result.append(py_row);
+      }
+      return result;
+    },
+    "Get the moment-to-discrete operator as a list of lists."
+  );
```
quocdang1998, line 150: Similarly, these functions are not necessary to be exported to Python.

And if you really need to export it, you should export this as a NumPy array The results of these methods are matrices, not lists of lists.

python/lib/aquad.cc
```diff
@@ -79,6 +79,79 @@ WrapQuadrature(py::module& aquad)
...
+  py::enum_<OperatorConstructionMethod>(aquad, "OperatorConstructionMethod")
+    .value("Standard", OperatorConstructionMethod::Standard)
+    .value("GalerkinMethodOne", OperatorConstructionMethod::GalerkinMethodOne)
+    .value("GalerkinMethodTwo", OperatorConstructionMethod::GalerkinMethodTwo);
```
quocdang1998, line 154: pls replace enum usage with explicit string usage for option.

framework/math/quadratures/angular/angular_quadrature.h
```diff
@@ -50,30 +57,54 @@ class AngularQuadrature
...
+
+  explicit AngularQuadrature(
+    AngularQuadratureType type,
+    unsigned int dimension,
+    unsigned int scattering_order,
+    OperatorConstructionMethod method = OperatorConstructionMethod::Standard)
+    : type_(type),
+      dimension_(dimension),
+      scattering_order_(scattering_order),
+      construction_method_(method)
```
quocdang1998, line 70: Since you add the `construction_method_` to the constructor, don't forget to update the constructor of derived classes as well.

quocdang1998 (changes requested):

python/lib/aquad.cc
Aggie04: I had this mostly so I can bug fix python scripts and see what was being passed where. It has been removed.

framework/math/quadratures/angular/harmonic_selection_rules.h
```diff
@@ -0,0 +1,89 @@
...
+    size_t num_angles = 0;
+
+    // Quadrature-specific order parameter
+    // For Lebedev: Lebedev order (3, 5, 7, 9, etc.)
+    // For SLDFESQ: Uniform Refinement level
+    unsigned int quadrature_order = 0;
+
+    // Info needed to Product + Triangular
+    unsigned int n_polar = 0;     // Number of positive-mu polar points
+    unsigned int n_azimuthal = 0; // Number of azimuthal points in (0, pi/2)
```
wdhawkins: `n_azimuthal` is the full azimuthal count.

framework/math/quadratures/angular/harmonic_selection_rules.cc
```diff
@@ -0,0 +1,674 @@
...
+{
+  // 3D Lebedev sets have specific numbers of points (full sphere)
+  static const std::map<size_t, int> angle_to_order_3d =
+    {
+      {6, 3},   // Order 3
+      {14, 5},  // Order 5
+      {26, 7},  // Order 7
+      {38, 9},  // Order 9
+      {50, 11}, // Order etc...
+      {74, 13}, // FIX ME ---- ONLY 62 / 74 ORTHOGONAL VECTORS FOUND
```
wdhawkins: For G1, this (and 25 and 27) won't work... right? Do we test for that?

framework/math/quadratures/angular/angular_quadrature.cc
```diff
@@ -16,97 +21,435 @@ AngularQuadrature::MakeHarmonicIndices()
...
+                                : scattering_order_;
+    params.num_angles = abscissae.size();
+    params.quadrature_order = quadrature_order_;
+    params.n_polar = n_polar_;
+    params.n_azimuthal = n_azimuthal_;
+
+    // Get rule-selected harmonics
+    m_to_ell_em_map_ = HarmonicSelectionRules::SelectHarmonics(params);
+
+    // For GALERKIN_ONE, update scattering_order_ to the maximum ℓ in the selected harmonics
```
andrsd: Could you either spell it with normal ascii codes (i.e. not use unicode), or maybe put the ascii spelling into parentheses after the character. It is hard to read that this character actually is.

framework/math/quadratures/angular/angular_quadrature.cc
```diff
@@ -16,97 +21,435 @@ AngularQuadrature::MakeHarmonicIndices()
...
-  else if (dimension_ == 2)
-    for (auto ell = 0; ell <= scattering_order_; ++ell) // NOLINT
-      for (auto m = -ell; m <= ell; m += 2)
-        m_to_ell_em_map_.emplace_back(ell, m);
-  else if (dimension_ == 3)
-    for (auto ell = 0; ell <= scattering_order_; ++ell) // NOLINT
-      for (auto m = -ell; m <= ell; ++m)
-        m_to_ell_em_map_.emplace_back(ell, m);
+  // For Galerkin methods with Product Quadrature, use rule-based selection
+  if ((construction_method_ == OperatorConstructionMethod::GALERKIN_ONE ||
```
andrsd: `||` -> `or`

framework/math/quadratures/angular/angular_quadrature.cc
```diff
@@ -16,97 +21,435 @@ AngularQuadrature::MakeHarmonicIndices()
...
 }
 
 void
 AngularQuadrature::BuildDiscreteToMomentOperator()
 {
-  d2m_op_.clear();
-
   const size_t num_angles = abscissae.size();
-  const unsigned int num_moms = m_to_ell_em_map_.size();
+  const size_t num_moms = m_to_ell_em_map_.size();
```
andrsd: `const auto num_moms = ...`

framework/math/quadratures/angular/angular_quadrature.cc
```diff
@@ -16,97 +21,435 @@ AngularQuadrature::MakeHarmonicIndices()
...
-    auto& row = d2m_op_[n];
+    case OperatorConstructionMethod::STANDARD:
+    {
+      d2m_op_.clear();
+      d2m_op_.assign(num_angles, std::vector<double>(num_moms, 0.0));
+
+      for (size_t n = 0; n < num_angles; ++n)
+      {
+        const auto& cur_angle = abscissae[n];
+        double w = weights[n];
```
andrsd: `auto w = ...`

framework/math/quadratures/angular/angular_quadrature.cc
```diff
@@ -16,97 +21,435 @@ AngularQuadrature::MakeHarmonicIndices()
...
+
+        for (size_t m = 0; m < num_moms; ++m)
+        {
+          const auto& ell_em = m_to_ell_em_map_[m];
+          double value = Ylm(ell_em.ell, ell_em.m, cur_angle.phi, cur_angle.theta);
```
andrsd: `auto value = ...`

framework/math/quadratures/angular/harmonic_selection_rules.h
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  /**
+   * \brief Select harmonics using predetermined mathematical rules
```
andrsd: Remove the `\brief` tag

framework/math/math.h
```diff
@@ -43,48 +43,40 @@ enum class NormType : int
...
- * Supplying a random number between 0 and 1 should indicate sampling one
- * of the bins 0,1,5 or 8. The most inefficient way to do this is to
- * linearly loop through the cdf and check \f$ cdf_{i-1} \ge \theta < cdf_i \f$.
- *  An optimized version of this sampling would be to perform a recursive
- *  block search which starts with a course view of the cdf and then gradually
- *  refines the view until the final linear search can be performed.*/
+/// Sample a Cumulative Distribution Function (CDF) given a probability.
+///
+/// The supplied vector should contain the upper bin boundary for each bin and will return
+/// the bin associated with the bin that brackets the supplied probability.
```
quocdang1998: For multi-line doc, use `/** */` style comment.

PR #774, opened 2025-09-30
Removing extraneous print statements from c5g7_restart.py

wdhawkins:
The print statements were being printed from all ranks and clobbering the elapsed time output that the regression test script depends on.


quocdang1998 (approved):

ragusa:
why merge when tests fail?

PR #784, opened 2025-10-02
Scattering order is stored as `unsigned int`

andrsd:
Refs #90

### DIFF
--- a/framework/materials/multi_group_xs/openmc_import.cc
+++ b/framework/materials/multi_group_xs/openmc_import.cc
@@ -59,7 +59,7 @@ MultiGroupXS::LoadFromOpenMC(const std::string& file_name,
   mgxs.temperature_ = temperature;
 
   // Scattering order
-  if (not H5ReadGroupAttribute<size_t>(file, dataset_name, "order", mgxs.scattering_order_))
+  if (not H5ReadGroupAttribute<unsigned int>(file, dataset_name, "order", mgxs.scattering_order_))
     throw std::runtime_error("Failure reading \"order\" from " + file_name);
 
   // Scattering shape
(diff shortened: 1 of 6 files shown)


wdhawkins (approved):

framework/materials/multi_group_xs/openmc_import.cc
```diff
@@ -103,7 +103,7 @@ MultiGroupXS::LoadFromOpenMC(const std::string& file_name,
     int fidx = 0;
     for (size_t gp = 0; gp < mgxs.num_groups_; ++gp)
       for (int g = g_min[gp]; g <= g_max[gp]; ++g)
```
ragusa, line 105: any idea why those are `size_t` and `int` in those lines? should we add to #90 ?

andrsd: That's already on my radar and I will send a PR to change number of groups into `size_t`, i.e. I already have a fix this this.

andrsd: I checked my patch for fixing number of groups. I am updating the code such that it would be stored as `size_t`, since `LBSProblem` stores groups in `std::vector`. So, line 104 should be fine.

However, I checked the hdf5 file for `g_max` and `g_min` and it is stored as `H5T_STD_I64LE` which would suggest that we need to change the type from `int` to `int64_t`.
Relevant code:
```
    std::vector<int> g_min, g_max;
    H5ReadDataset1D<int>(file, path + "scatter_data/g_min", g_min);
    H5ReadDataset1D<int>(file, path + "scatter_data/g_max", g_max);
```
And that's not my patch currently handles... And it is not in #90 either.
It seems like you have stumbled upon another problem with datatypes...

PR #787, opened 2025-10-02
Number of groups is stored as `size_t`

andrsd:
Refs #90


wdhawkins (approved):

modules/linear_boltzmann_solvers/lbs_problem/device/carrier/total_xs_carrier.h
```diff
@@ -26,7 +27,7 @@ class TotalXSCarrier : public Carrier
   double* GetXSGPUData(int block_id);
 
   /// Number of groups.
-  std::uint32_t num_groups = UINT32_MAX;
+  std::size_t num_groups = std::numeric_limits<size_t>::max();
```
quocdang1998, line 30: Can you replace `size_t` with `std::size_t`? Some compilers like GCC distinguish between C `size_t` and C++ `std::size_t` (one cannot cast a `size_t *` into a `std::size_t *` because the 2 types are not alias of each other).

andrsd: Yep, good catch!

Issue #70, opened 2024-01-09
Arbitrary boundary IDs


zhardy-lanl:
Currently, the code assumes only `[x|y|z][min|max]` boundaries. Under the hood, this really means that boundary IDs are restricted from 0-5. To assign a boundary to boundary ID 2, for example, you have to know that `ymin` corresponds to index 2. Instead of this, it would be beneficial to be able to tie boundary conditions to arbitrary boundary IDs. Ultimately, it would be good to default all boundary IDs to vacuum, and to pass non-vacuum boundaries as a map that looks something like
```
boundary_conditions = `{<n1>, <bc1>, <n2> = `<bc2>`, ....}
```
It would make the most sense to have a `BoundarySource` object that is passed in, and to specify reflective boundaries as the string `reflecting`/`reflective`, but the mechanics to get there I am not sure.

Refs #66

zhardy-lanl:
Updating this issue to help close #267

andrsd:
Update (29 OCT 2025): I found some code in OpenSn that is setting up boundaries from exodusII file. So, I built a test mesh with some side sets. I ran a simple problem (`transport_3d_1b_ortho.py`) and varied the BC by applying it on a different sides and observed what I think is correct behavior. The code is capable of using the side set number from the file, so all seems to check out. But, I think this should be guarded by a test, so I can make one for this.

andrsd:
This is done now. Any new work related to boundary IDs/names should open a new issue.

PR #809, opened 2025-10-24
`LBSProblem` does not set {x|y|z}{min|max} boundary IDs

andrsd:
- `LBSProblem` does not set any boundary IDs anymore
- {x|y|z}{min|max} boundary IDs are set only by the orthogonal mesh generator, because it makes sense there
- if orthogonal mesh is loaded from a file and user want {x|y|z}{min|max} boundary IDs, they must call `grid.SetOrthogonalBoundaries()`

refs #70


andrsd:
This is part 1 that works towards #70. Follow-up PRs will update mesh readers to set the boundary IDs.

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cc
```diff
@@ -1023,26 +1019,7 @@ LBSProblem::InitializeParrays()
...
-          boundary_id = YMIN;
-        else if (n.Dot(jhat) > 0.999)
-          boundary_id = YMAX;
-        else if (n.Dot(khat) < -0.999)
-          boundary_id = ZMIN;
-        else if (n.Dot(khat) > 0.999)
-          boundary_id = ZMAX;
-
-        if (boundary_id >= 0)
-          face.neighbor_id = boundary_id;
```
ragusa: Not needed anymore?

ragusa (commented):
What happens when a gmsh mesh is read and boundary markers are not present and that mesh is not ina box so that set orthogonal boundary doesn't work?

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cc
andrsd: While this PR removes this code, it does preserves this functionality. It is just now located in `MeshContinuum::SetOrthogonalBoundaries()` and must be explicitly asked for by a user.

There is a slight change thou. We are not setting all of `[x|y|z][min|max]`, we only set the ones that make sense for given spatial dimension.

PR #810, opened 2025-10-24
Add surface angular flux I/O support

rodriguez-jay:
- Added functionality to export surface angular fluxes at a specified boundary id or user provided surface id. 
- Added functionality for computing the response function along a surface provided forward and adjoint surfaces.

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
@@ -7,6 +7,9 @@
 #include "framework/runtime.h"
 #include "framework/mesh/mesh_continuum/mesh_continuum.h"
 #include "framework/utils/hdf_utils.h"
+#include <algorithm>
+#include <cmath>
+#include <unordered_set>
 
 namespace opensn
 {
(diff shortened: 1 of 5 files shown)


python/lib/solver.cc
```diff
@@ -403,6 +403,77 @@ WrapLBS(py::module& slv)
...
+  lbs_problem.def(
+    "WriteSurfaceAngularFluxes",
+    [](DiscreteOrdinatesProblem& self, 
+      const std::string& file_base, 
+      py::list bndry_names,
+      py::object surfaces)
+    {
+      // Map boundary names
+      std::map<std::string, uint64_t> allowed_bd_names = LBSProblem::supported_boundary_names;
+      std::map<std::uint64_t, std::string> allowed_bd_ids = LBSProblem::supported_boundary_ids;
```
quocdang1998: replace this with `const std::map<...>&`. You don't need to copy these maps to use them.

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -228,9 +228,18 @@ LBSSolverIO::ReadAngularFluxes(
     H5ReadDataset1D<double>(file_id, group_name + "/values", values);
     for (uint64_t c = 0; c < file_num_local_cells; ++c)
     {
+      // bool isBndry = false;
```
andrsd: remove if this is not needed

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -239,9 +248,331 @@ LBSSolverIO::ReadAngularFluxes(
...
+  std::optional<std::pair<std::string, double>> surfaces)
+{
+  // Open the HDF5 file
+  std::string file_name = file_base + std::to_string(opensn::mpi_comm.rank()) + ".h5";
+  hid_t file_id = H5Fcreate(file_name.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+  OpenSnLogicalErrorIf(file_id < 0, "WriteSurfaceAngularFluxes: Failed to open " + file_name + ".");
+
+  log.Log() << "Writing surface angular flux to " << file_base;
+
+  std::vector<std::vector<double>>& psi = do_problem.GetPsiNewLocal();
```
andrsd:
```
auto & psi = ...
```

andrsd (changes requested):

rodriguez-jay:
- Added surface angular-flux write/read support for boundary and internal surfaces.
- Exposed the new surface angular-flux functionality through the Python bindings.
- Cleaned up the implementation and resolved the clang-tidy.
- Added response_2d_surf.py as an adjoint regression test, including checks for:
  - Forward and adjoint QoIs
  - Forward surface angular flux
  - Adjoint surface importance

python/lib/solver.cc
```diff
@@ -403,6 +403,77 @@ WrapLBS(py::module& slv)
...
+      // LBSSolverIO::WriteSurfaceAngularFluxes(self, file_base, bndry_map);
+      LBSSolverIO::WriteSurfaceAngularFluxes(self, file_base, bndrys, opt_surfaces);
+    },
+    R"(
+    Write surface angular flux data to file.
+
+    Parameters
+    ----------
+    file_base: str
+        File basename.
```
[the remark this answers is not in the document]
rodriguez-jay: Addressed in the updated implementation:
- The API now accepts a single surfaces argument, documented in solver.cc.
- Boundary surfaces are specified as strings, while internal surfaces are specified as (name, {axis: value}) tuples.
- Internal surface tuples are validated to contain exactly two elements, and the accompanying dictionary must contain exactly one axis-value pair.

See python/lib/solver.cc, lines 464–514.

python/lib/solver.cc
```diff
@@ -403,6 +403,77 @@ WrapLBS(py::module& slv)
...
+      const std::string& file_base, 
+      py::list bndry_names,
+      py::object surfaces)
+    {
+      // Map boundary names
+      std::map<std::string, uint64_t> allowed_bd_names = LBSProblem::supported_boundary_names;
+      std::map<std::uint64_t, std::string> allowed_bd_ids = LBSProblem::supported_boundary_ids;
+      std::vector<std::string> bndrys;
+      for (py::handle name : bndry_names)
+        bndrys.push_back(name.cast<std::string>());
```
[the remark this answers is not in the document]
rodriguez-jay: Boundary names are now explicitly validated against the boundary-name map and the boundaries present on the grid. See angular_flux_io.cc, lines 288–297.

python/lib/solver.cc
rodriguez-jay: Updated, see in angular_flux_io.cc, line 277.

python/lib/solver.cc
```diff
@@ -403,6 +403,77 @@ WrapLBS(py::module& slv)
...
+    {
+      std::map<std::string, std::uint64_t> supported_bd_names = LBSProblem::supported_boundary_names;
+      std::map<std::uint64_t, std::string> supported_bd_ids = LBSProblem::supported_boundary_ids;
+
+      // std::map<std::string, std::uint64_t> bndry_map;
+      std::vector<std::string> bndrys;
+      for (py::handle name : bndry_names)
+      {
+        std::string bndry = name.cast<std::string>();
+        // bndry_map[bndry] = supported_bd_names.at(bndry);
```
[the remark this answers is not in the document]
rodriguez-jay: When writing, the implementation accepts both boundary and internal surface definitions. Boundary names are explicitly validated against the boundary-name map and the boundaries present on the grid (angular_flux_io.cc, lines 288–297). When reading, each requested surface name is validated through checked HDF5 reads; a missing or misspelled name raises an exception (angular_flux_io.cc, lines 547–560).

python/lib/solver.cc
```diff
@@ -403,6 +403,77 @@ WrapLBS(py::module& slv)
...
+    py::arg("file_base"),
+    py::arg("bndry_names"),
+    py::arg("surfaces") = py::none()
+  );
+  lbs_problem.def(
+    "ReadSurfaceAngularFluxes",
+    [](DiscreteOrdinatesProblem& self, const std::string& file_base, py::list bndry_names)
+    {
+      std::map<std::string, std::uint64_t> supported_bd_names = LBSProblem::supported_boundary_names;
+      std::map<std::uint64_t, std::string> supported_bd_ids = LBSProblem::supported_boundary_ids;
```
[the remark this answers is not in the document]
rodriguez-jay: Done.

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
rodriguez-jay: Addressed, it is gone.

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
rodriguez-jay: Updated to const auto& psi because the writer only reads the angular-flux data. This avoids copying while preserving const correctness. See angular_flux_io.cc, line 27

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -239,9 +248,331 @@ LBSSolverIO::ReadAngularFluxes(
...
+    bndry_ids.push_back(bndry_id);
+    const auto id = std::find(unique_bids.begin(), unique_bids.end(), bndry_id);
+    OpenSnInvalidArgumentIf(id == unique_bids.end(),
+                            "Boundary " + bndry + "not found on grid.");
+  }
+
+  std::vector<std::string> bndry_tags;
+
+  // ===========================================================
+  // The goal is limit the number of itrations over all cells
```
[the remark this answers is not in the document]
rodriguez-jay: Addressed, the old comment containing the typo was removed during cleanup.

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -239,9 +248,331 @@ LBSSolverIO::ReadAngularFluxes(
...
+    OpenSnInvalidArgumentIf(id == unique_bids.end(),
+                            "Boundary " + bndry + "not found on grid.");
+  }
+
+  std::vector<std::string> bndry_tags;
+
+  // ===========================================================
+  // The goal is limit the number of itrations over all cells 
+  // Sweep through once and map cell_ids to boundaries then
+  // upack them into 1D vectors for exporting to H5
```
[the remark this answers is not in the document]
rodriguez-jay: Updated the terminology from H5 to HDF5 throughout the implementation. See angular_flux_io.cc, line 440.

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -239,9 +248,331 @@ LBSSolverIO::ReadAngularFluxes(
...
+      uint64_t num_cell_nodes = 0;
+
+      const auto& unit_cell_matrices = do_problem.GetUnitCellMatrices();
+	    const auto& fe_values = unit_cell_matrices.at(cell.local_id);
+
+      // Loop over each face of the cell
+      unsigned int f = 0;
+      for (const auto& face : cell.faces)
+      {
+        bool isSurf = false;
```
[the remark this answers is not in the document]
rodriguez-jay: Renamed isSurf to is_surface to follow the project’s naming convention. See angular_flux_io.cc, line 337.

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -239,9 +248,331 @@ LBSSolverIO::ReadAngularFluxes(
...
+
+  ssize_t num_open_objs = H5Fget_obj_count(file_id, H5F_OBJ_ALL);
   H5Fclose(file_id);
 }
 
+std::vector<LBSSolverIO::SurfaceAngularFlux>
+LBSSolverIO::ReadSurfaceAngularFluxes(
+  DiscreteOrdinatesProblem& do_problem,
+  const std::string& file_base,
+  std::vector<std::string>& bndrys)
```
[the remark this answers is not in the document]
rodriguez-jay: Updated the writer and reader surface parameters to use const&, see discrete_ordinates_problem_io.h, lines 115–131.

modules/linear_boltzmann_solvers/lbs_problem/io/angular_flux_io.cc
```diff
@@ -239,9 +248,331 @@ LBSSolverIO::ReadAngularFluxes(
...
+    }
+  }
+  H5Fclose(file_id);
+}
+
+void
+LBSSolverIO::WriteSurfaceAngularFluxes(
+  DiscreteOrdinatesProblem& do_problem,
+  const std::string& file_base,
+  std::vector<std::string>& bndrys,
```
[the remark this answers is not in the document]
rodriguez-jay: Done.

modules/linear_boltzmann_solvers/response_evaluator/response_evaluator.cc
```diff
@@ -315,8 +347,11 @@ double
 ResponseEvaluator::EvaluateResponse(const std::string& buffer) const
 {
   const auto& buffer_data = adjoint_buffers_.at(buffer);
-  const auto& phi_dagger = buffer_data.first;
-  const auto& psi_dagger = buffer_data.second;
+  // const auto& phi_dagger = buffer_data.first;
+  // const auto& psi_dagger = buffer_data.second;
```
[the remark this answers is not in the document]
rodriguez-jay: The earlier surface-integration changes to ResponseEvaluator were removed, so this comment no longer applies.

andrsd (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,427 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+    const std::string group_name = "groupset_" + std::to_string(groupset_id);
+    OpenSnLogicalErrorIf(not H5CreateGroup(file.Id(), group_name), CreateGroupError(group_name));
+    OpenSnLogicalErrorIf(
+      not H5CreateAttribute(file.Id(), group_name + "/num_directions", num_gs_dirs),
+      "Failed to write the direction count for " + group_name + ".");
+    OpenSnLogicalErrorIf(
+      not H5CreateAttribute(file.Id(), group_name + "/num_groups", num_gs_groups),
+      "Failed to write the group count for " + group_name + ".");
+
+    for (const auto& surf_id : surface_tags)
```
wdhawkins, line 457: This loop only writes h5 groups for surface tags that exist on the current rank. However, `ReadSurfaceAngularFluxes` tries to read every requested surface tag from every h5 file. If a surface exists only on some ranks, ranks without local faces for that surface will not have the corresponding data in their h5 files, causing the read to fail. We should either write empty datasets for all surfaces on every rank, or the reading routine shouldn't fail on missing surface data.

It also looks like we only validate groupset, group, and direction counts. We should also compare cell ids, face node counts, etc. I believe `ReadAngularFluxes` does pretty extensive checks. You might look at it to see what it does.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,427 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+                ++nodes_on_face;
+            }
+
+            if (nodes_on_face == num_face_nodes)
+            {
+              const Vector3 global_norm = axis == "x"   ? Vector3{1.0, 0.0, 0.0}
+                                          : axis == "y" ? Vector3{0.0, 1.0, 0.0}
+                                                        : Vector3{0.0, 0.0, 1.0};
+              const auto alignment = face.normal.Dot(global_norm);
+              surf_name = surf_id + (alignment > 0 ? "_u" : "_d");
```
wdhawkins, line 377: What happens if a face matches multiple definitions? Here it looks like we would just overwrite each match with the next.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,427 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+          }
+        }
+
+        // Boundary surface mapping
+        if (not bndry_surfs.empty() and not face.has_neighbor)
+        {
+          const auto id = std::find(bndry_ids.begin(), bndry_ids.end(), face.neighbor_id);
+          if (id != bndry_ids.end())
+          {
+            surf_name = allowed_bd_ids.at(*id);
```
wdhawkins, line 390: See comment at line 370.

wdhawkins (changes requested):

python/lib/solver.cc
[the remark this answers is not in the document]
quocdang1998: If I understand correctly, you expect `surfaces` to be either a `List[str]` (`boundary_surfaces`) or a `List[Tuple[str,Dict]]` (`internal_surfaces`).

However, from the C++ API, `WriteSurfaceAngularFluxes` can accept both form, which mean that on the C++ API, user is allowed to provide both a list of surface names and a list of `(name, {axis: value})`. This feature is not available on the Python interface. Is this intentional or just a bug?

Regarding the second case (`internal_surfaces`), change the type of this argument in C++ API to `std::map<std::string, std::pair<std::string, double>>`, which is much more readable than `std::vector<std::pair<...>>`. In the Python interface, allow this argument to accept either a `List[str]` or a `Dict[str,Tuple[str,float]]`. At the end, a vector of pairs is simply a map.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: Addressed in two parts:
1. Every rank now writes HDF5 groups for every requested surface. Ranks without local faces write empty datasets, so the reader can safely access all requested surfaces.
2. The reader now validates the stored surface mesh against the current mesh, including cell IDs, face-node counts, node coordinates, and angular-flux/finite-element dataset sizes.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: Addressed by stopping after the first matching surface definition, preventing subsequent definitions from overwriting the surface tag. Since the current implementation supports axis-aligned planar slices, a non-degenerate face cannot match planes on different axes. For duplicate/coincident definitions, the first supplied definition is used.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: Addressed with the fix discussed at line 370: the loop now stops after the first matching surface definition, so later definitions cannot overwrite the selected surface tag.

test/python/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_surf.py
```diff
@@ -0,0 +1,186 @@
...
+    from pyopensn.fieldfunc import FieldFunctionInterpolationVolume
+    from pyopensn.logvol import RPPLogicalVolume
+    from pyopensn.mesh import OrthogonalMeshGenerator
+    from pyopensn.response import ResponseEvaluator
+    from pyopensn.solver import DiscreteOrdinatesProblem, SteadyStateSourceSolver
+    from pyopensn.source import VolumetricSource
+    from pyopensn.xs import MultiGroupXS
+else:
+    barrier = MPIBarrier
+    global_sum = MPIAllReduce
```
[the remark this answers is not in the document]
rodriguez-jay: No change needed. The OpenSn console initializes rank and size in the main Python namespace before executing the input script. The console branch only aliases MPIBarrier and MPIAllReduce, so the later size and rank references are already defined.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,481 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+              const auto alignment = face.normal.Dot(global_norm);
+              surf_name = surf_id + (alignment > 0 ? "_u" : "_d");
+              is_surface = true;
+              break;
+            }
+          }
+        }
+
+        // Boundary surface mapping
+        if (not bndry_surfs.empty() and not face.has_neighbor)
```
wdhawkins, line 385: What happens if a face matches both an internal surface and a requested boundary? Shouldn't this be something like: `if (not is_surface and not bndry_surfs.empty() and not face.has_neighbor)`?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,481 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+DiscreteOrdinatesProblemIO::WriteSurfaceAngularFluxes(
+  DiscreteOrdinatesProblem& do_problem,
+  const std::string& file_base,
+  const std::vector<std::string>& bndry_surfs,
+  const std::vector<std::pair<std::string, std::pair<std::string, double>>>& int_surfs)
+{
+  OpenSnLogicalErrorIf(bndry_surfs.empty() and int_surfs.empty(),
+                       "No surface provided. Provide either boundary names or internal surface "
+                       "definitions.");
+
```
wdhawkins, line 258: There should probably be a check here that `save_angular_flux` is set: 
`OpenSnInvalidArgumentIf(not do_problem.GetOptions().save_angular_flux,
 "WriteSurfaceAngularFluxes requires `options.save_angular_flux=true`.");`

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,481 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+  const auto& discretization = do_problem.GetSpatialDiscretization();
+  const auto& groupsets = do_problem.GetGroupsets();
+  const auto num_groupsets = static_cast<uint64_t>(groupsets.size());
+
+  // Check the boundary IDs
+  std::vector<uint64_t> bndry_ids;
+  std::set<std::string> surface_tags;
+  if (not bndry_surfs.empty())
+  {
+    const auto unique_bids = grid->GetUniqueBoundaryIDs();
```
wdhawkins, line 288: `GetUniqueBoundaryIDs` is a collective, but `bndry_surfs.empty()` is rank-dependent. Some ranks may not make it inside the `if`, and the collective will fail. This should probably be something like:
```
const auto unique_bids = grid->GetUniqueBoundaryIDs();  // always call — this is collective
std::vector<uint64_t> bndry_ids;
std::set<std::string> surface_tags;
if (not bndry_surfs.empty())
{
  for (const auto& bndry : bndry_surfs)
  {
    ...
    const auto id = std::find(unique_bids.begin(), unique_bids.end(), bndry_id);
    ...
  }
}
```

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,481 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+  OpenSnLogicalErrorIf(not H5CreateGroup(file.Id(), "mesh"),
+                       "Failed to create the mesh group in " + file_name + ".");
+
+  // Mesh information is shared by all groupsets.
+  std::map<std::string, std::vector<uint64_t>> cell_map, node_map;
+  std::map<std::string, std::vector<double>> x_map, y_map, z_map;
+
+  constexpr double surface_tolerance = 1.0e-10;
+
+  for (size_t groupset_index = 0; groupset_index < groupsets.size(); ++groupset_index)
```
wdhawkins, line 325: It looks like there are a number of computations inside the loop that aren't groupset dependent. They can be hoisted outside of the loop. Not a blocker for merging, but could be a nice performance win.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,481 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+
+          const Vector3 file_node{surf_map.nodes_x[node_stride],
+                                  surf_map.nodes_y[node_stride],
+                                  surf_map.nodes_z[node_stride]};
+          centroid += file_node;
+          file_nodes.push_back(file_node);
+          const QuantizedCoordinate vertex = {Quantize(surf_map.nodes_x[node_stride]),
+                                              Quantize(surf_map.nodes_y[node_stride]),
+                                              Quantize(surf_map.nodes_z[node_stride])};
+          face_map[vertex] = stride;
```
wdhawkins, line 637: Is `face_map` actually used? I see where it gets pushed back into `faces`, but it doesn't seem to be used after that.

wdhawkins (changes requested):

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: Good point. I added validation requiring tuple-defined internal surfaces to be geometrically internal. A plane coincident with an exterior boundary is now rejected with an instruction to request that boundary by name instead. The validation is performed collectively before creating the output files, and the overlap case is covered by the regression test

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: Agreed. Surface export reads the stored angular-flux vector, so I added an early check using SaveAngularFluxEnabled(). It raises a clear error before creating the HDF5 file when options.save_angular_flux is false. The regression test now covers this rejection path and verifies that no output file is produced.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: Agreed. GetUniqueBoundaryIDs() is now called unconditionally on all ranks.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: I hoisted the surface selection, mesh geometry, node mapping, finite-element shape data, mass matrices, and mesh output outside the groupset loop. The loop now performs only groupset-dependent angular-flux work.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
rodriguez-jay: `face_map` was a remnant of earlier surface-response implementation, it is not used. I removed it and the associated SurfaceMap::faces field.

python/lib/solver.cc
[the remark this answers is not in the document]
rodriguez-jay: Mixed boundary and internal surfaces were already supported because each list element is classified independently. I clarified that behavior in the documentation and retained it. I also changed the C++ internal-surface container to a map and added support for the proposed Python dictionary form, with regression coverage.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,544 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+    return message;
+  };
+
+  // Get angular fluxes
+  const auto& psi = do_problem.GetPsiNewLocal();
+
+  // Check the boundary IDs
+  std::vector<uint64_t> bndry_ids;
+  std::set<std::string> surface_tags;
+  if (not bndry_surfs.empty())
```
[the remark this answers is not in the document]
rodriguez-jay: Changes:
- Moved boundary validation before file creation in [angular_flux_io.cc (line 337)] while preserving the collective boundary-ID lookup.
- Added a regression check in [response_2d_surf.py (line 127)] proving invalid input does not modify an existing output file.

python/lib/solver.cc
```diff
@@ -486,6 +486,146 @@ WrapLBS(py::module& slv)
...
+          {
+            const auto surface_tuple = surface.cast<py::tuple>();
+            if (surface_tuple.size() != 2)
+              throw std::invalid_argument(
+                "Surface tuple must have 2 elements (name, {axis: value})");
+
+            const auto surface_name = surface_tuple[0].cast<std::string>();
+            const auto surface_dict = surface_tuple[1].cast<py::dict>();
+            if (surface_dict.size() != 1)
+              throw std::invalid_argument("Surface dict must have exactly one key-value pair");
```
quocdang1998, line 521: In this case, I'd force user to input `Tuple[str,str,float]` directly and remove the possibility to input `Dict[str,...]`.

The more input formats you support, the more complex and less maintainable your code becomes.

I would suggest the wrapper to be simply:
```
      [](DiscreteOrdinatesProblem& self, const std::string& file_base, const std::vector<std::string>& boundary_surfaces, const std::map<std::string, std::pair<std::string, double>>& internal_surfaces)
      {
        DiscreteOrdinatesProblemIO::WriteSurfaceAngularFluxes(
          self, file_base, boundary_surfaces, internal_surfaces);
      },
      ...,
      py::arg("self"),
      py::arg("file_base"),
      py::arg("boundary_surfaces") = std::vector<std::string>{},
      py::arg("internal_surfaces") =
        std::map<std::string, std::pair<std::string, double>>{}
```

quocdang1998 (changes requested):

python/lib/solver.cc
rodriguez-jay: Good point. WriteSurfaceAngularFluxes now uses separate boundary_surfaces and interior_surfaces arguments, removing the heterogeneous-list parser. Interior surfaces use the form {"name": ("axis", value)}. The regression tests and documentation have been updated accordingly.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,544 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+  {
+    for (const auto& bndry : boundary_surfs)
+    {
+      const auto bndry_name_it = allowed_bd_names.find(bndry);
+      OpenSnInvalidArgumentIf(bndry_name_it == allowed_bd_names.end(),
+                              "Boundary " + bndry + " not found in the boundary-name map.");
+
+      const auto bndry_id = bndry_name_it->second;
+      const auto id = std::find(unique_bids.begin(), unique_bids.end(), bndry_id);
+      OpenSnInvalidArgumentIf(id == unique_bids.end(), "Boundary " + bndry + " not found on grid.");
```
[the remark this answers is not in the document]
rodriguez-jay: Proposed fix implemented.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/angular_flux_io.cc
```diff
@@ -242,5 +245,544 @@ DiscreteOrdinatesProblemIO::ReadAngularFluxes(
...
+
+      OpenSnLogicalErrorIf(surf_map.cell_ids.size() != surf_map.num_face_nodes.size(),
+                           "Incompatible surface mesh metadata for " + surface + ".");
+      OpenSnLogicalErrorIf(surf_map.nodes_x.size() != surf_map.nodes_y.size() or
+                             surf_map.nodes_x.size() != surf_map.nodes_z.size(),
+                           "Incompatible surface node coordinates for " + surface + ".");
+
+      std::vector<uint64_t> cell_stride;
+      std::vector<uint64_t> node_index = {0};
+      std::vector<uint64_t> dir_index = {0};
```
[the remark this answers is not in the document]
rodriguez-jay: Done.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/io/discrete_ordinates_problem_io.h
```diff
@@ -62,6 +65,72 @@ class DiscreteOrdinatesProblemIO
...
+  {
+    std::vector<double> omega;
+    std::vector<double> mu;
+    std::vector<double> wt_d;
+    std::vector<double> mass_matrix;
+    std::vector<double> fe_shape;
+    std::vector<double> psi;
+    std::vector<uint64_t> node_index;
+    std::vector<uint64_t> dir_index;
+  };
```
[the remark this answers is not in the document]
rodriguez-jay: Documentation added.

quocdang1998 (approved):

PR #812, opened 2025-10-28
Added fix when user requests for power scaling, the scalar flux should also be scaled accordingly

kitamuraa24:
Added extra routine when the ' power_field_function_on ' is toggled on in the power_generation subroutine. The scalar flux will now be scaled accordingly if the requested power is not that of the one calculated in the code.


kitamuraa24:
is the clang-format standard now 19 by my understanding?

wdhawkins (approved):

PR #813, opened 2025-10-29
Adding subtype capabilities to foam_io.cc

kitamuraa24:
' foam_io.cc ' now selects subtypes for cells instead of defaulting to polyhedrons; this should now allow OpenSn to take advantage of uniform cell type meshes and select a preferred sweep method.


andrsd:
I marked this as a draft. When you have updates we can convert it into a regular PR again...

PR #815, opened 2025-10-30
GPU implementation for AAH sweep (version 2.0)

quocdang1998:
In this PR, a second version of the GPU sweep kernel using the AAH sweep has been implemented.

Major changes:

- `AAHD_FLUDS` has been upgraded to eliminate the need for frequent host-device transfers between sweep levels. During the sweep, the kernel can fetch angular flux data directly from device memory without returning to the host. Host-device communication now only occurs at the beginning and end of the sweep, rather than between levels.
- Kernels are templated based on the number of cell DOFs, enabling compile-time loop unrolling.
- `AAHAsyncCommunicator` has been refactored to support both host-device transfers and preloc/delayed preloc/deploc operations.
- A new regression test for the GPU (`transport_3d_4_cycles_1_gpu.py`) has been added. This test checks the functionality of all banks (local, delayed, non-local, and boundary) in the AAH FLUDS.

Features to be added:

- Transfer portions of cell data to shared memory, particularly face data and node indices.
- Separate kernels for cells with small, medium, and large numbers of DOFs.
- Automatic GPU selection based on MPI rank when multiple GPUs are available to multiple MPI ranks.


modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/communicators/aah_async_comm.cc
```diff
@@ -214,18 +145,19 @@ AAH_ASynchronousCommunicator::ReceiveDelayedData(int angle_set_num)
...
   {
     auto& upstream_psi = fluds_.DelayedPrelocIOutgoingPsi()[i];
 
-    assert(delayed_preloc_msg_data_[i].size() <= std::numeric_limits<int>::max());
-    for (int m = 0; m < delayed_preloc_msg_data_[i].size(); ++m)
+    for (std::size_t m = 0; m < delayed_preloc_msg_data_[i].size(); ++m)
     {
       const auto& [source, size, block_pos] = delayed_preloc_msg_data_[i][m];
-      const int tag = max_num_messages_ * angle_set_num + m;
+      const std::size_t utag = max_num_messages_ * static_cast<std::size_t>(angle_set_num) + m;
```
andrsd: `tag` was changed to `int`, because MPI uses tags as `int`, so probably should stay as `int`, unless you prefer those `static_cast`s...

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
```diff
@@ -23,508 +16,163 @@ namespace crb = caribou;
...
-  std::uint32_t level_size;
-  // batch
-  std::uint32_t batch_size;
-};
 
-/// @brief Index for each thread.
-struct Index
+/// Kernel performing the sweep
+__global__ void
+UnsarturatedKernel(Arguments args, const std::uint32_t* level, double* saved_psi)
```
andrsd: "UnsarturatedKernel". Should this be "UnsaturatedKernel"?

modules/linear_boltzmann_solvers/lbs_problem/device/carrier/mesh_carrier.cu
```diff
@@ -38,6 +40,8 @@ MeshCarrier::ComputeSize(LBSProblem& lbs_problem)
     alloc_size += sizeof(std::uintptr_t);
     // phi address
     alloc_size += sizeof(std::uint64_t);
+    // offset for save angular flux
```
andrsd: "save**d** angular flux"?

andrsd:
As a general comment, there is a significant use of `@param`, etc for doxygen. Majority of the code uses `\param` (there was some recent creep-in in the `@` syntax). This is not codified in the coding standard, so should it be? @ragusa, @wdhawkins

andrsd:
Another thing in this PR. As a general rule "Comments should be added to explain why the code is doing what it is doing", so things like this are not very useful:
```
// get mesh reference
Grid & grid = var.GetGrid();
```
```
// Destructor
~MyClass();
```

I would highly advice against doing this. In the second case, it can be argued that it is for doxygen, but the question is "does that really help with anything but 100% coverage on API?"

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
quocdang1998: Fixed. Thanks a lot.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
```diff
@@ -23,508 +16,163 @@ namespace crb = caribou;
...
-  args.flud_data = flatten_fluds.GetDevicePtr();
-  args.flud_stride = flatten_fluds.GetStride();
-  // prepare memory for level
-  Storage<std::uint32_t>* level_storage = reinterpret_cast<Storage<std::uint32_t>*>(level_vector_);
-  args.level = level_storage->GetDevicePtr();
-  // allocate data for angular if required
-  Storage<double> psi_storage;
-  double* psi_device = nullptr;
+  phi->CopyFromDevice();
+  OutflowCarrier* outflow = reinterpret_cast<OutflowCarrier*>(problem_.GetCarrier(1));
```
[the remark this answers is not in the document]
quocdang1998: Fixed. I changed the type declaration to `auto*`. `clang-tidy` does not report this.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cu
```diff
@@ -23,508 +16,163 @@ namespace crb = caribou;
...
-           DirectionView& direction,
-           CellView& cell,
-           const double* src_moment,
-           const std::uint32_t& groupset_start,
-           const std::uint32_t& group_idx,
-           const std::uint32_t& num_groups,
-           const std::uint32_t& num_moments)
+/// Kernel performing the sweep
+__global__ void
+SarturatedKernel(Arguments args, const std::uint32_t* level, double* saved_psi)
```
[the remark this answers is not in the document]
quocdang1998: I fixed this misspelling in the latest version.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -2,353 +2,372 @@
...
+#include "framework/logging/log.h"
 
 namespace opensn
 {
 
-template <typename T>
-inline static char*
-write_to(char* buffer, const T& src)
+static inline bool
+IsTotalSizeNotEmpty(const std::vector<std::size_t>& sizes)
```
andrsd: 1. `IsTotalSizeNonZero` might be a better name. Size does not have an empty property, container does.

2. Also, consider changing this into `TotalSizeIsZero` (i.e. inverting the logic) and then user code doing:
```
if (not TotalSizeIsZero(...))
```
which usually reads easier than conditions in negation...

Either solution is fine, your call...

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/communicators/aah_async_comm.h
```diff
@@ -5,38 +5,121 @@
...
 
 namespace mpi = mpicpp_lite;
 
 namespace opensn
 {
 
-class MPICommunicatorSet;
-class FLUDS;
+/// \brief Message details for each message passing.
+using AAHMessageDetails = std::tuple<int, int, int>;
```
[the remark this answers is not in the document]
quocdang1998: Thanks for the comment. I changed this to a `struct` in the latest version.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/communicators/aah_async_comm.cc
```diff
@@ -31,7 +32,15 @@ AAH_ASynchronousCommunicator::AAH_ASynchronousCommunicator(FLUDS& fluds,
     data_initialized_(false),
     upstream_data_initialized_(false)
 {
-  this->BuildMessageStructure();
+  bool v1_success = this->BuildMessageStructureV1();
```
wdhawkins: Instead of V1 and V2, can we come up with some more descriptive names?

quocdang1998: I've changed the name to `BuildMessageStructureForCPUSweep` and `BuildMessageStructureForGPUSweep` in the latest version.

PR #819, opened 2025-11-05
Updating CMake dependencies script for better handling of HDF5 and improved output/logging

wdhawkins:
This PR fixes the following issues with the CMake dependencies script:

1. A system-installed `HDF5` could cause `VTK` installation to fail.
2. Final message about sourcing the `set_opensn_env.sh` script was not output at the end of the install process.
3. Script output was cluttered and not logged.

### DIFF
--- a/doc/source/install/install.rst
+++ b/doc/source/install/install.rst
@@ -56,17 +56,41 @@ To clone your fork of **OpenSn**:
 Install dependencies
 --------------------
 
-Dependencies can be installed to ``/path/to/dependencies/directory`` with CMake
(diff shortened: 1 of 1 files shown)


andrsd (approved):

wdhawkins:
> repeat of comment left: should we echo this in the online documentation (install instructions) as well? maybe as a soft reminder that the install script will likle ypoint out to an evn script to set was vars and that it'll be good to add that to the .bashrc ....

I've updated the install instructions with information about the dependencies script.

PR #823, opened 2025-11-06
Add multiple GPUs support for the Python interface

quocdang1998:
In this PR, the GPU selection feature is added to the Python interface, allowing users to launch tests on multiple GPUs when more than one GPU is accessible to an MPI rank.


wdhawkins:
A quick comment. This functionality could be useful for testing and/or running on single-user workstations, but really shouldn't be used for production runs on large machines. When there are multiple GPUs on a node, we should leave it up to the job scheduler to associate ranks with GPUs (with the hope that the scheduler does this intelligently and takes CPU/GPU affinity into account). We should probably add a note to that effect in the documentation for these methods.

quocdang1998:
Close due to incompatibility with state of the code.

PR #827, opened 2025-11-08
CMake changes to support proper OpenSn installs and `find_package(OpenSn)` for applications

wdhawkins:
This PR adds support for proper OpenSn installs to `CMAKE_INSTALL_PREFIX` and use of `find_package(OpenSn)` for applications.  For example, the following would work for an application:
```
find_package(OpenSn REQUIRED)
message(STATUS "Found OpenSn ${OpenSn_VERSION} (from: ${OpenSn_DIR})")
.
.
.
target_link_libraries(myapp
    PRIVATE
    opensn::opensn
)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,18 +229,19 @@ endif()
 
 target_include_directories(libopensn
     PRIVATE
-    $<INSTALL_INTERFACE:include/opensn>
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -378,19 +384,23 @@ install(
 install(
     FILES
         "${CMAKE_CURRENT_BINARY_DIR}/OpenSnConfig.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/opensnConfig.cmake"
```
andrsd: You may want to remove this line as well...

wdhawkins: Yep, missed that. Fixed.

PR #829, opened 2025-11-11
clang-tidy fixes

andrsd:


modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/communicators/aah_async_comm.h
```diff
@@ -61,19 +61,18 @@ SetupMessageData(const std::vector<int>& locations,
   for (std::size_t i = 0; i < num_locations; ++i)
   {
     // compute message count and size
-    std::size_t num_unknowns = get_unknown_count(i);
+    int num_unknowns = get_unknown_count(i);
```
andrsd: I converted this into `int` because that's what gets passed into MPI in the end. If the number of messages would be larger than `INT_MAX`, then this code must be significantly changed anyway. That's my reasoning behind this change.

wdhawkins (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/communicators/aah_async_comm.h
quocdang1998: We may perform a check as follows:

```
// compute message count and size
int num_unknowns = -1;
{  // open a scope here so variables in this scope won't be stored in the stack
    std::size_t num_unknowns_64 = get_unknown_count(i);
    if (num_unknowns_64 > static_cast<std::size_t>(std::numeric_limits<int>::max()))
    {
        throw std::overflow_error("Number of unknowns is too large for int");
    }
    num_unknowns = static_cast<int>(num_unknowns_64 );
}
```

quocdang1998 (changes requested):

quocdang1998 (approved):

PR #830, opened 2025-11-11
Local/Global ID in `ParallelSTLVector` is `uint64_t`

andrsd:
Refs #90

### DIFF
--- a/framework/field_functions/interpolation/ffinter_line.h
+++ b/framework/field_functions/interpolation/ffinter_line.h
@@ -51,7 +51,7 @@ private:
   int number_of_points_;
   double op_value_;
   Vector3 pi_, pf_;
-  std::vector<uint64_t> local_cells_;
+  std::vector<std::uint32_t> local_cells_;
   std::vector<Vector3> local_interpolation_points_;
   std::vector<double> local_interpolation_values_;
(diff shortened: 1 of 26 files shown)


wdhawkins (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/communicators/aah_async_comm.h
```diff
@@ -61,19 +61,18 @@ SetupMessageData(const std::vector<int>& locations,
   for (std::size_t i = 0; i < num_locations; ++i)
   {
     // compute message count and size
-    std::size_t num_unknowns = get_unknown_count(i);
+    int num_unknowns = get_unknown_count(i);
```
quocdang1998: @wdhawkins @andrsd Does the number of unknowns may exceed the max limit of int32? In the current version, we subtract `max_mpi_message_size` from `num_unknowns` until `num_unknowns` is less than `max_mpi_message_size` so that we can safely cast `num_unknowns` back to `int32` for the last message.

@andrsd Function `get_unknown_count` returns an `std::size_t`. I think we should perform a `static_cast<int>` here.

PR #833, opened 2025-11-14
Honor adjoint parameter set in `options`

andrsd:
When users set `adjoint` parameter in `options`, we save that state and use it during `LBSProblem::Initialize`.  This way users don't need to call `SetAdjoint` and `SetVolumetricSources` explicitly. It is done automatically during initialization.

### DIFF
--- a/test/python/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_4.py
+++ b/test/python/modules/linear_boltzmann_solvers/transport_adjoint/response_2d_4.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+2D PWLD adjoint transport test with localized material source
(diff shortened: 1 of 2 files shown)


wdhawkins (approved):

andrsd:
> Can we also add the adjoint input that Jason provided to the regression tests with this PR?

So do you want to guard that tutorials produce the same answer? If so, than that would be the test for this functionality. If I understood correctly, he provided the python version of the jupyter notebook.  But, I can make that into a regression test if you want me to...

ragusa:
there is a command to turn automatically a notebook to a python file (nbconvert I think). Should we just do this: convert all notebooks to python files (on the fly, while running the regression suite) and run them as additional tests? I am saying this because we will have to update both the notebook and the regression test file if we make a copy.

ragusa (approved):

PR #834, opened 2025-11-14
Remove scattering order parameter from DiscreteOrdinatesProblem

wdhawkins:
This PR removes the `scattering_order` parameter from `DiscreteOrdinatesProblem`. The solver scattering order and number of moments are now set to the values provided by the quadrature set(s). Note that all quadrature sets in the problem need to use the same scattering order. If the scattering order provided by the quadrature set is not equal to that provided by the cross-section data, the code will issue a warning.


modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
```diff
@@ -80,31 +77,58 @@ DiscreteOrdinatesProblem::DiscreteOrdinatesProblem(const InputParameters& params
...
-    // Build groupset angular flux unknown manager
     if (not groupset.quadrature)
     {
       std::stringstream oss;
       oss << GetName() << ":\nGroupset " << groupset.id
           << " does not have an associated quadrature set";
       throw std::runtime_error(oss.str());
     }
+
+    if (groupset.quadrature->GetScatteringOrder() != groupset.quadrature->GetScatteringOrder())
```
ragusa: I do not get this. Are you querying the same thing twice? It will never be different.

wdhawkins:
> This is all good but the documentation here needs to be changed as well. python/lib/solver.cc

Did I miss something? I removed all of the references to a `scattering_order` parameter for the solver.

PR #835, opened 2025-11-16
Refactoring SLDFESQ quadrature

wdhawkins:
This PR is a major refactor of the SLDFESQ angular quadrature. The overall class design has been simplified and refactored for readability and maintainability. Support for 2D quadrature sets has been added. The name of the class has been changed to SLDFEsqQuadrature with two derived classes: SLDFEsqQuadrature3DXYZ and SLDFEsqQuadratue2DXY. Python bindings have been added for these two methods. LocallyRefine is the only other user-facing method.

### DIFF
--- a/doc/source/pyapi/index.rst
+++ b/doc/source/pyapi/index.rst
@@ -88,7 +88,8 @@ Simplified LDFES quadrature
    :nosignatures:
    :template: python.rst
 
-   aquad.SLDFESQuadrature
+   aquad.SLDFEsqQuadrature3DXYZ
+   aquad.SLDFEsqQuadrature2DXY
 
 
 Field functions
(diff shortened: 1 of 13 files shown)


ragusa (commented):
long time ago, we had a name, `optimizeforpolarsymmetry`, and it was not great and, moreover, was exposed to the user. Just now, in sldfesq, we have `FilterQuadraturePoint`, and it is not exposed to the user (which is good). My only question: is it descriptive enough? I could imagine the following alternatives: `HasPositivePolarCosine`  or `FilterByPolarCosine`. I prefer the former because your routine returns true/false, and that goes well witht he question has ....

andrsd (changes requested):

PR #849, opened 2025-12-01
Adding extra checks for tutorials

andrsd:


doc/source/tutorials/lbs/src_driven/mg_src_driven/hdpe_example.ipynb
```diff
@@ -481,6 +481,17 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "avg_flux = np.average(flux)\n",
+    "if rank == 0:\n",
+    "    print(f\"Average flux = {avg_flux}\")"
+   ]
+  },
```
ragusa: i'd say we should add a little blurb stating the average is computed and printed out for integration in our testing suite but this value is not necessarily physical relevant here.

ragusa: so you kept average and did not change to max

andrsd: There is something weird going on. On my machine I see:
```
922e0a4a4 (HEAD -> issue/848, andrsd/issue/848) Adding extra checks for the HDPE tutorial
0b30f43fe Adding extra checks for logvol tutorial
cb8aa8919 Mesh tutorial print number of global cells and we check those in tests
```
and github still shows the old version:
```
afb01319a9ed8d1ebd50d0585bb077c423107d5e
0b30f43feb2c7443cf708944f1cc679b18f64297
cb8aa8919b0edacccc9646f58188af7ccc8d5ae8
```

Let me fiddle with this...

PR #852, opened 2025-12-03
Correctly mapping material id to block id for OBJ file reader

wdhawkins:
The OBJ file reader treats material ids as integers with the potential for a material id of -1. Assigning a -1 to block id (unsigned int) causes a sanitizer failure. This PR correctly maps integer material ids to unsigned int block ids.


wdhawkins:
> This works, but `std::optional<unsigned int> material_id` would be probably better. Your call if you want to do that...

Time to get out of the c++11 era.  :smile: I've refactored this patch to use `std::optional`.

PR #859, opened 2025-12-09
Adding time-dependent solver and time-dependent sources

wdhawkins:
This PR adds a number of enhancements:

1. Adds support for time-dependent problems with `TimeDependentSourceSolver` (theta-weighted time discretization)
2. Adds support for simple time-dependence to volumetric and point sources
3. Allows for use of Krylov and Classic Richardson solvers with time-dependent problems (Krylov was tricky)
4. Simplifies sweepchunk constructors
5. Adds time-dependent regression tests with analytic/semi-analytic solutions (see comments in regression test files for details)
6. Exposes `TimeDependentSourceSolver` methods via the Python API so that users can write their own time stepping loops in Python (see `transport_td/transport_3d_2g_inf_med_pydrvr.py` regression test)

Still to do:

1. Support for tiime-dependent boundary conditions (I have other boundary work to get to, so I'll do this relatively soon)
2. Both sources and bcs need to allow time-dependent, user-supplied functions
3. TD support for RZ
4. The number of sweepchunk sources is multiplying. We need to explore consolidating without a performance penalty.
5. GPU support for TD problems.
6. The user should be able to specify theta as a numeric value or use `BackwardEuler` or `CrankNicolson` identifiers.
7. Generate an OpenMC xs file for testing.


modules/linear_boltzmann_solvers/solvers/time_dependent_solver.cc
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  ags_solver_->Solve();
+  lbs_problem_->UpdateFieldFunctions();
+  lbs_problem_->UpdatePsiOld();
+
+  current_time_ = t_end;
+  lbs_problem_->SetSimulationTime(current_time_);
+  ++step_;
+}
```
ragusa: should we have hooks to let the user can things, like material properties, etc... as a function of time? I think in MOOSE there are pre-solve and post-solve hooks

modules/linear_boltzmann_solvers/discrete_ordinates_curvilinear_problem/sweep_chunks/aah_sweep_chunk_rz.cc
```diff
@@ -1,44 +1,34 @@
...
                groupset,
-               xs,
-               num_moments,
-               max_num_cell_dofs,
-               min_num_cell_dofs),
-    secondary_unit_cell_matrices_(secondary_unit_cell_matrices),
+               problem.GetBlockID2XSMap(),
+               static_cast<int>(problem.GetNumMoments()),
+               static_cast<int>(problem.GetMaxCellDOFCount()),
+               static_cast<int>(problem.GetMinCellDOFCount())),
```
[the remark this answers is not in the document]
wdhawkins: These all return `size_t`. Let me look at this.

modules/linear_boltzmann_solvers/lbs_problem/point_source/point_source.h
```diff
@@ -64,12 +65,17 @@ class PointSource
...
+  double end_time_ = std::numeric_limits<double>::infinity();
+
   std::vector<Subscriber> subscribers_;
   size_t num_global_subscribers_ = 0;
 
 public:
   static InputParameters GetInputParameters();
   static std::shared_ptr<PointSource> Create(const ParameterBlock& params);
+
+  bool IsActive(double time) const;
```
andrsd: This should be moved to the public API section at the beginning of the `class`.

modules/linear_boltzmann_solvers/lbs_problem/volumetric_source/volumetric_source.h
```diff
@@ -68,6 +72,9 @@ class VolumetricSource
 
 private:
   static int next_id_;
+
+public:
+  bool IsActive(double time) const;
```
andrsd: This should be moved to the public API section at the beginning of the `class`.

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.h
```diff
@@ -235,6 +264,9 @@ class LBSProblem : public Problem
    */
   virtual void ReorientAdjointSolution() {};
 
+  /// Copy psi_new to psi_old
```
andrsd: This looks misplaced, `DiscreteOrdinatesProblem` does this, not `LBSProblem`

modules/linear_boltzmann_solvers/solvers/pi_keigen_solver.cc
```diff
@@ -5,8 +5,8 @@
 #include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.h"
 #include "modules/linear_boltzmann_solvers/lbs_problem/lbs_vecops.h"
 #include "modules/linear_boltzmann_solvers/lbs_problem/lbs_compute.h"
-#include "framework/logging/log_exceptions.h"
 #include "framework/logging/log.h"
+#include <stdexcept>
```
andrsd: Should be at the end of the `#include` list

modules/linear_boltzmann_solvers/solvers/time_dependent_solver.cc
```diff
@@ -0,0 +1,126 @@
+// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
```
andrsd: 2024 -> 2025

modules/linear_boltzmann_solvers/solvers/time_dependent_solver.cc
```diff
@@ -0,0 +1,126 @@
...
+
+void
+TimeDependentSourceSolver::Advance()
+{
+  const double t_start = current_time_;
+  const double t_end = current_time_ + dt_;
+
+  lbs_problem_->SetTimeStep(dt_);
+  lbs_problem_->SetTheta(theta_);
+  lbs_problem_->SetSimulationTime(t_start);
```
andrsd: Is there going to be another time than simulation time? 
- If so, then `SetTimeStep` may need to be `SetSimulationTimeStep`. 
- If not, then `SetTime` should be enough, right?

modules/linear_boltzmann_solvers/solvers/pi_keigen_solver.cc
wdhawkins: Fixed.

modules/linear_boltzmann_solvers/solvers/time_dependent_solver.cc
wdhawkins: Fixed.

test/python/modules/linear_boltzmann_solvers/transport_steady/simple_1g_td.cxs
```diff
@@ -0,0 +1,16 @@
+# Manually generated XS
+# Date : Sep 27, 2023
```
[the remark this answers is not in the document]
wdhawkins: Removed this entire file. It was no longer needed since the td tests moved to their own directory.

test/python/modules/linear_boltzmann_solvers/transport_td/simple_1g_v0.5_td.cxs
ragusa: I see you do not use the built-in function to compute simple 1g XS. Why not use the built-in function, even if you need to add velocity to it, or maybe we do not want to make the built-in 1g creator more complicated?

test/python/modules/linear_boltzmann_solvers/transport_td/simple_2g_downscatter_td.cxs
```diff
@@ -0,0 +1,24 @@
...
+0 2.0    # v0
+1 0.5    # v1
+VELOCITY_END
+
+TRANSFER_MOMENTS_BEGIN
+# M_GPRIME_G_VAL ell gprime g value  (ell = 0 only)
+# group 0 -> 0: no self-scatter
+M_GPRIME_G_VAL   0   0   0   0.0
+# group 0 -> 1: downscatter
+M_GPRIME_G_VAL   0   0   1   0.5
```
ragusa: looks like gprime is g_departure and g is g_arrival. if you quiz me out of the blue on this, I would have said the opposite.

test/python/modules/linear_boltzmann_solvers/transport_td/transport_3d_2g_inf_med_pydrvr.py
```diff
@@ -0,0 +1,152 @@
...
+    current_time = 0.0
+    step = 0
+
+    while current_time < stop_time:
+        target_time = min(current_time + dt, stop_time)
+        step_dt = target_time - current_time
+
+        # Configure the time step on the solver
+        solver.SetTimeStep(step_dt)
+        solver.SetTheta(theta)
```
ragusa: should this be inside the loop?

test/python/modules/linear_boltzmann_solvers/transport_td/transport_3d_2g_inf_med_pydrvr.py
```diff
@@ -0,0 +1,152 @@
...
+                f"*************** Time step #{step:d}  t = {target_time:.6f} "
+                f"(from {current_time:.6f}, dt = {step_dt:.6f}, theta = {theta:.3f}) "
+                f"***************"
+            )
+
+        # Advance the solution from current_time to target_time
+        solver.Advance()
+
+        fflist = phys.GetScalarFieldFunctionList()
+        monitor_volume = RPPLogicalVolume(infx=True, infy=True, infz=True)
```
ragusa: should these be repeated inside the loop. If so, why?

test/python/modules/linear_boltzmann_solvers/transport_td/transport_3d_2g_inf_med_pydrvr.py
```diff
@@ -0,0 +1,152 @@
...
+    from pyopensn.logvol import RPPLogicalVolume
+
+if __name__ == "__main__":
+
+    meshgen = FromFileMeshGenerator(filename="../../../../assets/mesh/cube3.2.msh")
+    grid = meshgen.Execute()
+    grid.SetUniformBlockID(0)
+    grid.SetOrthogonalBoundaries()
+
+    num_groups = 2
```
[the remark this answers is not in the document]
wdhawkins: Fixed. We don't have a way in the input to check that it matches the xs data, but I'm pretty sure we have sanity check in the problem setup.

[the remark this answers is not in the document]
wdhawkins: Maybe we should have something like `num_groups = xs_diag.GetNumGroups()`. That gets a bit messy and convoluted with lots of xs data files though.

test/python/modules/linear_boltzmann_solvers/transport_td/transport_3d_2g_inf_med_pydrvr.py
```diff
@@ -0,0 +1,152 @@
...
+            {"name": "zmin", "type": "reflecting"},
+            {"name": "zmax", "type": "reflecting"},
+        ],
+        options={
+            "verbose_inner_iterations": False,
+        },
+        time_dependent=True,
+    )
+
+    # Create the time-dependent solver without stop_time, we will loop in Python
```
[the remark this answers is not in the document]
wdhawkins: At least for the simple problem we're testing here, there is no appreciable difference between the two. `Advance` does all of the work, and it's the same regardless of whether the driver is the c++ `Execute` function or Python. IMHO, using Python to drive the time step loop is the way we should go. It allows us all of the flexibility of an event system without needing an event system.

test/python/modules/linear_boltzmann_solvers/transport_td/transport_3d_2g_inf_med_pydrvr.py
wdhawkins: They should not. Fixes.

quocdang1998 (approved):

PR #863, opened 2025-12-11
Replacing our bespoke SCC algorithm with Boost. Improving global sweep graph cycle breaking

wdhawkins:
The PR makes the following improvements:

1. Removes our bespoke SCC algorithm in favor of Boost's algorithm
2. Refactors our FAS algorithm to fix hangs when using edge weights and to improve performance
3. Adds edges weights to the global sweep graph to better inform edge removal
  - For each partition and SPDS, we compute the accumulated outgoing edge weight from the current partition to a neighboring partition, loc, as: `W(loc) = sum( (dot(omega, n_f)^2) * area_f for all outgoing faces f whose neighbor is on location loc )`. This weight is summed across all SPDS's and used as the edge weight for the global sweep graph.
  - These edge weights are used in the FAS algorithm to create the global sweep DAG. With this edge weighting, we can reduce the number of extra iterations caused by lagging.

Note that this changes the sweep ordering slightly and required modifying the gold values for some regression tests.

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/spds/spds.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/spds/spds.cc
@@ -8,9 +8,11 @@
 #include "framework/runtime.h"
 #include "caliper/cali.h"
 #include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/strong_components.hpp>
 #include <algorithm>
 #include <vector>
 #include <queue>
+#include <limits>
 
 namespace opensn
 {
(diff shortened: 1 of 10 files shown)


test/python/modules/linear_boltzmann_solvers/transport_adjoint/tests.json
ragusa: I see the note about the sweep ordering but causes the change in gold value. I am wondering if the old and the new version of the code, under much tighter tolerances in both cases, would then converge to the same value. Can we just do one such check?

wdhawkins: Yes, they should, I'll check a test or two.

wdhawkins: I tightened up the tolerances and re-ran all of the tests that required gold value modifications. They all converge to the same answers with main and this PR.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/spds/spds.cc
```diff
@@ -20,77 +22,124 @@ SPDS::FindApproxMinimumFAS(Graph& g, std::vector<Vertex>& scc_vertices)
...
+      auto w = boost::target(e, g);
+      if (in_component[w])
+        return true;
+    }
+    return false;
+  };
 
-    // Get the edge weight property map
-    auto weightmap = boost::get(boost::edge_weight, g);
+  auto has_incoming_in_component = [&](Vertex v)
```
andrsd: `has_incoming_in_component` -> `HasIncomingInComponent` (lambda `PascalCase`)

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/spds/spds.cc
```diff
@@ -20,77 +22,124 @@ SPDS::FindApproxMinimumFAS(Graph& g, std::vector<Vertex>& scc_vertices)
...
+  using boost::make_iterator_range;
+
+  const auto num_vertices_g = boost::num_vertices(g);
+  std::vector<bool> in_component(num_vertices_g, false);
+  for (auto v : scc_vertices)
+    in_component[v] = true;
+
+  auto weightmap = boost::get(boost::edge_weight, g);
+
+  auto has_outgoing_in_component = [&](Vertex v)
```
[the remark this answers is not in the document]
wdhawkins: Fixed.

quocdang1998 (approved):

PR #877, opened 2025-12-18
Fixing clang-tidy internal linkage errors

wdhawkins:


andrsd (approved):

wdhawkins:
> Any thoughts on using anonymous namespace for these things?

I'd be ok with that. I'll look at doing that in a future PR.

Issue #858, opened 2025-12-08
arbitrary boundary condition


ragusa:
# Arbitrary Boundary Condition via Python Function (Incoming Angular Flux)

## Goals
Enable a new boundary condition type that allows the user to bind a **boundary ID** to a **user-supplied Python function** defining the **incoming angular flux**. In the v1 implementation, the function provides values that are **uniform in space** over all faces with that boundary ID and depends only on **energy group index** and **direction index**. The solver uses these values to set the boundary inflow for directions with $\Omega \cdot n < 0$ on the tagged boundary faces. A spatially dependent extension (function of face coordinates) may be added later.

- Provide a flexible mechanism for user-defined inflow angular flux without hard-coding spectral/anisotropic sources.
- Support boundary inflow that varies by **energy group** and **discrete ordinate direction**.
- Keep v1 simple and robust (no spatial dependence).

---

## Scope

**In scope (v1)**
- New BC type: `python_inflow` (name TBD).
- Association of a single boundary ID with a Python callback.
- Callback returns boundary inflow **uniform across that boundary**.
- Energy- and angle-dependent values:
  - per group `g`
  - per direction index `d`
- Time-dependent inflow.

**Out of scope (v1)**
- Spatially varying inflow on a boundary (coordinate-dependent evaluation).
- Boundary conditions defined per face element rather than boundary ID.

PR #878, opened 2025-12-19
Adding arbitrary boundary functions that depend on energy group index and angle index

wdhawkins:
Addresses the requirements of v1 of #858 without time dependence. I'll introduce time-dependent boundary conditions in a future commit.

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.h
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.h
@@ -1,0 +1,67 @@
+// SPDX-FileCopyrightText: 2025 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/sweep_boundary.h"
+#include "framework/math/functions/function.h"
+#include "framework/data_types/ndarray.h"
+#include <memory>
+
+namespace opensn
+{
+
+/// Arbitrary incident angular flux specified by a user-supplied function.
+///
+/// The AngularFluxFunction is expected to return the incoming angular flux for a given
+/// energy group index and quadrature direction index. The returned values are uniform for
+/// all faces with this boundary id.
+class ArbitraryBoundary : public SweepBoundary
+{
+public:
+  ArbitraryBoundary(size_t num_groups,
+                    std::shared_ptr<AngularFluxFunction> angular_flux_function,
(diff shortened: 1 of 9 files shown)


modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.h
```diff
@@ -0,0 +1,67 @@
...
+#include "framework/data_types/ndarray.h"
+#include <memory>
+
+namespace opensn
+{
+
+/// Arbitrary incident angular flux specified by a user-supplied function.
+///
+/// The AngularFluxFunction is expected to return the incoming angular flux for a given
+/// energy group index and quadrature direction index. The returned values are uniform
```
andrsd: ...are uniform **for** all faces... ?

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_ortho_arbitrary_bc.py
```diff
@@ -0,0 +1,113 @@
...
+    norm = 0.0
+    sum = 0.0
+
+    for i, (omega, w) in enumerate(zip(pquad.omegas, pquad.weights)):
+        ox = omega.x
+        oy = omega.y
+        oz = omega.z
+        if ox > 0.0 and oy > 0.0:
+            angle_mask[i] = True
+            sum += ox * w * num_groups
```
ragusa: this line is obscure. what is the purpose of sum? why is it multiplied ny num_groups? If you want to normalize so that the leakage/current is 1., why multiply by num_groups?

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_ortho_arbitrary_bc.py
ragusa: This file should also be added as a tutorial (we are going to have more tutorials, especially since we will be doing LLM/RAG soon). I think we should compute the leakage on all faces for a void and show that nothing leaks from the bottom and that the sum of the leakage top and right is 1. I would also make this a rectangular domain

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_ortho_arbitrary_bc.py
wdhawkins: `num_groups` is there to normalize the bc across groups. The bc function returns the same angular flux for every group, and I want the total inflow to equal 1.0.

ragusa: that's ok for a test. It means each group has the same fraction of in-leakage but it is not a uniform spectrum in energy.

ragusa (approved):

PR #884, opened 2025-12-26
Simplifications to time dependent source solver. Adding pre-advance and post-advance hooks

wdhawkins:
Adds pre-advance and post-advance hooks to `TimeDependentSourceSolver::Execute` to improve usability of c++ time-stepping method.

### DIFF
--- a/modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cc
+++ b/modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cc
@@ -153,6 +153,8 @@ LBSProblem::SetTime(double time)
 void
 LBSProblem::SetTimeStep(double dt)
 {
+  if (dt <= 0.0)
+    throw std::runtime_error(GetName() + " dt must be greater than zero.");
   dt_ = dt;
 }
 
@@ -171,6 +173,8 @@ LBSProblem::IsTimeDependent() const
 void
 LBSProblem::SetTheta(double theta)
 {
+  if (theta < 0.0 or theta > 1.0)
+    throw std::runtime_error(GetName() + " theta must be between 0.0 and 1.0.");
   theta_ = theta;
 }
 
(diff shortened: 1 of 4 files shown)


modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cc
```diff
@@ -153,6 +153,8 @@ LBSProblem::SetTime(double time)
 void
 LBSProblem::SetTimeStep(double dt)
 {
+  if (dt < 0.0)
+    throw std::runtime_error(GetName() + " dt must be non-negative");
```
ragusa: can dt be = 0? should we not check for strict positivity?

modules/linear_boltzmann_solvers/solvers/time_dependent_solver.h
```diff
@@ -25,6 +27,14 @@ class TimeDependentSourceSolver : public Solver
 
   void Advance() override;
 
+  void PreAdvance(std::function<void()> callback);
+
+  void PreAdvance(std::nullptr_t);
+
+  void PostAdvance(std::function<void()> callback);
+
+  void PostAdvance(std::nullptr_t);
```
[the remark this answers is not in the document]
wdhawkins: Yep. Fixed.

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cc
ragusa: I still see `if (dt < 0.0)`, meaning that zero is still acceptable. Is it?

wdhawkins: I think you happened to look at the code before I pushed the update. I updated the comment first and then pushed the changes. :)  It should be correct now.

ragusa (approved):

PR #888, opened 2026-01-07
Fixing comment

### LINKED ISSUE #837: incorrect comment
https://github.com/Open-Sn/opensn/blob/576bf19b722a754ed6f9ec137abcc5ecd13715e3/framework/math/quadratures/angular/product_quadrature.cc#L152

The positive polar angles are between 0 and pi, not lss than pi/2

andrsd:
Closes #837


andrsd:
> I do not think the change in the comment is correct.

I removed the text in parentheses. This matches the commenting style in the function - just say what the code below is going to do...

PR #889, opened 2026-01-07
Fix PETSc build issue when using cmake 4

### LINKED ISSUE #761: Forbid CMake 4. in installation guide
The current CMake install script does not work with CMake >= 4.0.

Reason: PETSc cannot configure METIS with CMake 4.

A warning on the maximum version of CMake should be added to the documentation.

andrsd:
Closes #761


wdhawkins:
> So adding `--download-cmake` works and PETSc does not install it, so it does not override the cmake installed by the system package manager.
> 
> Also, I found out that PETSc 3.24 uses cmake 4.1.1, so updating our PETSc version would fix this as well.
> 
> I am gonna change this PR to add `--download-cmake` for right now...

I've also noticed that recent versions of at least `gcc` have issues compiling `PTSCOTCH` in versions of `PETSc` prior to 3.24. We should probably bump the recommended version in the docs too.

PR #891, opened 2026-01-12
Removing one level of std::vector in AngleAggregation

quocdang1998:
In this PR:

- One level of `std::vector` in `AngleAggregation` is removed to simplify the code syntax. `AngleSet`s can be accessed through `AngleAggregation`'s iterators.
- Asynchronous copy is added to caribou. `caribou::Stream` and `caribou::Event` can be directly used as if they are `cudaStream_t` and `cudaEvent_t` **without casting or using member accessing methods**.
- HIP and CUDA API are unified under `caribou`. HIP adaptation is also added to CMake script. Regression tests passed with HIP compiler on NVIDIA system.
- Add multi-threading mode to `MPI_Init` for Python console and test executable.
- Add __restrict__ to all device pointers to speed up GPU sweep.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,12 +81,27 @@ _add_cxx_cfg(Native "-fno-trapping-math")
 _add_cxx_cfg(Native "-ffp-contract=fast")
 
 option(OPENSN_WITH_CUDA "Enable CUDA support" OFF)
+option(OPENSN_WITH_HIP "Enable HIP support" OFF)
+if (OPENSN_WITH_CUDA AND OPENSN_WITH_HIP)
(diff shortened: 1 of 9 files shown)


python/main.cc
```diff
@@ -12,7 +12,11 @@ namespace mpi = mpicpp_lite;
 int
 main(int argc, char** argv)
 {
+#ifdef __OPENSN_USE_CUDA__
+  mpi::Environement env(argc, argv, mpi::Environment::Threading::MULTIPLE);
```
andrsd: You will want to update CMakeLists.txt around line 100 to say:
```
find_package(mpicpp-lite 2.2 REQUIRED)
```

andrsd (changes requested):

PR #896, opened 2026-01-14
Number of groups is `unsigned int`

andrsd:
Refs #90

### DIFF
--- a/modules/linear_boltzmann_solvers/lbs_problem/acceleration/wgdsa.cc
+++ b/modules/linear_boltzmann_solvers/lbs_problem/acceleration/wgdsa.cc
@@ -19,7 +19,7 @@ WGDSA::Init(DiscreteOrdinatesProblem& do_problem,
   if (groupset.apply_wgdsa)
   {
     // Make UnknownManager
-    const size_t num_gs_groups = groupset.groups.size();
+    const auto num_gs_groups = groupset.GetNumGroups();
     opensn::UnknownManager uk_man;
     uk_man.AddUnknown(UnknownType::VECTOR_N, num_gs_groups);
 
@@ -30,7 +30,7 @@ WGDSA::Init(DiscreteOrdinatesProblem& do_problem,
     // Make xs map
     const auto& block_id_to_xs_map = do_problem.GetBlockID2XSMap();
     auto matid_2_mgxs_map =
-      PackGroupsetXS(block_id_to_xs_map, groupset.groups.front(), groupset.groups.back());
+      PackGroupsetXS(block_id_to_xs_map, groupset.first_group, groupset.last_group);
 
     // Create solver
     const auto& sdm = do_problem.GetSpatialDiscretization();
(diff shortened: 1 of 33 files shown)


ragusa:
> The only reason for `std::size_t` is that `LBSProblem` stores groups like this:
> 
> ```
>   std::vector<LBSGroup> groups_;
> ```
> 
> which I admit is somewhat silly. It kinda says that groups might be stored in random order, but I don't think the code is capable of handling that. If you always want to iterate over groups from `0..G`, I can remove the container and use a smaller integer type.

Pinging @wdarylhawkins to get his opinion, since both you and I think `size_t` is silly.

In forward mode, we go `0..G`. In adjoint mode, we may go `G..0`.

framework/materials/multi_group_xs/multi_group_xs.cc
```diff
@@ -412,7 +412,7 @@ MultiGroupXS::TransposeTransferAndProduction()
   {
     const auto& S_ell = transfer_matrices_[ell];
     SparseMatrix S_ell_transpose(num_groups_, num_groups_);
-    for (size_t g = 0; g < num_groups_; ++g)
+    for (unsigned int g = 0; g < num_groups_; ++g)
     {
       const size_t row_len = S_ell.rowI_indices[g].size();
       const size_t* col_ptr = S_ell.rowI_indices[g].data();
```
ragusa: those stay size_t ?

framework/materials/multi_group_xs/opensn_export.cc
```diff
@@ -100,7 +100,7 @@ MultiGroupXS::ExportToOpenSnXSFile(const std::string& file_name, const double fi
       ofile << "\nCHI_DELAYED_BEGIN\n";
       const auto& precursors = GetPrecursors();
       for (size_t j = 0; j < GetNumPrecursors(); ++j)
```
ragusa: size_t

modules/linear_boltzmann_solvers/lbs_problem/acceleration/wgdsa.cc
```diff
@@ -74,7 +74,7 @@ WGDSA::AssembleDeltaPhiVector(DiscreteOrdinatesProblem& do_problem,
   const auto& phi_uk_man = do_problem.GetUnknownManager();
   const auto& block_id_to_xs_map = do_problem.GetBlockID2XSMap();
 
-  const auto gsi = groupset.groups.front().id;
+  const auto gsi = groupset.groups.front();
   const size_t gss = groupset.groups.size();
```
ragusa: size_t?

modules/linear_boltzmann_solvers/lbs_problem/groupset/lbs_groupset.cc
```diff
@@ -144,7 +144,7 @@ LBSGroupset::LBSGroupset( // NOLINT(cppcoreguidelines-pro-type-member-init)
   {
     for (size_t g = from; g <= to; ++g)
```
ragusa: size_t?

test/python/framework/tutorials/tutorial_06_wdd.cc
```diff
@@ -384,7 +384,7 @@ SetSource(const std::shared_ptr<MeshContinuum> grid,
   std::vector<double> source_moments(num_local_phi_dofs, 0.0);
 
   const size_t num_moments = phi_uk_man.unknowns.size();
```
ragusa: can you make a note to bring this up in a meeting and simply create an issue right now. I wonder if there are not other quantities that also should not be size_t, like moments, number of precursors, ....

framework/materials/multi_group_xs/opensn_export.cc
andrsd: So, I looked at the code for "fixing" number of precursors. I unified them to `std::size_t` since they are stored in `std::vector`, but I can see if I can do `unsigned int` instead, if that would be preferred...

ragusa (approved):
I suppose there will be another PR for num_angles and maybe a few other vars that are still size_t but could be changed to unsigned int

PR #898, opened 2026-01-17
Simplifying AAH sweep chunks to reduce duplicated code

### LINKED ISSUE #864: Enhancements for time-dependent problems
- [ ] Support for tiime-dependent boundary conditions
- [ ] Both sources and bcs need to allow time-dependent, user-supplied functions
- [ ] TD support for RZ
- [ ] The number of sweepchunk sources is multiplying. We need to explore consolidating without a performance penalty
- [ ] GPU support for TD problems
- [x] The user should be able to specify theta as a numeric value or use BackwardEuler or CrankNicolson identifiers
- [x] Add Python callbacks to `Advance` method
- [x] Generate an OpenMC xs file for testing
- [ ] Ability to swap block-xs map during time-dependent calculations

wdhawkins:
Partially resolves #864

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cc
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 #include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.h"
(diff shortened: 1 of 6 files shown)


modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aah_sweep_chunk.cc
```diff
@@ -92,236 +74,29 @@ AAHSweepChunk::Sweep(AngleSet& angle_set)
 void
 AAHSweepChunk::CPUSweep_Generic(AngleSet& angle_set)
 {
```
quocdang1998: `GPUSweep` have been isolated in another class after PR 908. You can consider replacing all the `CPUSweep` occurrence with just `Sweep`.

PR #911, opened 2026-01-29
Adding ability to switch solver mode between steady state and time dependent

wdhawkins:


modules/linear_boltzmann_solvers/discrete_ordinates_problem/iterative_methods/sweep_wgs_context.h
```diff
@@ -29,8 +29,10 @@ struct SweepWGSContext : public WGSContext
 
   void PostSolveCallback() override;
 
+  void ResetSweepChunk(std::shared_ptr<SweepChunk> new_chunk);
+
```
quocdang1998: Why do we need to `ResetSweepChunk` while we're not using it at all?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/iterative_methods/sweep_wgs_context.h
```diff
@@ -30,7 +30,7 @@ struct SweepWGSContext : public WGSContext
   void PostSolveCallback() override;
 
   std::shared_ptr<SweepChunk> sweep_chunk;
-  SweepScheduler sweep_scheduler;
+  std::unique_ptr<SweepScheduler> sweep_scheduler;
```
quocdang1998: Is there a specific reason why `SweepScheduler` is changed to `std::unique_ptr`?

Since we don't need to reset the `SweepScheduler` anymore, I think there's no need for converting to `std::unique_ptr` here.

quocdang1998 (approved):

Issue #856, opened 2025-12-08
volumetric postprocessors


ragusa:
# Volumetric Post-Processor (VPP) Specification

## 
The Volumetric Post-Processor (VPP) computes scalar-flux-based volumetric quantities over a user-defined spatial subset of cells. The spatial subset is defined by an optional set of material IDs and an optional list of logical volumes; if both are provided, a cell is included only if its material ID matches **and** it belongs to **at least one** of the specified logical volumes. The VPP applies an optional multiplier (constant or multigroup response, including cross sections used by the solver or external response cross sections such as energy deposition or DPA) and then produces one or more spatial reductions (volume integral, volume average, min, max). By default, results are summed over all energy groups, with options to return per-group results or user-defined group-set (macro-group) collapsed results.

- Provide a consistent, minimal, and extensible mechanism to compute volumetric QoIs from **scalar flux**.
- Clearly separate volumetric post-processing from surface/current-based post-processing.
- Support both solver cross sections and non-solver response functions.

---

## Scope
- Scalar flux input only: $\phi_{i,g}$ per dof and energy group.
- Spatial filtering by material IDs and logical volumes.
- Energy aggregation modes.
- Multiplier/response applied to scalar flux.
- Global reductions over the selected cells.

---

## Definitions
Let:
- $c$ be a cell, $g\in\{1,\dots,G\}$ an energy group.
- $V_c$ the cell volume.
- $\phi_{i,g}$ the scalar flux in cell $c$, group $g$. $i \in DOF(c)$
- $R_{c,g}$ the multiplier/response value for cell $c$, group $g$.
- $S$ the selected set of cells after spatial filtering.

Per-cell quantity:
$
q_{c,g} = R_{c,g} \sum_{i\in DOF(c)} w_i\phi_{i,g}
$

---

### User inputs
- `material_ids` (optional list of integers)
- `logical_volumes` (optional list of strings)

### Inclusion rule
A cell $c$ is included if:
- `material_ids` is not provided **or** `material(c) ∈ material_ids`,
**AND**
- `logical_volumes` is not provided **or** $c$ belongs to **at least one** listed logical volume.

---

### Default
- `energy.mode = all_groups`  
  Return results summed over all groups.

### Supported modes
- `all_groups`

   $   Q = \sum_{g=1}^G Q_g   $
- `per_group`
   Return $Q_g$ for all $g$.
- `group_sets`
   User provides named group sets $\mathcal{G}_k$:

[...]

PR #916, opened 2026-02-02
Adding volume postprocessor

andrsd:
- [x] add user doco

Refs #856


ragusa (approved):

wdhawkins:
Can you add a section in the post processors chapter of the user guide on how to use this? @andrsd

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_gmsh_blk_bnds_v4.py
```diff
@@ -30,7 +30,7 @@
     num_groups = 64
     vol0 = RPPLogicalVolume(infx=True, infy=True, infz=True)
     xs_diag = MultiGroupXS()
-    xs_diag.LoadFromOpenSn("diag_XS_64g_1mom_c0.99.xs")
+    xs_diag.LoadFromOpenSn("../../../../assets/xs/diag_XS_64g_1mom_c0.99.xs")
```
quocdang1998, line 33: If we move `diag_XS_64g_1mom_c0.99.xs`, we should also moved all xs files (`xs_168g.xs`, `xs_graphite_pure`, etc) to the `assets` folder too.

python/lib/post.cc
```diff
@@ -0,0 +1,85 @@
...
+void
+WrapPostprocessors(py::module& post)
+{
+  // clang-format off
+  // Volume post-processor value type enum
+  py::enum_<VolumePostprocessor::ValueType>(post, "VolumePostprocessorValueType")
+    .value("INTEGRAL", VolumePostprocessor::ValueType::INTEGRAL)
+    .value("MAX", VolumePostprocessor::ValueType::MAX)
+    .value("MIN", VolumePostprocessor::ValueType::MIN)
+    .value("AVERAGE", VolumePostprocessor::ValueType::AVERAGE);
```
quocdang1998: Don't use Python enum. Use `std::map<std::string, VolumePostprocessor::ValueType>` and `std::map::at` instead.

Enums are not native to Python. NumPy and SciPy doesn't even have enum. Python user would prefer string for enum inputs more than true enums.

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
```diff
@@ -0,0 +1,424 @@
...
+    throw std::invalid_argument("'value_type' can be only 'min', 'max', 'integral', or 'avg'");
+
+  CreateSpatialRestriction();
+  CreateEnergyRestriction();
+  CreateMultipliers();
+
+  auto n_lvs = std::max(static_cast<std::size_t>(1), logical_volumes_.size());
+  values_.resize(n_lvs);
+  for (auto& vals : values_)
+    vals.resize(groups_.size());
```
quocdang1998: Since values is an 2D array, not a vector of vector, I think it would be better to use OpenSn `NDArray` for values.

It would also make it easier to directly convert (without copying) a 2D array to Python NumPy instead of having a list of lists.

quocdang1998 (approved):

python/lib/post.cc
quocdang1998: On top of this, I didn't see any usage of these enums anywhere in the code. Are they really necessary to export?

wdhawkins (approved):

test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_gmsh_blk_bnds_v4.py
andrsd: I plan on doing so in a follow up PR.

andrsd:
> Can you add a section in the post processors chapter of the user guide on how to use this?

Wil do.

python/lib/post.cc
andrsd: Yes. I stared with this, but I did not use this in the end. Thanks for pointing this out. I forgot to remove this piece.

PR #926, opened 2026-02-09
Adding a mesh reader for Gmsh v4.1 binary format

wdhawkins:


test/python/modules/linear_boltzmann_solvers/transport_steady/transport_2d_gmsh_v4_binary.py
ragusa: out of curiosity, did you also try to generate the gmsh file as ASCII and see that when read in as ASCII, it gave the exact same answer as the binary. It would be good to have a test that reads in the mesh generated both in ascicc and binary, and check that the difference in the solution is 0.

PR #928, opened 2026-02-10
Adding angular flux field function

wdhawkins:


ragusa (commented):
Just wondering if we should push a test case.

wdhawkins:
> Just wondering if we should push a test case.

Done. I've added `transport_steady/transport_3d_disk_scalar_angular_ff.py` as an example. It doesn't currently run as part of our regression tests. We'll need to figure out the best way to add diff'ing of images to our regression testing.

ragusa (approved):
Nice. Thanks!!

andrsd (approved):

PR #935, opened 2026-02-14
Updating qa clang-format workflow to run locally

wdhawkins:
ghcr.io seems to be having issues... at least for this workflow. Moving this to the local runner.

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -8,16 +8,25 @@ on:
 jobs:
   check-formatting:
     name: Check formatting
-    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -el {0}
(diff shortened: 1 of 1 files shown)


.github/workflows/qa.yaml
```diff
@@ -8,16 +8,25 @@ on:
...
-    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -el {0}
+    runs-on: self-hosted
     steps:
       - uses: actions/checkout@v4
 
       - name: clang-format style check
-        uses: jidicula/clang-format-action@v4.15.0
```
ragusa: I always wondered what that was. Some third-party CI stuff. It could be risky to use such stuff, no?

andrsd (approved):

PR #936, opened 2026-02-15
AAH GPU sweep enhancements

wdhawkins:
Performance enhancements for AAH GPU sweeps. These changes give 4-5X improvement on AMD MI300.

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/aahd_angle_set.cu
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/aahd_angle_set.cu
@@ -4,9 +4,8 @@
 #include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/aahd_angle_set.h"
 #include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aahd_sweep_chunk.h"
 #include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.h"
-#include "framework/logging/log.h"
-#include "framework/runtime.h"
 #include "caliper/cali.h"
+#include <algorithm>
 
 namespace opensn
 {
@@ -27,6 +26,10 @@ AAHD_AngleSet::AAHD_AngleSet(size_t id,
   std::dynamic_pointer_cast<AAHD_FLUDS>(fluds_)->GetStream() = stream_;
   crb::MemoryPinningManager angle_indices_pinner_(angles_);
   crb::copy(device_angle_indices_, angle_indices_pinner_, angles_.size());
+  has_reflecting_boundaries_ =
+    std::any_of(boundaries_.begin(),
+                boundaries_.end(),
+                [](auto& bndry_pair) { return bndry_pair.second->IsReflecting(); });
 }
 
 void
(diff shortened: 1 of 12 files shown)


andrsd (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/scheduler/sweep_scheduler.h
```diff
@@ -87,6 +91,17 @@ class SweepScheduler
...
+  // Persistent worker threads for AAO GPU scheduling.
+  std::vector<std::thread> aao_worker_threads_;
+  std::mutex aao_mutex_;
+  std::condition_variable aao_cv_start_;
+  std::condition_variable aao_cv_done_;
+  [[maybe_unused]] bool aao_workers_initialized_ = false;
+  bool aao_stop_workers_ = false;
+  [[maybe_unused]] std::size_t aao_workers_remaining_ = 0;
+  std::size_t aao_work_epoch_ = 0;
+  [[maybe_unused]] SweepChunk* aao_active_sweep_chunk_ = nullptr;
```
quocdang1998: **Update 1: changes in sweep_scheduler make the code slower**

I've made a comparison between 1 rank - 1 GPU - 32 iterations (no MPI interaction) on Tuo, and here's the result:

`srun --exclusive -p pdebug -N 1 -n 1 -c 24 --gpus-per-task=1 -t 1:00:00 ./python/opensn -i strong_scaling.py`

Modified sweep scheduler (this PR):
```
[0]         Average sweep time (s):        1.9598
[0]         Sweep Time/Unknown (ns):       0.534303
[0]         Number of unknowns per sweep:  3667951616
```

Old sweep_scheduler version:
```
[0]         Average sweep time (s):        1.7824
[0]         Sweep Time/Unknown (ns):       0.485938
[0]         Number of unknowns per sweep:  3667951616
```

Other files (FLUDS, AngleSet, Kernel) are taken from this PR.

In my opinion, we don't need to use `std::mutex` and `std::condition_variable` at all. `std::latch` is already sufficient to handle the same functionality.

I think we should revert `sweep_scheduler.*` back to the previous version.

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aahd_sweep_chunk.cu
```diff
@@ -52,6 +60,65 @@ UnsaturatedKernel(Arguments args, const std::uint32_t* level, double* saved_psi)
...
+
+    auto& entry = direction_cache[angle_idx];
+    entry.omega[0] = direction_view.omega[0];
+    entry.omega[1] = direction_view.omega[1];
+    entry.omega[2] = direction_view.omega[2];
+    entry.weight = direction_view.weight;
+    entry.m2d = direction_view.m2d;
+    entry.d2m = direction_view.d2m;
+  }
+  __syncthreads();
```
quocdang1998: **Update 2: caching direction data to shared memory does not improve performance**

Here's the comparison between 2 versions of the kernel on Tuo:

 - This PR:
```
   [0]         Average sweep time (s):        1.7824
   [0]         Sweep Time/Unknown (ns):       0.485938
   [0]         Number of unknowns per sweep:  3667951616
```
 - Old kernel:
```
   [0]         Average sweep time (s):        1.78588
   [0]         Sweep Time/Unknown (ns):       0.489613
   [0]         Number of unknowns per sweep:  3667951616
```

The sweep time is pretty much the same.

On top of this, MI300A only has 32kB of shared memory (max), which is **just enough to fit 8 DOFs of a wavefront** (64 * 8 * 8 * 8 bytes). I think we can save memory for caching the linear system rather than caching the direction entry.

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep_chunks/aahd_sweep_chunk.cu
wdhawkins: Same comment as above. Not opposed to reverting... especially this change if we need the shared memory for other purposes. But, I'd like to see that other improvements using shared memory actually result in improvements to sweep time before reverting.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/aahd_angle_set.cu
```diff
@@ -66,66 +65,103 @@ AAHD_AngleSet::AngleSetAdvance(SweepChunk& sweep_chunk, AngleSetStatus permissio
...
 
-  // stream: copy data to device, sweep and copy data back to host buffers
   aahd_fluds->CopyNonLocalIncomingPsiToDevice();
   aahd_fluds->AllocateSaveAngularFlux(aahd_sweep_chunk.GetProblem(),
                                       aahd_sweep_chunk.GetGroupset());
-  sweep_chunk.Sweep(*this);
+  aahd_sweep_chunk.Sweep(*this);
   aahd_fluds->CopyPsiFromDevice();
-  // thread: wait for all sweep computations and data transfer to finish
+  aahd_fluds->CopySaveAngularFluxFromDevice();
```
quocdang1998: In the old version, the reaseon I put `CopySaveAngularFluxFromDevice` after stream synchronization is because I want the copy to happen when the host is copying boundary data into AngleSet.

Since saved_angluar_flux is a large array, copying it will take a lot of time if you synchronise it with the host immediately after you launch the copy.

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/aahd_angle_set.cu
```diff
@@ -66,66 +65,103 @@ AAHD_AngleSet::AngleSetAdvance(SweepChunk& sweep_chunk, AngleSetStatus permissio
...
   auto* aahd_fluds = static_cast<AAHD_FLUDS*>(fluds_.get());
   auto& aahd_sweep_chunk = static_cast<AAHDSweepChunk&>(sweep_chunk);
+  const bool has_reflecting_bc = reflecting_compatible_mode_;
 
-  // Note: Order matters. Simultaneous operations (thread, stream, mpi) must not be separated by
-  // empty lines. Empty lines represent synchronization points with thread.
+  if (has_reflecting_bc)
+    return AdvanceReflectingPath(aahd_sweep_chunk);
+  else
+    return AdvanceNonReflectingOptimizedPath(aahd_sweep_chunk);
```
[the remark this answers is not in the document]
wdhawkins: On a side note, I've noticed that both main and this PR segfault with a GPU memory error on the MI300 with reflecting boundary conditions. I'll open an issue.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/scheduler/sweep_scheduler.h
quocdang1998: I've run the test with:

`srun --exclusive -p pdebug -N 1 -n 4 -c 24 --gpus-per-task=1 -t 1:00:00 ./python/opensn -i strong_scaling.py`

with 32 iterations for each run.

Old sweep scheduler (4 runs):
```
[0]         Average sweep time (s):        7.38668
[0]         Average sweep time (s):        7.55714
[0]         Average sweep time (s):        7.39527
[0]         Average sweep time (s):        7.61228
```
Mean: 7.4878425 +/- 0.11414954

New CV sweep (4 runs):
```
[0]         Average sweep time (s):        7.49419
[0]         Average sweep time (s):        7.20745
[0]         Average sweep time (s):        7.46717
[0]         Average sweep time (s):        7.27784
```
Mean: 7.3616625 +/- 0.14083451

I couldn't get the sweep time down to 3-4s as yours. From what I can see, the improvement is marginal and falls within the noise range.

I'll test on 2 nodes and I'll update the result.

wdhawkins: What do you have `max_mpi_message_size` set to?

quocdang1998: I still have it set at default (32 kB). I'll set it to 256 kB and I'll update the result.

quocdang1998: I do observe the difference thanks to the re-usage of the thread pool. I agree that we should keep this.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/aahd_angle_set.cu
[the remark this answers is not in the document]
wdhawkins: I was testing with `test/python/modules/linear_boltzmann_solvers/transport_keigen/keigenvalue_transport_3d_openmcxs_U235_gpu.py`. It's a fully reflecting orthogonal mesh.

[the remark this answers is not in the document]
wdhawkins: > I perfomred a comparison between the `AdvanceReflectingPath` (by forcing has_reflecting_bc=true) and `AdvanceNonReflectingOptimizedPath`, and I didn't notice any significant improvement at all.
> 
> If you have time, can you perform the same check on your side to see if you can replicate the same thhing?

I see almost a 0.8s improvement in sweep time when running 4 gpus/node. 4.074 s without the reflecting/non-reflecting split and 3.286 s with the split. This was only a sample size of 5. It's not huge, but when we're scrambling for seconds, it's not insignificant either.

quocdang1998 (approved):

PR #945, opened 2026-02-20
Field-function updates

wdhawkins:
Removing `GetFieldFunctions` in favor of field-function specific getters:

- `GetScalarFluxFieldFunction`
- `GetAngularFluxFieldFunction`
- `GetPowerFieldFunction`

Renaming `GetScalarFieldFunction` to `GetScalarFluxFieldFunction`.

### DIFF
--- a/doc/source/tutorials/lbs/keigen/pincell_example.ipynb
+++ b/doc/source/tutorials/lbs/keigen/pincell_example.ipynb
@@ -307,7 +307,7 @@
    },
    "outputs": [],
    "source": [
-    "fflist = phys.GetScalarFieldFunctionList()\n",
+    "fflist = phys.GetScalarFluxFieldFunction()\n",
     "vtk_basename = \"pin_cell\"\n",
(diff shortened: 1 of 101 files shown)


quocdang1998 (approved):

ragusa:
I am surprized there are no keigen test cases. If I remember correctly, with the GetFieldFunctions method, a user can export all scalar fluxes AND power to the same vtk. Is this still possible now? Because we don't have an keigen case (the place where power is most likely computed), I can't tell for sure

wdhawkins:
> I am surprized there are no keigen test cases. If I remember correctly, with the GetFieldFunctions method, a user can export all scalar fluxes AND power to the same vtk. Is this still possible now? Because we don't have an keigen case (the place where power is most likely computed), I can't tell for sure

This works. We can supply a list of field functions to write to VTK. I've added a k-eigenvalue regression test that writes scalar flux and power to a single VTK.

ragusa (approved):

andrsd (approved):

PR #951, opened 2026-02-24
Number of DOF is `std::uint64_t`

andrsd:


framework/math/spatial_discretization/finite_volume/finite_volume.cc
```diff
@@ -286,10 +286,10 @@ FiniteVolume::MapDOFLocal(const Cell& cell,
   return -1;
 }
 
-size_t
+std::uint64_t
 FiniteVolume::GetNumGhostDOFs(const UnknownManager& unknown_manager) const
```
ragusa: Including @wdhawkins do we still have FV? Is it to export Block and Partition IDs?

framework/math/spatial_discretization/finite_element/piecewise_linear/piecewise_linear_continuous.h
```diff
@@ -44,7 +44,7 @@ class PieceWiseLinearContinuous : public PieceWiseLinearBase
     return MapDOFLocal(cell, node, UNITARY_UNKNOWN_MANAGER, 0, 0);
   }
 
-  size_t GetNumGhostDOFs(const UnknownManager& unknown_manager) const override;
+  std::uint64_t GetNumGhostDOFs(const UnknownManager& unknown_manager) const override;
```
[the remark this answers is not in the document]
wdhawkins: `cmake` will fail to configure the project if PETSc isn't compiled with 64-bit integers. We specifically check for that and throw an error.

framework/math/spatial_discretization/finite_volume/finite_volume.cc
wdhawkins: We don't have it as a spatial discretization... except in  a tutorial (old-school tutorial) that implements WDD. I don't think it's needed for block-ids. It was included as a kludge to get partition visualization to work on a single rank. Let me see if we can make that work without needing FV.

wdhawkins (approved):

framework/math/spatial_discretization/finite_volume/finite_volume.cc
ragusa: so we can resolve this discussion and merge now, correct?

ragusa (approved):

PR #955, opened 2026-02-27
Add support for CEPXS cross sections

wdhawkins:
This PR adds support for CEPXS cross sections in `bxslib` format. This gives us the ability to do coupled electron-photon transport. Note that this PR also includes a new field function for reporting energy deposition. This is just a reaction rate field function and can be removed when the post-processing code is merged.  I've added two regression tests from SAND-89-2211 (https://www.osti.gov/servlets/purl/6758232):

**1. II.3.C2 - 1.0 MeV electrons normally incident on an Al slab (no coupling):**
image

**2. II.3.D - 1.0 MeV electrons normally incident on a Cu slab (full coupling):** 
image

### DIFF
--- a/framework/materials/multi_group_xs/cepxs_import.cc
+++ b/framework/materials/multi_group_xs/cepxs_import.cc
@@ -1,0 +1,117 @@
+// SPDX-FileCopyrightText: 2026 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#include "framework/materials/multi_group_xs/multi_group_xs.h"
+#include "framework/runtime.h"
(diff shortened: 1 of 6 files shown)


andrsd (commented):
The 2 tests are using `inp` extension instead of `py`. Why?

wdhawkins:
> The 2 tests are using `inp` extension instead of `py`. Why?

No good reason. :)  Probably just a function of the late hour at which they were added. I'll fix that and add the reference solutions so they pass.

wdhawkins:
> The 2 tests are using `inp` extension instead of `py`. Why?

Fixed.

andrsd (approved):

ragusa (approved):

PR #981, opened 2026-03-17
Improving density handling

### LINKED ISSUE #161: Add cell-wise density tests once the infrastructure supports setting them
> @zhardy-lanl can you create an issue to add the test I described earlier: homogeneous medium, fully reflected, solution saturates to q/sigma_a. next, double the density and the solution should now saturate at half the previous value. simple enough but would test the change in density.

Sure, I can create an issue. I'd like to think a little more carefully about the test. The test you propose would catch whether or not we missed a spot in the code where we should be multiplying by density but are not. I think it would be equally as useful to have a multi-material test where instead of scaling cross sections we set material densities via whatever API I end up implementing for this.

_Originally posted by @zhardy-lanl in https://github.com/Open-Sn/opensn/issues/153#issuecomment-1973448540_

wdhawkins:
This PR contains the following changes:

1. Fixing density handling throughout the code (there were a number of places where density wasn't being used).
2. Add an optional `density` parameter to `Problem` that allows for setting a uniform problem density or setting density by block.
3. Added runtime density update support (`SetUniformDensities`, `SetDensities`)
4. Adding checks in acceleration methods to ensure that when acceleration is used, densities are uniform in each block (not strictly necessary right now, but a useful check).
5. Adding a number of regression tests.

Closes #161


ragusa (commented):
How do you feel about removing the density factor altogether?

PR #992, opened 2026-03-21
Providing fallback for `std::hardware_destructive_interference_size`

wdhawkins:
Some old standard libraries (clang 17/18) don't implement this.


andrsd (changes requested):
Remove the `test/python/modules/linear_boltzmann_solvers/transport_transient/.tests.json.swp` file

andrsd (approved):

PR #994, opened 2026-03-22
Print boundary names in error messages

### LINKED ISSUE #906: Improve error messages for boundary conditions in 1D
In 1D, it's not necessarily obvious that boundary conditions should be `zmin` and `zmax`. If the user supplies something else, say `xmin` and `xmax`, the code throws an error, but it is not helpful:
```
[1]  **** ERROR ****  IndexError: map::at:  key not found
[1]  **** ERROR ****
[1]  **** ERROR ****  At:
[1]  **** ERROR ****    TransientKEigenXSMapSwap.py(75): <module>
```

We need to improve our error messages for this case.

wdhawkins:
Closes #906


andrsd (changes requested):
You have the patch from #993 in this. This branch is probably based on a wrong HEAD.

Otherwise :+1:

ragusa (approved):

PR #997, opened 2026-03-24
Update strong and weak scaling

quocdang1998:
Refactor strong scaling and weak scaling script and extend the code for GPU scaling study.

To run strong scaling:
```
python3 generate_scaling_study --type=strong
pushd output/strong_cpu
source submit_jobs.sh
popd
python3 plot_strong.py --history=save
```

To run weak scaling:
```
python3 generate_scaling_study --type=weak
pushd output/weak_cpu
source submit_jobs.sh
popd
python3 plot_weak.py --history=save
```

Basic example:
strong_plot

weak_plot


andrsd (approved):

ragusa:
> > Don't delete the jupyter notebooks. This is our historical record of scaling performance. We need to be better about recording our scaling performance (both cpu and gpu). This might be something we should do quarterly... at least every time we release a new version. The typical process has been to conduct the scaling study and, when warranted, to update the jupyter notebooks to keep a record. I'm fine with the changes to the scripts... including plotting, but we still need to main a record in our notebooks.
> 
> @quocdang1998 I concur. We need to keep historical data somehow. For now, I guess it lives in the notebooks


any update?

wdhawkins:
Can we add the existing historical data to the yaml file? Is there a way to select the job scheduler template from the command line?

quocdang1998:
Can we add the existing historical data to the yaml file?
Yes, we can add historical data manually into the YAML file.

Is there a way to select the job scheduler template from the command line?
Do we run OpenSn on any job scheduler other than SLURM? If so, can you send me an example file? I can create a new option for user to choose between them.

wdhawkins:
> Can we add the existing historical data to the yaml file? Yes, we can add historical data manually into the YAML file.
> 
> Is there a way to select the job scheduler template from the command line? Do we run OpenSn on any job scheduler other than SLURM? If so, can you send me an example file? I can create a new option for user to choose between them.

Let's definitely add the existing data since it's published. On the scheduler front, we should add support for flux. It should also be easy to add other job schedulers in the future (no telling what we'll run into).

wdhawkins:
Just a couple of things. Do you want to commit `lib/__init__.py`? Can you adjust the history time for the Dane results to better match when they were actually generated. I don't remember when that was, but it should be in the original Jupyter notebooks.

wdhawkins (approved):

quocdang1998:
@wdhawkins 
`lib/__init__.py` is part of the source code. It cannot be removed.

The history time for the Dane results was updated. I couldn't find the exact time (hour/minute/second) of the run, so I kept the time stamp at 2AM.

wdhawkins (approved):

PR #999, opened 2026-03-24
Correctly find LAPACK and link against it

andrsd:
PETSc is built with BLAS/LAPACK support, but it treats it as a private library. That means the libraries used when building PETSc are not exposed to the PETSc client code. Suprisingly there is no issue on linux. However, on a Mac and in a conda environment there might be multiple installations of BLAS/LAPACK (Apple supplies one and conda can install additional ones).

Then, it can happen that OpenSn tries to link against the wrong version which may have different name mangling.  This patch adds support at the cmake level to find LAPACK (this shows now in the configure output). However, this by itself may not be enough to get a successful build. A cmake variable `BLA_VENDOR` can be set on the command line to hint cmake with BLAS/LAPACK to use. Typically, users should point to the BLAS/LAPACK used by PETSc if multiple libraries are installed.

### DIFF
--- a/.github/workflows/builds.yaml
+++ b/.github/workflows/builds.yaml
@@ -15,23 +15,36 @@ jobs:
         include:
           - compiler: gcc/13
             tag: gcc-13
+            dir: gcc/13.3.0
           - compiler: gcc/14
             tag: gcc-14
+            dir: gcc/14.2.0
           - compiler: clang/17
             tag: clang-17
+            dir: clang/17.0.6
           - compiler: clang/18
             tag: clang-18
+            dir: clang/18.1.8
           - compiler: clang/19
             tag: clang-19
+            dir: clang/19.1.5
           - compiler: intel/oneapi
             tag: intel-oneapi
+            dir: intel/2025.0.0
     steps:
       - uses: actions/checkout@v7
       - name: build
         shell: bash
         run: |
           module load python3/3.12.3 opensn/${{ matrix.compiler }}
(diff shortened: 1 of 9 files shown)


andrsd:
So, I have some updates:

I finally forced cmake to find/use BLAS/LAPACK. Had to force it to use fortran - OpenBLAS built with fortran. cmake tries to compile a little program to check to BLAS/LAPACK and fails to link if fortran is not available.

There are 2 ways to go about it
   1. Force the link libraries via `-DBLAS_LIBRARIES="/opt/local/OpenBLAS-0.3.31/lib64/libopenblas.a;pthread;m;gfortran"` - this is the current state of this PR. It basically forces the use of the BLAS library along with the libraries that must be linked with.
   2. Add `Fortran` in `project(... LANGUAGES ...)`. This works and cmake finds BLAS/LAPACK. However, it breaks in `find_package(MPI...`, since no mpi-fortran binary. We can restrict cmake to look for just C and CXX via `COMPONENTS` (the whole line is `find_package(MPI REQUIRED COMPONENTS C CXX)`), but this hits a bug in mpicpp-lite 2.5 which does `find_package(MPI REQUIRED)` and tries to pull C, CXX and Fortran (because this honors the enabled languages). I fixed that in 2.5.1, so I only pull the C component. Then, in the workflow file, we can do just `-DBLA_VENDOR=OpenBLAS -DBLAS_ROOT="/opt/local/OpenBLAS-0.3.31" -DLAPACK_ROOT="/opt/local/OpenBLAS-0.3.31"` - which IMO is cleaner than alternative 1
   3. ok, there is a third way: make fortran available in the CI environment :-)

PR #1003, opened 2026-03-29
Adds lazy flux-moment field-function allocation to LBSProblem

wdhawkins:
1. scalar-flux field functions are still created eagerly at initialization
2. higher-moment field functions are now allocated on demand when requested
3. newly allocated higher-moment field functions are initialized from the current phi_new state immediately
4. `SetPhiFromFieldFunctions()` now requires the requested field functions to already exist instead of silently allocating them


ragusa (approved):
I left a comment due to my curiosity but I can approve now

andrsd (approved):

PR #1017, opened 2026-04-06
Reallocate device memory for each WGS iteration

quocdang1998:
This PR enable device memory reallocation for each WGS iterations rather than always keeping the memory allocated on device.


andrsd (changes requested):

andrsd (approved):

wdhawkins:
My first question is: Are we really running into memory issues that necessitate this change? This change causes fluds allocation/deallocation to bleed into the solver, which I'd like to prevent. The only alternative would be to move it into the sweep wgs context which I think would be worse.  So... I'm curious what problem we're trying to solve here. If we haven't actually run into solid memory issues that we're trying to fix, I'm not sure we need this.

PR #1026, opened 2026-04-13
Change default for options.use_precursors to true.

wdhawkins:
## PR Checklist

### DIFF
--- a/doc/source/userguide/solvers.rst
+++ b/doc/source/userguide/solvers.rst
@@ -1,4 +1,3 @@
-=======
 Solvers
 =======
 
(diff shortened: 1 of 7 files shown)


andrsd (approved):

doc/source/userguide/solvers.rst
```diff
@@ -178,6 +178,11 @@ Problem options available in Python include:
 * ``field_function_prefix_option``
 * ``field_function_prefix``
 
+``use_precursors`` controls whether delayed neutron precursors are included in
+the transport problem. The default is ``True``. If the supplied cross sections
+do not contain precursor data, precursor treatment is disabled automatically
```
ragusa: with a print out on the console, right?

wdhawkins: Ah... good point. I'll make sure we do that.

wdhawkins: I've fixed this. If `options.use_precursors=True` and no XS in the active map has precursor data, we print a warning and continue in prompt-only mode.

ragusa (approved):

PR #1027, opened 2026-04-13
Add support for OpenMC MGXS delayed neutron data.

wdhawkins:
## PR Checklist


andrsd (approved):

doc/source/userguide/materials_xs.rst
```diff
@@ -309,6 +310,12 @@ Example:
    absorption dataset directly. This makes the imported data less sensitive to
    small inconsistencies or statistical noise.
 
+   When delayed-neutron data is present, OpenSn also imports precursor-group
+   information, prompt and delayed fission production, precursor decay
+   constants, and delayed emission spectra. If ``chi-prompt`` is not present,
+   OpenSn falls back to the steady fission spectrum ``chi`` as the prompt
+   spectrum.
```
ragusa: how do we ensure that we have a null transient when going from steady state to time-dependent? i think my comment has a lot of ramification that should be best discussed in person

wdhawkins: Yes, we should probably discuss this in person. But, if I understand your question correctly, using `initial_state=existing` with the transient solver should prevent any artificial transient jump from the mode transition itself. So, the workflow would be:

1.  Solve the steady-state problem to convergence.
2.  Switch to time-dependent mode.
3. Initialize `TransientSolver` with `initial_state=existing`.

ragusa: for a null transient test to pass, nu prompt + nu delayed must be = to nu tot. furthermore, chi is not easily deduced from chi prompt and chi delayed (there is an approximation that happens to yield chi used in steady-state, so consistency is also an issue). finally, when coming from MC, there is always statistical noise in the values. So, all in all, it is not obvious to how a null transient will happen. but we definitely need to add some consistency checks between steady-state data and transient data and report back to the user what those are, so that they are informed.

wdhawkins: I see. We should talk more about this.

Issue #925, opened 2026-02-06
`AAH_Angleset` improvements


wdhawkins:
Instead of storing `AngleAggregation::following_angle_sets_map_` and `SweepScheduler::preceding_angle_sets_`, you can consider adding a dependency counter and a following angleset `std::vector` in `AAH_AngleSet`.

aah_angle_set.h
```
class AAH_AngleSet : public AngleSet
{
...
protected:
std::size_t num_dependencies_ = 0;
std::size_t dependency_counter_ = 0;
std::vector<AAH_AngleSet*> following_angle_sets_;
}
```
Then, instead of performing a loop check, you'd simply have to do
```
num_dependencies_ = preceding_angle_sets_.size();

dependency_counter_ = num_dependencies_;  // before all angleset sweeps

// readiness check
bool deps_ready = (dependency_counter_ == 0);

// at the end of the sweep of each angleset
if (status == AngleSetStatus::FINISHED)
{
  std::for_each(following_angle_sets_.begin(), following_angle_sets_.end(), [](AAH_AngleSet* as) {as->dependency_counter_ -= 1;};
}
```

_Originally posted by @quocdang1998 in https://github.com/Open-Sn/opensn/pull/919#discussion_r2770739881_

PR #1030, opened 2026-04-22
Refactor SweepBoundary

quocdang1998:
In this PR, `SweepBoundary` is refactored:

- [x] `ArbitraryBoundary` and `ReflectingBoundary` accept different quadratures and angle aggregations per groupset.
- [x] Memory for all boundary are packed into one single `std::vector<double>` buffer (`BoundaryBank`) attached to `DiscreteOrdinatesProblem`. This enables one single contiguous copy of the boundary to GPU without bufferring.
- [x] `SweepBoundary` initialization steps are unified. Members are initialized in either the constructor or `InitializeAngleDependent`.
- [x] Fix a bug in `ReflectingBoundary` that write outgoing flux at group 0 instead of groupset first group.
- [x] Resolve issue [925](https://github.com/Open-Sn/opensn/issues/925).
- [x] For the GPU sweep, angle indices in an `AngleSet` are reordered so that their ordering is preserved between each `AngleSet` and its reflected `AngleSet`, regardless of the reflection map.
- [x] Add 2 regression tests for different quadratures per groupset.
- [x] Update new boundary accessing scheme for AAH GPU sweep. Boundary are copied from/to host's memory without bufferring.
- [x] Update time-dependent to boundary.

CPU sweep is not affected. Only improvement in the GPU sweep of low ranks are observed.

Strong scaling result (286k):
image

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
@@ -53,6 +53,68 @@ namespace opensn
 
 OpenSnRegisterObjectParametersOnlyInNamespace(lbs, DiscreteOrdinatesProblem);
 
+namespace
+{
+
+std::set<std::uint64_t>
+GetGlobalUniqueBoundaryIDs(const std::shared_ptr<MeshContinuum>& grid, mpi::Communicator& mpi_comm)
+{
+  std::set<std::uint64_t> local_unique_bids_set;
+  for (const auto& cell : grid->local_cells)
+    for (const auto& face : cell.faces)
+      if (not face.has_neighbor)
+        local_unique_bids_set.insert(face.neighbor_id);
+
+  const std::vector<std::uint64_t> local_unique_bids(local_unique_bids_set.begin(),
+                                                     local_unique_bids_set.end());
+
+  std::vector<std::uint64_t> recvbuf;
+  mpi_comm.all_gather(local_unique_bids, recvbuf);
+
+  std::set<std::uint64_t> global_unique_bids_set = local_unique_bids_set;
+  global_unique_bids_set.insert(recvbuf.begin(), recvbuf.end());
+
+  return global_unique_bids_set;
+}
+
+void
+RecursiveAngleSort(AngleSet* angleset,
+                   AngleAggregation& angle_agg,
+                   const std::vector<std::vector<std::uint32_t>>& reflected_maps,
+                   std::set<AngleSet*>& unsorted,
+                   std::set<AngleSet*>& sorted)
+{
+  sorted.insert(angleset);
+  std::uint32_t angle_zero = angleset->GetAngleIndices()[0];
+  for (const auto& reflected_map : reflected_maps)
+  {
+    auto reflected_angle_zero = reflected_map[angle_zero];
+    auto* reflected_angleset = angle_agg.GetAngleSetForAngleIndex(reflected_angle_zero);
+    if (sorted.contains(reflected_angleset))
+    {
+      bool is_coherent = std::equal(angleset->GetAngleIndices().begin(),
+                                    angleset->GetAngleIndices().end(),
+                                    reflected_angleset->GetAngleIndices().begin(),
+                                    [&](std::uint32_t angle, std::uint32_t reflected)
+                                    { return reflected_map[angle] == reflected; });
+      if (not is_coherent)
+        throw std::logic_error("Cannot find an unanimous sort order for the angle set.\n");
+    }
+    else
+    {
+      std::transform(angleset->GetAngleIndices().begin(),
+                     angleset->GetAngleIndices().end(),
+                     reflected_angleset->GetAngleIndices().begin(),
+                     [&](std::uint32_t angle) { return reflected_map[angle]; });
+      reflected_angleset->SyncDeviceAngleIndices();
+      unsorted.erase(reflected_angleset);
+      RecursiveAngleSort(reflected_angleset, angle_agg, reflected_maps, unsorted, sorted);
+    }
+  }
+}
+
+} // namespace
+
 InputParameters
 DiscreteOrdinatesProblem::GetInputParameters()
 {
@@ -95,11 +157,23 @@ DiscreteOrdinatesProblem::GetBoundaryOptionsBlock()
                                            {},
                                            "Required only if \"type\" is \"isotropic\". An array "
                                            "of isotropic strength per group");
+  params.AddOptionalParameter("start_time",
+                              -std::numeric_limits<double>::infinity(),
+                              "Time at which an isotropic boundary condition becomes active.");
+  params.AddOptionalParameter("end_time",
+                              std::numeric_limits<double>::infinity(),
+                              "Time at which an isotropic boundary condition becomes inactive.");
   params.AddOptionalParameter<std::shared_ptr<AngularFluxFunction>>(
     "function",
     std::shared_ptr<AngularFluxFunction>{},
     "Angular flux function to be used for arbitrary boundary conditions. The function takes an "
     "energy group index and a direction index and returns the incoming angular flux value.");
+  params.AddOptionalParameter<std::shared_ptr<AngularFluxTimeFunction>>(
+    "time_function",
+    std::shared_ptr<AngularFluxTimeFunction>{},
+    "Time-dependent angular flux function to be used for arbitrary boundary conditions. The "
+    "function takes an energy group index, a direction index, and time, and returns the incoming "
+    "angular flux value.");
   params.ConstrainParameterRange(
     "type", AllowableRangeList::New({"vacuum", "isotropic", "reflecting", "arbitrary"}));
 
@@ -206,13 +280,15 @@ DiscreteOrdinatesProblem::DiscreteOrdinatesProblem(const InputParameters& params
       grpset_psi_uk_man.AddUnknown(VarVecN, gs_num_groups);
 
     if (use_gpus_)
-      groupset.InitializeGPUCarriers();
+      groupset.InitializeQuadratureCarrier();
   }
 }
 
 DiscreteOrdinatesProblem::~DiscreteOrdinatesProblem()
 {
   CALI_CXX_MARK_FUNCTION;
+
+  ResetBoundaryCarrier();
 
   for (auto& groupset : groupsets_)
   {
@@ -293,16 +369,10 @@ DiscreteOrdinatesProblem::GetSweepBoundaries() const
   return sweep_boundaries_;
 }
 
-const std::map<uint64_t, DiscreteOrdinatesProblem::BoundaryDefinition>&
+const std::map<uint64_t, BoundaryDefinition>&
 DiscreteOrdinatesProblem::GetBoundaryDefinitions() const
 {
   return boundary_definitions_;
-}
-
-void
-DiscreteOrdinatesProblem::SetBoundaryOptions(const InputParameters& params)
-{
-  SetBoundaryOptionsImpl(params, true);
 }
 
 void
@@ -313,14 +383,17 @@ DiscreteOrdinatesProblem::SetBoundaryOptions(const std::vector<InputParameters>&
     boundary_definitions_.clear();
 
   for (const auto& params : boundary_params)
-    SetBoundaryOptionsImpl(params, false);
+    UpdateBoundaryDefinition(params);
 
-  RebuildBoundaryRuntimeData();
+  if (clear_existing or not boundary_params.empty())
+  {
+    InitializeBoundaries();
+    RebuildBoundaryRuntimeData();
+  }
 }
 
 void
-DiscreteOrdinatesProblem::SetBoundaryOptionsImpl(const InputParameters& params,
-                                                 bool rebuild_runtime_data)
+DiscreteOrdinatesProblem::UpdateBoundaryDefinition(const InputParameters& params)
 {
   const auto boundary_name = params.GetParamValue<std::string>("name");
   const auto coord_sys = grid_->GetCoordinateSystem();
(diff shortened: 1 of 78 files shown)


modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/isotropic_boundary.cc
```diff
@@ -0,0 +1,40 @@
...
+IsotropicBoundary::IsotropicBoundary(BoundaryBank& bank,
+                                     const std::vector<LBSGroupset>& groupsets,
+                                     std::vector<double>& boundary_flux)
+  : SweepBoundary(bank, LBSBoundaryType::ISOTROPIC)
+{
+  for (const auto& groupset : groupsets)
+  {
+    offset_[groupset.id] = bank_[groupset.id].counter++;
+    double* flux = bank_.ExtendBoundaryFlux(groupset.id, groupset.GetNumGroups());
+    for (unsigned int g = 0; g <= groupset.GetNumGroups(); ++g)
```
wdhawkins: Shouldn't this be `< groupset.GetNumGroups()` instead of <=?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
```diff
@@ -332,20 +396,10 @@ DiscreteOrdinatesProblem::SetBoundaryOptions(const InputParameters& params)
...
-      names << name;
-      first = false;
-    }
-    throw std::runtime_error(GetName() + ": Boundary name '" + boundary_name +
-                             "' not found in mesh. Available boundaries: [" + names.str() + "].");
+    throw std::runtime_error("Boundary name \"" + boundary_name + "\" not found in mesh.");
   }
   const auto bid = it->second;
-  boundary_definitions_[bid] = CreateBoundaryFromParams(params);
+  boundary_definitions_.emplace(bid, BoundaryDefinition(params, GetNumGroups()));
```
wdhawkins: Is `emplace` ok here? If this code is every called more than once, the second invocation will fail to correctly set the boundary definition.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.cc
```diff
@@ -0,0 +1,95 @@
...
+    map_dirnum.assign(num_angles, std::numeric_limits<std::uint64_t>::max());
+
+    auto& counter = bank_[groupset.id].counter;
+    offset_[groupset.id] = counter;
+    counter += num_angles;
+  }
+}
+
+void
+ArbitraryBoundary::InitializeAngleDependent(const std::vector<LBSGroupset>& groupsets)
```
wdhawkins, line 60: It looks to me like there's no longer a way to update the arbitrary boundary function after initialization. This won't work for time-dependent boundary conditions. We need to restore the ability to reset readiness every sweep and re-evaluate the boundary conditions.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/isotropic_boundary.cc
```diff
@@ -0,0 +1,40 @@
...
+    {
+      flux[g] = boundary_flux[groupset.first_group + g];
+    }
+  }
+  boundary_flux.clear();
```
wdhawkins: Why clear this? I don't think we should modify `boundary_definitions_` except when prompted by the user.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
```diff
@@ -599,6 +544,22 @@ DiscreteOrdinatesProblem::BuildRuntime()
     WGDSA::Init(*this, groupset);
     TGDSA::Init(*this, groupset);
   }
+
+  // Initialize angle-dependent boundary
+  for (auto& [bid, boundary] : sweep_boundaries_)
+    boundary->InitializeReflectingMap(groupsets_);
+  if (use_gpus_)
+    SortAngleSetsAngleIndices(); // DO NOT sort in RZ geometry
```
wdhawkins: Is this only for GPUs or for both GPUs and RZ?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/reflecting_boundary.cc
```diff
@@ -208,204 +294,181 @@ ReflectingBoundary::FinalizeDelayedAngularFluxSetup(
...
-    return true;
-
-  for (const auto& n : angles)
-    if (not boundary_flux_[reflected_anglenum_[n]].empty())
-      if (not angle_readyflags_[n])
-        return false;
-
-  return true;
+  FaceNode fn(cell_local_id, face_num, fi);
+  std::uint64_t facenode_idx = facenode_to_index_[fn];
```
wdhawkins: Should we check for a valid `fn` here?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_definition.cc
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void
+CheckRequiredParams(const InputParameters& params,
+                    const std::string& boundary_name,
+                    const std::string& boundary_type,
+                    const std::string& param_name)
+{
+  if (not params.Has(param_name))
+    throw std::runtime_error(std::format(R"(Boundary '{}' of type="{}" does not support "{}".)",
```
wdhawkins: If this is a required parameter check, should the error read "requires" instead of "does not support"?

wdhawkins (changes requested):

wdhawkins:
I've added comments about some of the larger issues that need to be addressed. I'll go back through again later. Right now, the code segfaults after execution with this branch (maybe because of the isotropic boundary issue I flagged above). We also need to restore the functionality for arbitrary boundaries to re-evaluate the boundary between sweeps. This is critical for time-dependent boundary conditions. This PR touches a lot of code and makes some very broad changes. I think it's going to take some time to fully review.

wdhawkins:
Once the segfault and arbitrary boundary issue have been fixed, it's probably worth running this through the address sanitizer on the CPU and compute sanitizer on the GPU.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_carrier.cu
```diff
@@ -0,0 +1,43 @@
...
+
+void
+BoundaryCarrier::UploadToDevice(int groupset_id)
+{
+  auto& flux = boundary_flux_[groupset_id];
+  crb::copy(flux.device_memory, flux.host_pin, flux.host_pin.size());
+}
+
+void
+BoundaryCarrier::DownloadToHost(int groupset_id)
```
andrsd, line 39: Do we want to use upload/download here? To me that is more related to network operations. Why not just `CopyToDevice`/`CopyToHost`?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/sweep_boundary.h
```diff
@@ -14,51 +16,51 @@ namespace opensn
...
-  explicit SweepBoundary(LBSBoundaryType bndry_type,
-                         unsigned int num_groups,
-                         CoordinateSystemType coord_type)
-    : num_groups_(num_groups), type_(bndry_type), coord_type_(coord_type)
+  explicit SweepBoundary(BoundaryBank& bank, LBSBoundaryType bndry_type)
+    : bank_(bank), type_(bndry_type)
   {
-    zero_boundary_flux_.resize(num_groups_, 0.0);
+    offset_.reserve(bank_.GetSize());
+    offset_.resize(bank_.GetSize());
```
andrsd, line 31: You don't need `reserve`, if you do `resize`

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
```diff
@@ -54,6 +54,68 @@ namespace opensn
...
+    auto* reflected_angleset = angle_agg.GetAngleSetForAngleIndex(reflected_angle_zero);
+    if (sorted.contains(reflected_angleset))
+    {
+      bool is_coherent = std::equal(angleset->GetAngleIndices().begin(),
+                                    angleset->GetAngleIndices().end(),
+                                    reflected_angleset->GetAngleIndices().begin(),
+                                    [&](std::uint32_t angle, std::uint32_t reflected)
+                                    { return reflected_map[angle] == reflected; });
+      if (not is_coherent)
+        throw std::logic_error("Cannot find an unanimous sort order for the angle set.\n");
```
andrsd, line 101: "Angle set indices are not coherent with the reflected mapping." would be probably a better error message.

modules/linear_boltzmann_solvers/lbs_problem/groupset/lbs_groupset.cc
```diff
@@ -142,6 +143,7 @@ LBSGroupset::LBSGroupset( // NOLINT(cppcoreguidelines-pro-type-member-init)
   last_group = groups_from_to[1];
   OpenSnInvalidArgumentIf(last_group < first_group,
                           "\"to\" field is less than the \"from\" field.");
+  size = last_group - first_group + 1;
```
andrsd, line 146: `size = GetNumGroups()`?

andrsd: nvm. i missed the lines below...

andrsd (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
quocdang1998: This part is only called once for every DiscreteOrdinatesProblem. Currently we don't support user to "rewrite" boundary, so this should be safe.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.cc
quocdang1998: Actually we can. We only fix the memory allocated for the arbitrary boundary. Value stored for the boundary can be updated.
If you want, I can separate the last loop that updates the values of the incoming fluxes out, so you can modify the values of the boundary in the future with this function.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/isotropic_boundary.cc
quocdang1998: In the old code, this is performed by an std::move, which delete the boundary_flux after this method is executed.

I can erase this part if you want.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
quocdang1998: We only sort for non RZ GPU sweep. For some reason, changing angle indices in an AngleSet actually changes the result of RZ sweep.
Cartesian sweep remains the same, regardless of sorting angle indicies or not.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/reflecting_boundary.cc
quocdang1998: Is it in the map. If fn is not valid, std::map will throw.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_carrier.cu
quocdang1998: I named it to be consistent with Carrier in LBSProblem. I think this should be addressed in a separate PR.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/sweep_boundary.h
quocdang1998: I did both because I don't want vector::capacity > vector::size (extra wasted memory), and I don't want to reallocate with shrink_to_fit.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_carrier.cu
andrsd: Ok

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.cc
wdhawkins: Yes, but all of the angle ready status infrastructure has been removed. How would time-dependent boundaries work in this version?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
wdhawkins: Yes, angle ordering is important in RZ. In this case, since the sort is guarded for gpus only, I think you can remove the comment (or move it outside of the conditional).

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
quocdang1998: The goal of `SortAngleIndices` is to find an order for all `AngleSet`s so that: under any reflection map, any pair of `AngleSet`s that are reflection of each other will have their angle indices sorted in the same order.

The error here indicates that: given all the reflection maps and the way angle indices are grouped, there is a conflict among these constraints that we cannot resolve. There is no unanimous order that satisfy all the requirements.

Your proposed error message just describes the state of the Angleet before sorting. It doesn't mean that we cannot find a way to sort that satisfies all the constraints we want.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/isotropic_boundary.cc
quocdang1998: Yes. Thanks for pointing this out.

It was indeed the **only cause** of the segfault.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.cc
quocdang1998: I've added a method named `UpdateBoundaryFlux`, which will update the boundary flux for all anglesets. Allocated memory is still retained, only its content is changed.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/isotropic_boundary.cc
quocdang1998: Fixed.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
quocdang1998: Fixed. The comment was removed.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_definition.cc
quocdang1998: Fixed.

quocdang1998:
@wdhawkins I ran this PR with sanitizer and `compute-sanitizer --tool memcheck`. After fixing the bug in `IsotropicBoundary`, no segfault error occurs.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
andrsd: Ok

andrsd (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/arbitrary_boundary.cc
quocdang1998: I updated the boundary with time-dependent feature. Please review it.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
wdhawkins: Isn't this also called via the Python API using `SetBoundaryOptions`?  If so, that will fail if we continue to use `emplace` here.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_carrier.cu
```diff
@@ -0,0 +1,47 @@
...
+    auto& host_data = bank[groupset.id].boundary_flux;
+    device_boundary_flux_[groupset.id] = crb::DeviceMemory<double>(host_data.size());
+  }
+}
+
+void
+BoundaryCarrier::UploadToDevice(int groupset_id)
+{
+  auto& host_data = bank_[groupset_id].boundary_flux;
+  crb::MemoryPinningManager<double> host_pinner(host_data);
```
wdhawkins, line 33: Can we refactor this so we don't pin/unpin memory every sweep? That's going to be extremely expensive.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_definition.cc
```diff
@@ -0,0 +1,145 @@
...
+        if (not param_angular_flux_function)
+          throw std::runtime_error(std::format(
+            "Boundary '{}' with type=\"arbitrary\" expected parameter \"time_function\" to hold a "
+            "non-null AngularFluxFunction.",
+            boundary_name));
+        time_angular_flux_function = std::move(param_angular_flux_function);
+      }
+      break;
+    }
+    default:
```
wdhawkins, line 137: We should also reject time_function, start_time, and end_time for vacuum/reflecting boundaries.

wdhawkins (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_carrier.cu
quocdang1998: I have a plan to do that. But I think it should deserve its own PR after this PR is merged.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
quocdang1998: Fixed. I used `boundary_definitions_[bid]` like the old version.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/boundary/boundary_definition.cc
quocdang1998: Fixed.

wdhawkins (approved):

PR #1034, opened 2026-04-26
Improving warnings when computing diffusion parameters.

### LINKED ISSUE #753: clarification of warnings in ComputeDiffusionParameters
Often, one can see this:
`*** WARNING ***  Negative transport cross section found for group 0 in call to ComputeDiffusionParameters. sigma_t=0 sigma_1=0. Setting sigma_1=0, sigma_tr=sigma_t for this group.`

I propose the following:

1. first, the above printout should be moved inside the loop `if (S.size() > 1)` because there's no reason for that warning to show when only isotropic scattering XS are used
2. the code computes `sigma_tr_[g] = sigma_t_[g] - sigma_1` but then keeps on using `sigma_t_[g] - sigma_1` afterwards, why? not clean.
3. add another warning just before computing D when `|sigma_tr_[g]|<eps` starting what value is being used instead of 0

wdhawkins:
Closes #753

### DIFF
--- a/framework/materials/multi_group_xs/multi_group_xs.cc
+++ b/framework/materials/multi_group_xs/multi_group_xs.cc
@@ -4,6 +4,7 @@
 #include "framework/materials/multi_group_xs/multi_group_xs.h"
 #include "framework/runtime.h"
 #include "framework/logging/log.h"
+#include <cmath>
 
 namespace opensn
 {
@@ -383,6 +384,9 @@ MultiGroupXS::ComputeDiffusionParameters()
   if (diffusion_initialized_)
     return;
 
+  constexpr double sigma_tr_min = 1.0e-12;
+  constexpr double diffusion_coeff_max = 1.0 / 3.0 / sigma_tr_min;
+
   // Initialize diffusion data
   sigma_tr_.resize(num_groups_, 0.0);
   diffusion_coeff_.resize(num_groups_, 1.0);
@@ -395,7 +399,8 @@ MultiGroupXS::ComputeDiffusionParameters()
   {
     // Determine transport correction
     double sigma_1 = 0.0;
-    if (S.size() > 1)
+    const bool has_p1_scattering = S.size() > 1;
+    if (has_p1_scattering)
     {
       for (unsigned int gp = 0; gp < num_groups_; ++gp)
       {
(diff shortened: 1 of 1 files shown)


framework/materials/multi_group_xs/multi_group_xs.cc
```diff
@@ -389,22 +394,27 @@ MultiGroupXS::ComputeDiffusionParameters()
...
     sigma_tr_[g] = sigma_t_[g] - sigma_1;
 
     // Compute the diffusion coefficient
-    // Cap the value for when sigma_t - sigma_1 is near zero
-    if ((sigma_t_[g] - sigma_1) < 1.0e-12)
-      diffusion_coeff_[g] = 1.0e12;
+    // Cap the value for when the transport cross section is near zero
+    if (std::fabs(sigma_tr_[g]) < sigma_tr_min)
+    {
+      log.Log0Warning() << "Transport cross section near zero for group " << g
```
ragusa, line 432: "near zero" --> do we want to give the value that is actually used in `sigma_tr_min` in the message ?

framework/materials/multi_group_xs/multi_group_xs.cc
```diff
@@ -361,6 +362,9 @@ MultiGroupXS::ComputeDiffusionParameters()
   if (diffusion_initialized_)
     return;
 
+  constexpr double sigma_tr_min = 1.0e-12;
+  constexpr double diffusion_coeff_max = 1.0e12;
```
ragusa: I would recommend that `sigma_tr_min ` and `diffusion_coeff_max ` are consistent with one another., so make `diffusion_coeff_max = 1.0/3.0/sigma_tr_min `

ragusa (commented):
see comments

framework/materials/multi_group_xs/multi_group_xs.cc
wdhawkins: Fixed.

framework/materials/multi_group_xs/multi_group_xs.cc
wdhawkins: Fixed.

framework/materials/multi_group_xs/multi_group_xs.cc
```diff
@@ -361,6 +362,9 @@ MultiGroupXS::ComputeDiffusionParameters()
   if (diffusion_initialized_)
     return;
 
+  constexpr double sigma_tr_min = 1.0e-12;
```
andrsd, line 387: `sigma_tr_min` -> `SIGMA_TR_MIN` (coding standard)

framework/materials/multi_group_xs/multi_group_xs.cc
```diff
@@ -361,6 +362,9 @@ MultiGroupXS::ComputeDiffusionParameters()
...
+  constexpr double diffusion_coeff_max = 1.0 / 3.0 / sigma_tr_min;
```
andrsd, line 388: `diffusion_coeff_max` -> `DIFFUSION_COEFF_MAX` (coding standard)

andrsd (changes requested):

andrsd:
Maybe, we should be doing UPPER_CASE only when the constants are in the outside scope? i.e., not inside the function?

wdhawkins:
> Maybe, we should be doing UPPER_CASE only when the constants are in the outside scope? i.e., not inside the function?

I think I like that best. I can see using upper case for file scope constants.

ragusa (commented):
good to go, after David's comments are addressed.

framework/materials/multi_group_xs/multi_group_xs.cc
wdhawkins: Kept all lower case in favor of upper case for outside-scope constants.

andrsd (approved):

PR #1037, opened 2026-04-26
Add Diffusion MIP MMS convergence test and fix RHS reassembly

### LINKED ISSUE #311: what happened to this file SimTest_IP_MMS_L2error ?
in `[test/modules/dfem_diffusion/d_diffusion_2d_3c_l2_error_mms.lua](https://github.com/Open-Sn/opensn/blob/054d07bc2a9f6a4e718d4fb7fc03930a621ab8ef/test/modules/dfem_diffusion/d_diffusion_2d_3c_l2_error_mms.lua#L57)`

wdhawkins:
This PR adds a diffusion MIP MMS unit test based on the old `sim_IP_MMS_L2error` code. The test performs a 2D convergence study on square meshes with N = 5, 10, 20, 40, 80 and verifies L2 error reduction.

In addition, while converting the old MMS workflow, the test exposed a bug in `DiffusionMIPSolver::Assemble_b_wQpoints`. For MMS Dirichlet boundary terms, the full assembly path used `ref_solution_function_`, but the RHS-only reassembly path used `source_function_`. This meant RHS-only reassembly was applying the manufactured source on the boundary instead of the manufactured reference solution.  The RHS-only assembly path now uses `ref_solution_function_`, making it consistent with full matrix/RHS assembly.

Closes #311.

### DIFF
--- a/test/unit/diffusion/diffusion_mip_mms_l2_error.cc
+++ b/test/unit/diffusion/diffusion_mip_mms_l2_error.cc
@@ -1,0 +1,181 @@
+// SPDX-FileCopyrightText: 2026 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#include "gmock/gmock.h"
+#include "test/unit/common/mesh_builders.h"
+
+#include "modules/diffusion/diffusion_mip_solver.h"
+#include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/acceleration/acceleration.h"
+#include "modules/linear_boltzmann_solvers/lbs_problem/lbs_structs.h"
+
+#include "framework/field_functions/field_function_grid_based.h"
+#include "framework/logging/log.h"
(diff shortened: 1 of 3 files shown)


quocdang1998 (approved):

test/unit/diffusion/diffusion_mip_mms_l2_error.cc
```diff
@@ -0,0 +1,181 @@
+// SPDX-FileCopyrightText: 2024 The OpenSn Authors <https://open-sn.github.io/opensn/>
```
andrsd: 2024 -> 2026

andrsd (changes requested):

test/unit/diffusion/diffusion_mip_mms_l2_error.cc
wdhawkins: Fixed.

andrsd (approved):

PR #1047, opened 2026-05-01
Moving cross-section files in `tests/` in `assets/xs/`

andrsd:
## PR Checklist


wdhawkins:
There are also cross sections in `transport_keigen/c5g7/materials`. We'll probably want to move them to something like `assets/xs/c5g7`. And maybe meshes as well.

quocdang1998 (approved):

wdhawkins (approved):

ragusa (commented):
Why does this show up as 289 new lines and about 89,000 deleted lines?

andrsd:
> Why does this show up as 289 new lines and about 89,000 deleted lines?

I noticed that too. I think it is because most files were moved (no line change). However, there were some duplicate files like `xs_168g.xs` (~60k lines), and `xs_graphite_pure.xs` (~30k lines). So, one instance was moved and one was deleted, hence the large number of removed lines.

ragusa (approved):

PR #1048, opened 2026-05-01
Adding surface postprocessor

andrsd:
## PR Checklist

### DIFF
--- a/test/python/modules/linear_boltzmann_solvers/pps/transport_2d_1_poly.py
+++ b/test/python/modules/linear_boltzmann_solvers/pps/transport_2d_1_poly.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+2D PWLD transport test with vacuum and Incident-isotropic bundary conditions
+Test: Max-value=0.50758 and 2.52527e-04
+"""
+
+import math
+import os
+import sys
+
+if "opensn_console" not in globals():
+    from mpi4py import MPI
+
+    size = MPI.COMM_WORLD.size
+    rank = MPI.COMM_WORLD.rank
+    sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../../../")))
+    from pyopensn.aquad import GLCProductQuadrature2DXY
+    from pyopensn.logvol import RPPLogicalVolume
+    from pyopensn.mesh import FromFileMeshGenerator, KBAGraphPartitioner
+    from pyopensn.post import SurfacePostprocessor
+    from pyopensn.solver import DiscreteOrdinatesProblem, SteadyStateSourceSolver
+    from pyopensn.source import VolumetricSource
+    from pyopensn.xs import MultiGroupXS
+
+if __name__ == "__main__":
+    # Setup mesh
+    meshgen = FromFileMeshGenerator(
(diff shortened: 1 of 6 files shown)


andrsd:
This is on top of PR 1047.
I will add doco sometime today. 
This is to show the code and get feedback.
Testing is not as extensive as for the volumetric PPS, so any ideas on improving that are welcome.

test/python/modules/linear_boltzmann_solvers/pps/transport_2d_1_poly.py
```diff
@@ -0,0 +1,213 @@
...
+    ss_solver = SteadyStateSourceSolver(problem=phys)
+    ss_solver.Initialize()
+    ss_solver.Execute()
+
+    # PPS over xmin boundary
+    pps_xmin_min = SurfacePostprocessor(
+        problem=phys,
+        value_type="min",
+        current_type="net",
+        boundaries=["xmin"]
```
ragusa, line 106: can we add an example with boundary IDs as well?

andrsd: I am assuming you are asking to add a tutorial. What problem should I use to demonstrate this on? These tests are kinda contrived. I can reuse the test, but if you have something better in mind, I would use that instead, I think.

ragusa: is this still pending because you are waiting on an examaple from me?

andrsd: Yes. If you have anything in mind that would help users let me know, so I can work on it.

ragusa: @rodriguez-jay Jason, please create several simple 2d, 2-material zone, fixed source problems. All of them should use a mesh from gmsh. Let a subspace of subset of the outer boundary be where the surface postprocessor should act.
- 2 geometries: two embedded squares, two embedded circles.
- surface of interest for the surface postprocessor: a fraction of an edge in the square case, an arc of circle in the circular case
- 2 ways of marking the surface of interest: once in gmsh itself, once using logvol (rectangles for the square case, cylinders for the circular case)

So 4 cases in total: 2 geometries times 2 ways of marking. Upload the data here or email it to David and me. Thanks

andrsd: This sounds like you want to set up the post-processors on an internal boundary. This is not possible with the current code that is in `Mesh`/`MeshContinuum`. The `CellFace` class stores `neighbor_id`. If the neighbor is another cell, then it is a cell id, otherwise it is a boundary id. So, we would need to change this. And then either allow boundaries to be assigned on internal sides - which may be questionable, since I could put a real BC on an internal side (may be be accident). Or introduce some new `InternalBoundary` class to handle PPS over internal sides. Or may be something completely different.

Right now, this PR only allows the surface postprocessor to run over external sides.

ragusa: I said outer boundary.

andrsd: Ok. I misunderstood the embedding part. I thought you were suggesting to embed the squares and circles into a larger domain... My bad.

PR #1058, opened 2026-05-07
Remove host memory registration

quocdang1998:
This PR removes the use of `cudaHostRegister`/`cudaHostUnregister`.

Memory registration was initially implemented for asynchronous transfer on stream. However, with many improvements of single-time copy, memory registration is not required anymore. They are removed to save some sweep time.

Note: no improvement/degradation of performance was observed.

image

## PR Checklist


modules/linear_boltzmann_solvers/lbs_problem/device/storage.h
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 #pragma once
-
+/*
```
andrsd: Do you plan on reusing this in future. Or did you just forgot to remove this?

quocdang1998: I've forgotten to remove this. Thanks for pointing this out.

andrsd (approved):

wdhawkins (approved):

PR #1065, opened 2026-05-13
Removing trailing _ from local variables

andrsd:
## PR Checklist


wdhawkins (approved):

ragusa (approved):
did we discuss, a long time ago, adding _ to local vars to distinguish them from member vars? is this still true? should we update the developer user guide? just wondering ...

andrsd:
AFAIK, `_` should be added only to protected or private member variables. All other variables or functions arguments should not use it. At least, that's what is in the coding standard...

PR #1070, opened 2026-05-14
Adding distributed orthogonal mesh creation.

### LINKED ISSUE #19: SplitMesh
With orthogonal grids, the grid creation and KBA partition should not require rank-0 to do all the work.

wdhawkins:
This PR adds distributed orthogonal mesh creation. This implementation differs from true distributed mesh creation in that rank 0 creates the entire global cell graph. This was intentional, as it allows the existing partitioning algorithms to work as is. After partitioning, rank 0 distributes partition and cell data to each rank, and the individual ranks build their own cells and populate the mesh. 

Once this is tested at scale, I'll remove the "Draft" designation.

Closes #19 

## PR Checklist

### DIFF
--- a/framework/mesh/mesh_generator/orthogonal_mesh_generator.cc
+++ b/framework/mesh/mesh_generator/orthogonal_mesh_generator.cc
@@ -2,16 +2,469 @@
 // SPDX-License-Identifier: MIT
 
 #include "framework/mesh/mesh_generator/orthogonal_mesh_generator.h"
+#include "framework/graphs/graph_partitioner.h"
+#include "framework/mesh/mesh_continuum/mesh_continuum.h"
+#include "framework/mpi/mpi_utils.h"
 #include "framework/object_factory.h"
+#include "framework/runtime.h"
 #include "framework/logging/log.h"
+#include <algorithm>
+#include <array>
+#include <map>
+#include <unordered_map>
 
 namespace opensn
 {
+namespace
+{
+
+struct OrthoCellInfo
+{
+  unsigned int dimension = 0;
+  std::array<size_t, 3> cell_counts = {1, 1, 1};
+  std::array<size_t, 3> node_counts = {2, 2, 2};
+};
+
+OrthoCellInfo
+MakeOrthoCellInfo(const std::vector<std::vector<double>>& node_sets)
+{
+  OrthoCellInfo info;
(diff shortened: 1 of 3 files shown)


ragusa:
should there be an example/tutorial and/or a regression test added?

wdhawkins:
> should there be an example/tutorial and/or a regression test added?

Yes. I'll add regression tests and a tutorial.

python/lib/mesh.cc
```diff
@@ -402,6 +402,9 @@ WrapMeshGenerator(py::module& mesh)
         Sets of nodes per dimension. Node values must be monotonically increasing.
     coord_sys: {'cartesian', 'cylindrical', 'spherical'}
         The coordinate system of the mesh.
+    distributed_generation: bool, default=False
+        When true, partition the orthogonal cell graph on rank 0 and build each
+        rank's local and ghost cells directly on that rank.
```
ragusa, line 407: this could be confusing. So the cell graph always exists on rank 0, but when this flag is true, the mesh generation takes place on individual ranks. But a user may not appreciate the difference between a mesh continuum and just a cell graph. This is just a comment.

ragusa (approved):
read the comments and decide. you can just resolve the conversation if appropriate.

andrsd (commented):
This seems ok.
It is marked as draft, so I can check again when ready for review.

python/lib/mesh.cc
wdhawkins: I'll update this for clarity.

framework/mesh/mesh_generator/orthogonal_mesh_generator.cc
```diff
@@ -2,16 +2,467 @@
...
+{
+  partI_num_cells.assign(num_partitions, 0);
+  std::map<int, std::vector<uint64_t>> send_payloads;
+
+  for (uint64_t cell_gid = 0; cell_gid < cell_pids.size(); ++cell_gid)
+  {
+    const int cell_pid = cell_pids[cell_gid];
+    ++partI_num_cells[cell_pid];
+
+    std::array<int, 27> recipient_pids = {};
```
andrsd, line 437: Maybe a note on the `27`?

wdhawkins: Fixed!

andrsd (approved):

PR #1073, opened 2026-05-14
Minor bug fix for classic Richardson delayed-angular-flux convergence check.

wdhawkins:
This PR updates the delayed angular flux convergence check to use the maximum pointwise change across all ranks.

### DIFF
--- a/framework/math/math.cc
+++ b/framework/math/math.cc
@@ -570,17 +570,17 @@ ComputeL2Change(std::vector<double>& x, std::vector<double>& y)
 }
 
 double
-ComputePointwiseChange(std::vector<double>& x, std::vector<double>& y)
+ComputePointwiseChange(const std::vector<double>& x, const std::vector<double>& y)
 {
   assert(x.size() == y.size());
 
   double pw_change = 0.0;
   for (auto i = 0; i < x.size(); ++i)
   {
-    double max = std::max(x[i], y[i]);
-    double delta = std::fabs(x[i] - y[i]);
-    if (max >= std::numeric_limits<double>::min())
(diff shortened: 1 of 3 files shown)


framework/math/math.cc
```diff
@@ -570,17 +570,17 @@ ComputeL2Change(std::vector<double>& x, std::vector<double>& y)
...
 
   double pw_change = 0.0;
   for (auto i = 0; i < x.size(); ++i)
   {
-    double max = std::max(x[i], y[i]);
-    double delta = std::fabs(x[i] - y[i]);
-    if (max >= std::numeric_limits<double>::min())
-      pw_change = std::max(delta / max, pw_change);
+    const double max_value = std::max(std::fabs(x[i]), std::fabs(y[i]));
+    const double delta = std::fabs(x[i] - y[i]);
```
ragusa, line 581: why is it better to have const here? seem overkill

ragusa (approved):

quocdang1998 (approved):

framework/math/math.cc
wdhawkins: No particular reason. It would be fine without `const`. Just good practice.

andrsd (approved):

framework/math/math.cc
andrsd: The argument I usually see is that by using `const` you hint the compiler about mutation and it can produce better code...

andrsd: Another aspect is that you communicate your intent more clearly. If you see `const` you know this value won't be changed anywhere within it scope... (unless somebody does `const_cast`, but then you see that quite well...)

PR #1080, opened 2026-05-22
Bump mpicpp-lite to 2.7.1

andrsd:
## PR Checklist


andrsd:
This will fail, since it needs an update on the CI machine.
This is to support PR #1060

andrsd:
Bumped to 2.7.1. 

Previous version caused a compile error with gcc-15. This should fix the problem.

wdhawkins (approved):

PR #1086, opened 2026-05-28
CBC now requires at least Boost v1.86

wdhawkins:
CBC needs `boost::unordered_flat_map`, making 1.81 the minimum version.


andrsd (changes requested):
I suggest you bump this all the way to 1.86, because:
a) doco install says 1.86+
b) build dependencies script grabs and builds 1.86
c) spack also uses 1.86

wdhawkins:
> I suggest you bump this all the way to 1.86, because: a) doco install says 1.86+ b) build dependencies script grabs and builds 1.86 c) spack also uses 1.86

Yes, good points. Fixed.

andrsd (approved):

PR #1094, opened 2026-06-01
Adding config for dependabot

andrsd:
## PR Checklist


andrsd:
Every month `dependabot` will check and send us PRs to update outdated actions we are using.

wdhawkins (approved):

PR #1095, opened 2026-06-03
Moving sweep runtime construction out of DiscreteOrdinatesProblem

wdhawkins:


andrsd (approved):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/spds/aah.h
```diff
@@ -51,19 +52,19 @@ class AAH_SPDS : public SPDS
...
   /// Returns the global sweep FAS as a vector of edges.
-  std::vector<int> GetGlobalSweepFAS() { return global_sweep_fas_; }
+  const std::vector<int>& GetGlobalSweepFAS() const { return global_sweep_fas_; }
 
   /**
    * Sets the global sweep FAS.
    * \param edges A vector of edges representing the FAS.
    */
-  void SetGlobalSweepFAS(std::vector<int>& edges) { global_sweep_fas_ = edges; }
+  void SetGlobalSweepFAS(std::vector<int> edges) { global_sweep_fas_ = std::move(edges); }
```
quocdang1998, line 61: Why don't you use `const ... &` and the copy assignment?

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
```diff
@@ -1206,234 +1197,19 @@ DiscreteOrdinatesProblem::InitializeSweepDataStructures()
...
-      for (const auto& spds : spds_list)
-      {
-        quadrature_fluds_commondata_map_[quadrature].push_back(
-          std::make_unique<CBC_FLUDSCommonData>(*spds, grid_nodal_mappings_));
-      }
-    }
+    if (sweep_type_ == "AAH")
+      CreateAAHD_FLUDSCommonData();
+    else if (sweep_type_ == "CBC")
+      CreateCBCD_FLUDSCommonData();
```
quocdang1998: You can consider move `CreateAAHD_FLUDSCommonData` and `CreateCBCD_FLUDSCommonData` into sweep_runtime_builder too.

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/spds/aah.h
wdhawkins: That would work, but the current version handles lvalue/rvalue inputs in one overload. If the input is an lvalue, we copy into edges and then move to the member. If the input is an rvalue/temporary, we move into edges and them move into the member. If we did `const ... &`, the first op would always be a copy, even if the input was movable.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/discrete_ordinates_problem.cc
wdhawkins: Good catch. Fixed.

quocdang1998 (approved):

PR #1120, opened 2026-07-27
Linear interpolator for cross section

quocdang1998:
## PR Checklist

This PR implements `CartesianGrid` and `LinearInterpolator` for `MultiGroupXS`. `Interpolator` copied scattered XS data into one contiguous buffer and perform interpolation operation on it.

Usage:
```
grid = CartesianGrid([[0.0, 1.5], [-1.0, 2.0], [10.0, 13.0]])

xs_data = np.empty((2, 2, 2), dtype=object)
    for i, x in enumerate(grid_data[0]):
        for j, y in enumerate(grid_data[1]):
            for k, z in enumerate(grid_data[2]):
                xs_data[i, j, k] = make_simple_one_group_xs(x, y, z)

interpolator = LinearInterpolator(
    grid,
    xs_data,
    XSType.Total | XSType.Absorption | XSType.Transfer,
)
```

### DIFF
--- a/framework/materials/interpolator/interpolator.h
+++ b/framework/materials/interpolator/interpolator.h
@@ -1,0 +1,151 @@
+// SPDX-FileCopyrightText: 2026 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "framework/data_types/sparse_matrix/sparse_matrix.h"
+#include "framework/materials/interpolator/rt_ndarray.h"
+#include "framework/materials/interpolator/cartesian_grid.h"
+#include "framework/materials/multi_group_xs/multi_group_xs.h"
+#include <cstdint>
+#include <limits>
+#include <memory>
+#include <set>
+#include <utility>
+#include <vector>
+
+namespace opensn
+{
+
+enum XSType : std::uint64_t
+{
+  Total = 1 << 0,
+  Absorption = 1 << 1,
+  Fission = 1 << 2,
+  NuFission = 1 << 3,
+  Chi = 1 << 4,
+  ProductionMatrix = 1 << 5,
+  NuPromptFission = 1 << 6,
+  NuDelayedFission = 1 << 7,
+  Precursor = 1 << 8,
+  Transfer = 1 << 9,
+  Default = std::numeric_limits<std::uint64_t>::max(),
+};
+
+/**
+ * Sparse matrix metadata.
+ * Instances can be merged to form the union of the sparsity patterns found at all
+ * interpolation-grid points.
+ */
+struct SparseMatrixProperties
+{
+  /// Number of matrix rows.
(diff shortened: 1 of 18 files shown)


framework/materials/interpolator/interpolator.h
```diff
@@ -0,0 +1,151 @@
...
+#include <set>
+#include <utility>
+#include <vector>
+
+namespace opensn
+{
+
+enum XSType : std::uint64_t
+{
+  Total = 1 << 0,
```
andrsd, line 22: enum values are UPPER_CASE, see [coding standard](https://open-sn.github.io/opensn/devguide/coding_standard.html#enums)

andrsd (changes requested):

framework/materials/interpolator/linear_interpolator.h
wdhawkins, line 1: A major issue I see so far is that the interpolator can produce a `MuliGroupXS` object that is only partially populated (this is by design based on `XSType`), but pretty much every other part of the code expects that `MultGroupXS` is fully populated and consistent once it's built.

Examples:
1. If we interpolate without `Total`, `ComputeDiffusionParameters` will read out of bounds on `sigma_t`.
2. If we interpolate without `PrductionMatrix`, the solvers will fail calling `GetProductionMatrix` (only guarded by `IsFissionable`).
3. Interpolating adjoint-mode source XS will give the wrong diffusion coefficients since `ComputeDiffusoinParameters` needs the untransposed members.
4. `Scale` writes `sigma_f/product_matrix_` when a material is fissionable, but this will fail if we interpolate without fission.
5. `num_precursors` is set unconditionally from the source XS, but `GetPrecursors` is empty unless we set `XSType::Precursor`. So `GetNumPrecursors` and `GetPrecursors().size()` disagree.

I think we need to re-evaluate the design here. `MultiGroupXS` was never designed to be built incrementally from outside the class. It's always built from its own load functions that populate every field in one pass, so the internal data is always complete and consistent.  To support the interpolator, `MultiGroupXS` added a host of non-const getters that return mutable references to its internal arrays, effectively making them setters. `Evaluate` then uses these to fill in only the fields the caller asked for, but nothing enforces that the fields filled in are the ones other parts of the class actually depend on.

The class was never intended to be populated in this way. `Scale` is really the only accessible method that can mutate the internal data, but it's careful to explicitly reset the dependent state (`diffusion_initialized_ = false`) so it gets recomputed correctly on next use. `Evaluate` doesn't follow that same pattern. It writes fields directly through the exposed getters without tracking which derived state depends on which fields, so the resulting XS object can end up in an inconsistent state that doesn't make sense.

I think the interpolater should be written as a load function for the XS object and fully and consistently populate the XS object instead of trying to do things piecemeal after the object has been constructed (this is going to be the only way to guarantee internal data consistency).

wdhawkins (changes requested):

PR #1121, opened 2026-07-30
Factor SIMD

quocdang1998:
This PR extracts SIMD instructions into `framework/simd` and exposes them through a unified `Simd` and `SimdIndex` API.

This allows to remove architecture-specific macro checks in the AAH/CBC AVX sweep-chunk implementations. All macro checks are confined only to `framework/simd`.

In addition, this PR adds Neon64 support for ARM CPUs, including macOS on Apple Silicon, as well as SSE-only implementations.

### DIFF
--- a/framework/simd/simd.h
+++ b/framework/simd/simd.h
@@ -1,0 +1,20 @@
+// SPDX-FileCopyrightText: 2026 The OpenSn Authors <https://open-sn.github.io/opensn/>
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <cstdint>
+
+#ifdef __AVX512F__
+#include "framework/simd/simd_avx512.h"
+#elif defined(__AVX__) || defined(__AVX2__)
+#include "framework/simd/simd_avx.h"
+#elif defined(__SSE2__) || defined(_M_X64) || defined(_M_AMD64) ||                                 \
+  (defined(_M_IX86_FP) && (_M_IX86_FP >= 2))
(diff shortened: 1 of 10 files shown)


andrsd (approved):

wdhawkins:
Since we don't yet have a performance test suite, have you run 1-2 largish problem on class02 to verify performance?

quocdang1998:
@wdhawkins `simd` and `main` shows no differences.

In term of bytecode, I can confirm that both branches produce the same binary.
Result on Tuolumne (may vary from machine to machine):
```
$ nm -S -C --size-sort libopensn.so.1.0.1 | rg 'AAH_Sweep_FixedN'
000000000059abc0 0000000000005848 W void opensn::AAH_Sweep_FixedN<4u, false>(opensn::AAHSweepData&, opensn::AngleSet&)
$ dd if=libopensn.so.1.0.1 bs=1 skip=$((0x59abc0)) count=$((0x5848)) status=none | sha256sum
8c82b4851611856389f0c53f19c1c86d73b427f317cdc71c2e886b64dcdc8d0d  -
```
The hashes are the same.

In term of performance, I ran a 12k tets problem (based on scaling study) and a 20x20x20 orthogonal mesh problem. For both cases, runtime results vary within 3%, depending on which branch is launched first. Somehow, the one launched after has a sweep time higher by a few percent than the first, regardless of which branch is launched first.

wdhawkins (approved):

PR #1125, opened 2026-08-12
Add non-copy path for GPU-aware MPI and integrated memory system

quocdang1998:
This PR adds a no host-device copy route for non-local flux and delayed flux.

On Tuo:
```
cmake -DOPENSN_USE_GPU_AWARE_MPI=ON ...
```
To run:
```
export MPICH_GPU_SUPPORT_ENABLED=1
```

Compared with `main`:

image

[hip-aware-mpi.pdf](https://github.com/user-attachments/files/31111017/hip-aware-mpi.pdf)

## PR Checklist


modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cu
```diff
@@ -8,9 +8,39 @@
...
+namespace crb = caribou;
+
 namespace opensn
 {
 
+namespace
+{
+
+bool
+CheckIntegratedMemory()
```
wdhawkins, line 21: I'm surprised clang-tidy didn't flag this. This method needs a default return or needs to throw an error. Also, we don't check the return codes of any of the `GetDevice` calls. So, if they fail, `GetAttribute` potentially gets garbage. We should probably handle return codes appropriately.

wdhawkins (changes requested):

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cu
quocdang1998: This is because clang-tidy are not executed on .cu files.

In order to do this, clang must be configure with NVPTX support alongside X86-64.

andrsd (approved):

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.cu
quocdang1998: Fixed both `#else` branch and add `GetDevice*` error check.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -217,7 +314,7 @@ AAHD_FLUDS::GetDevicePointerSet()
     assert(pointer_set.local_psi != nullptr);
   // delayed local psi
   pointer_set.delayed_local_psi = delayed_local_psi_bank_.device_storage.get();
-  pointer_set.delayed_local_psi_old = delayed_local_psi_old_bank_.device_storage.get();
+  pointer_set.delayed_local_psi_old = delayed_local_psi_bank_.GetOldDeviceStorage();
```
wdhawkins, line 317: I think there's an aliasing issue here. If `has_integrated_memory == false`, `pointer_set.delayed_local_psi` and `pointer_set.delayed_local_psi_old` are the same, but are declared as `__restrict__`. Before this PR, they were two separate pointers, but this is a real bug now.

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/fluds/aahd_fluds.cu
```diff
@@ -85,7 +155,15 @@ AAHD_NonLocalDelayedBank::AAHD_NonLocalDelayedBank(const std::vector<std::size_t
                                                    std::size_t stride)
   : AAHD_NonLocalBank(loc_sizes, loc_offsets, stride)
 {
-  host_current_storage = crb::HostVector<double>(loc_offsets.back() * stride_size, 0.0);
+  if (LBSProblem::use_gpu_aware_mpi and LBSProblem::has_integrated_memory)
```
wdhawkins, line 158: This doesn't seem to match with line 140 above. It looks like this will bypass device-to-device if `use_gpu_aware_mpi` is true and `has_integrated_memory` is false.

wdhawkins (changes requested):

PR #1126, opened 2026-08-13
K-effective Sensitivity Postprocessor

quincy-huhn98:
## PR Checklist

Code for a cross section sensitivity post processor originally written by codex. I added the ability to have sensitivities with respect to k.


quincy-huhn98:
The failing tests were new synthetic ones that were added before the change to the python api. I've removed them because there doesn't seem to be an elegant way to set a synthetic/manufactured psi with the current api.

wdhawkins:
> The failing tests were new synthetic ones that were added before the change to the python api. I've removed them because there doesn't seem to be an elegant way to set a synthetic/manufactured psi with the current api.

If you just need psi as function of angle index and group (and time - but you're probably not interested in that), we do support that.

quincy-huhn98:
I don't need psi, the test was using the old `GetPsi()` to set an artificial psi by modifying its contents. Now that `GetPsi()` returns a copy the test was failing.

test/python/modules/linear_boltzmann_solvers/xs_sensitivity/xs_sensitivity_2g.xs
wdhawkins, line 1: This file doesn't seem to be used in any of the tests. If it's used in a follow-up commit, that's fine. If it's not used at all, it might be best to drop it.

python/lib/solver.cc
```diff
@@ -328,6 +329,17 @@ WrapLBS(py::module& slv)
         Memory view of the local new scalar flux vector.
     )"
   );
+  lbs_problem.def(
+    "SetPhiNewLocal",
```
wdhawkins: I also don't see that this is used. Again, if this is needed for the following commit, just ignore this comment.

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/cross_section_sensitivity_postprocessor.cc
```diff
@@ -0,0 +1,580 @@
...
+    std::vector<int>{},
+    "Block restriction for the postprocessor. Empty/unspecified means no block restriction.");
+  params.AddOptionalParameterArray("logical_volumes",
+                                   std::vector<std::shared_ptr<LogicalVolume>>{},
+                                   "Logical volumes to restrict the computation to.");
+  params.AddOptionalParameter<std::string>(
+    "sensitivity_type", "sigma_t", "Sensitivity type: 'sigma_t', 'scatter', or 'production'.");
+  params.AddOptionalParameter("group", 0, "Group for sigma_t or production sensitivities.");
+  params.AddOptionalParameter("moment", 0, "Scattering moment/order for scatter sensitivities.");
+  params.AddOptionalParameter("ell", 0, "Deprecated alias for 'moment'.");
```
wdhawkins: Since this post-processor is new functionality, I don't see any reason to add this as a deprecated alias. It can just be removed.

wdhawkins (changes requested):

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/cross_section_sensitivity_postprocessor.h
```diff
@@ -0,0 +1,88 @@
...
+
+  explicit CrossSectionSensitivityPostprocessor(const InputParameters& params);
+
+  void Execute();
+  void ApplyKEigenvalueScaling(double k_eff);
+
+  const NDArray<double, 2>& GetValue() const;
+
+  static InputParameters GetInputParameters();
+  static std::shared_ptr<CrossSectionSensitivityPostprocessor> Create(const ParameterBlock& params);
```
andrsd: `static` member functions should be after normal member function/variables, see [coding standard](https://open-sn.github.io/opensn/devguide/coding_standard.html#classes-and-structs)

andrsd (changes requested):

test/python/modules/linear_boltzmann_solvers/xs_sensitivity/xs_sensitivity_2g.xs
quincy-huhn98: I've removed it

python/lib/solver.cc
quincy-huhn98: It was used in the previous synthetic tests. I've removed it

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/cross_section_sensitivity_postprocessor.cc
quincy-huhn98: I agree, thanks for catching it I've removed it.

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/cross_section_sensitivity_postprocessor.h
quincy-huhn98: Fixed.

wdhawkins (approved):

andrsd (approved):

PR #1127, opened 2026-08-16
Input hardening

wdhawkins:
This PR includes various checks to prevent OpenSn crashing or hanging on bad input. I asked Codex to extensively fuzz inputs and intentionally try to crash or hang the code by constructing invalid inputs.

## PR Checklist

### DIFF
--- a/framework/materials/multi_group_xs/xsfile.cc
+++ b/framework/materials/multi_group_xs/xsfile.cc
@@ -76,6 +76,10 @@ XSFile::Read()
       OpenSnLogicalErrorIf(not(static_cast<bool>(line_stream >> n_groups)),
                            "Failed parsing NUM_GROUPS value.");
       OpenSnLogicalErrorIf(n_groups <= 0, "The number of energy groups must be positive.");
+      // Reject implausibly large groups counts before they can exhaust available memory
+      OpenSnLogicalErrorIf(n_groups > 5000,
+                           "The number of energy groups (" + std::to_string(n_groups) +
+                             ") is implausibly large. Values greater than 5000 are rejected.");
       num_groups_ = n_groups;
       continue;
     }
@@ -88,6 +92,11 @@ XSFile::Read()
       OpenSnLogicalErrorIf(not(static_cast<bool>(line_stream >> n_moments)),
                            "Failed parsing NUM_MOMENTS value.");
       OpenSnLogicalErrorIf(n_moments < 0, "The number of scattering moments must be non-negative.");
+      // ReadTransferMatrices allocates one SparseMatrix(num_groups, num_groups) per moment. Reject
+      // implausibly large counts before they can exhaust available memory.
+      OpenSnLogicalErrorIf(n_moments > 1000,
+                           "The number of scattering moments (" + std::to_string(n_moments) +
+                             ") is implausibly large. Values greater than 1000 are rejected.");
       scattering_order_ = std::max(0, n_moments - 1);
       continue;
     }
@@ -101,6 +110,9 @@ XSFile::Read()
                            "Failed parsing NUM_PRECURSORS value.");
       OpenSnLogicalErrorIf(n_prec < 0,
                            "The number of delayed neutron precursors must be non-negative.");
+      OpenSnLogicalErrorIf(n_prec > 100,
+                           "The number of delayed neutron precursors (" + std::to_string(n_prec) +
+                             ") is implausibly large. Values greater than 100 are rejected.");
       num_precursors_ = n_prec;
       precursors_.resize(num_precursors_);
       continue;
--- a/framework/math/quadratures/angular/angular_quadrature.h
+++ b/framework/math/quadratures/angular/angular_quadrature.h
@@ -5,6 +5,7 @@
 
 #include "framework/data_types/vector3.h"
 #include "framework/data_types/ndarray.h"
+#include "framework/utils/error.h"
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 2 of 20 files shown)


framework/materials/multi_group_xs/xsfile.cc
```diff
@@ -76,6 +76,10 @@ XSFile::Read()
       OpenSnLogicalErrorIf(not(static_cast<bool>(line_stream >> n_groups)),
                            "Failed parsing NUM_GROUPS value.");
       OpenSnLogicalErrorIf(n_groups <= 0, "The number of energy groups must be positive.");
+      // Reject implausibly large groups counts before they can exhaust available memory
+      OpenSnLogicalErrorIf(n_groups > 100000,
```
ragusa: in practic, i've seen group structures with up to 2,200 groups. would you consider a lower number than 100,000? or maybe a warning for numbers larger than 5,000

framework/materials/multi_group_xs/xsfile.cc
```diff
@@ -88,6 +92,12 @@ XSFile::Read()
       OpenSnLogicalErrorIf(not(static_cast<bool>(line_stream >> n_moments)),
                            "Failed parsing NUM_MOMENTS value.");
       OpenSnLogicalErrorIf(n_moments < 0, "The number of scattering moments must be non-negative.");
+      // ReadTransferMatrices allocates one SparseMatrix(num_groups, num_groups) per moment. Reject
+      // implausibly large counts before they can exhaust available memory (mirrors the group cap
+      // above).
+      OpenSnLogicalErrorIf(n_moments > 1000,
```
ragusa, line 97: I can imagine a study in 1d, where 1,000 is doable, but no studies in 3D where 1,000 is warranted.  in 1d, I can use 1-group and delta scattering expansion or FP expansion i do mylsef to test, so 1000 moments in 1d is possibly ok. note that 1000 moment is 3d is only about scattering order = 30. I can imagine 3d calculations with L=7 (and this will be memory heavy) but my point is that L=7 is not that farfrom L=30. nothing to change here, just a comment

framework/materials/multi_group_xs/xsfile.cc
```diff
@@ -101,6 +111,9 @@ XSFile::Read()
                            "Failed parsing NUM_PRECURSORS value.");
       OpenSnLogicalErrorIf(n_prec < 0,
                            "The number of delayed neutron precursors must be non-negative.");
+      OpenSnLogicalErrorIf(n_prec > 1000,
```
ragusa: i do not know of numbers greater than 8, so maybe make this 100.

framework/materials/multi_group_xs/xsfile.cc
```diff
@@ -101,6 +111,9 @@ XSFile::Read()
...
+                           "The number of delayed neutron precursors (" + std::to_string(n_prec) +
+                             ") exceeds the maximum supported value of 1000.");
```
ragusa: this is a general comment: everywhere you say "exceeds the maximum supported value of", but “Maximum supported value” strongly suggests a technical/code limitation. I would instead make the message clearly about input validation / reasonableness. possible other choice: "is unreasonably large; values greater than NNNN are rejected."

framework/materials/multi_group_xs/xsfile.cc
```diff
@@ -88,6 +92,12 @@ XSFile::Read()
       OpenSnLogicalErrorIf(not(static_cast<bool>(line_stream >> n_moments)),
                            "Failed parsing NUM_MOMENTS value.");
       OpenSnLogicalErrorIf(n_moments < 0, "The number of scattering moments must be non-negative.");
+      // ReadTransferMatrices allocates one SparseMatrix(num_groups, num_groups) per moment. Reject
+      // implausibly large counts before they can exhaust available memory (mirrors the group cap
```
ragusa: not sure what it mirrors. 1000 is not 1000000

framework/math/quadratures/angular/angular_quadrature.h
```diff
@@ -81,6 +82,12 @@ class AngularQuadrature
       scattering_order_(scattering_order),
       construction_method_(method)
   {
+    // Harmonic evaluation cost grows with scattering_order, and an implausibly
+    // large value can appear as a hang. Reject it up front.
+    OpenSnInvalidArgumentIf(scattering_order > 100,
```
ragusa: see my previous comment above on n_moment. here, I think you need to allow L=1000 for 1d runs. and given that this is in the quadrature object and that we give the dim to that object, you may want to adapt the error message to the requested dim and somehow the limits on L here should be made consistent with the limits on n_moments above

ragusa (changes requested):
just a few thoughts that may lead to some minor adjustments

andrsd (approved):

framework/materials/multi_group_xs/xsfile.cc
wdhawkins: Fixed.

framework/materials/multi_group_xs/xsfile.cc
wdhawkins: This is fixed and I incorporated your suggestion about about the per-dimension limit.

framework/materials/multi_group_xs/xsfile.cc
wdhawkins: Fixed.

framework/materials/multi_group_xs/xsfile.cc
wdhawkins: Agreed. Fixed.

framework/materials/multi_group_xs/xsfile.cc
wdhawkins: Fixed.

framework/math/quadratures/angular/angular_quadrature.h
wdhawkins: Fixed.

ragusa (approved):

PR #1133, opened 2026-08-23
Refactor `Cell` class

andrsd:
This is on top of #1118

- Moves `CellFace` out of `Cell`
- Moves cell face vertices out of `Cell`
- Removing Lightweight Cell/CellFace

## PR Checklist


quocdang1998 (approved):

modules/linear_boltzmann_solvers/lbs_problem/device/carrier/mesh_carrier.cu
```diff
@@ -75,32 +76,32 @@ void
...
+    const CellMapping& cell_mapping = discretization.GetLocalCellMapping(cell);
     std::size_t cell_num_nodes = cell_mapping.GetNumNodes();
     // check for cell num nodes compatibility with sweep kernel
     if (cell_num_nodes > LBSProblem::max_dofs_gpu)
     {
       throw std::runtime_error(
         std::format("GPU acceleration error: Cell local ID {} has {} DOFs which exceeds the "
                     "maximum supported DOFs per cell on GPU: {}.",
-                    cell.local_id,
+                    cell_local_id,
```
wdhawkins, line 106: Where does `cell_local_id` come from for this routine?

framework/mesh/mesh/mesh.h
```diff
@@ -90,10 +207,20 @@ class MeshContinuum
...
+  Cell& GetLocalCell(uint64_t id);
+
+  std::vector<Cell>& GetLocalCells();
+  const std::vector<Cell>& GetLocalCells() const;
+
   /// Returns whether the cell with the given global id is locally owned.
   bool IsCellLocal(uint64_t global_id) const noexcept
   {
-    return global_cell_id_to_local_id_map_.contains(global_id);
+    assert(global_to_local_cell_id_map_.contains(global_id));
```
wdhawkins, line 214: This routine is marked `noexcept` but calls `assert` followed by `at`.

framework/mesh/mesh/mesh.cc
```diff
@@ -0,0 +1,1370 @@
...
+  std::size_t total_len = 0;
+  for (const auto& cell : connectivity)
+    total_len += cell.size();
+  connect_ids_.reserve(total_len);
+
+  connect_ofst_.reserve(connectivity.size() + 1);
+  for (const auto& cell : connectivity)
+  {
+    connect_ofst_.push_back(connect_ids_.size());
+    std::copy(cell.begin(), cell.end(), connect_ids_.end());
```
wdhawkins, line 255: Since we call `reserve` on `connect_ids_` above, I don't think this will work correctly. We'll just keep overwriting data.

framework/mesh/mesh/mesh.h
```diff
@@ -69,9 +142,53 @@ class MeshContinuum
...
+  const Cell& GetGlobalCell(uint64_t cell_global_index) const;
+
+  /**
+   * Returns the cell global ids of all ghost cells. These are cells that neighbors to this
+   * partition's cells but are on a different partition.
+   */
+  std::vector<uint64_t> GetGhostGlobalIDs() const;
+
+  /**
+   * Returns the local storage address of a ghost cell. If the ghost is not truly a ghost then -1 is
```
wdhawkins: I believe this method now just returns an error about calling a delete method.

framework/mesh/mesh/mesh.cc
```diff
@@ -0,0 +1,1370 @@
...
+std::size_t
+Mesh::GetLocalCellCount() const
+{
+  return local_cells_.size();
+}
+
+Cell&
+Mesh::GetLocalCell(uint64_t id)
+{
+  if (id < local_cells_.size())
```
wdhawkins, line 442: What if `id > local_cells_.size()`?

framework/mesh/mesh/mesh.cc
```diff
@@ -0,0 +1,1370 @@
...
+  mpi_comm.all_gather(local_bndry_ids, global_bndry_ids);
+
+  std::set<uint64_t> global_bndry_ids_set(global_bndry_ids.begin(), global_bndry_ids.end());
+
+  std::vector<uint64_t> unique_bdnry_ids(global_bndry_ids_set.begin(), global_bndry_ids_set.end());
+  return unique_bdnry_ids;
+}
+
+void
+Mesh::ComputeGeometricInfo()
```
wdhawkins, line 130: Will this perform as well as the original on very large mesh? I think the original was one loop over cells that computed everything.

framework/mesh/mesh/mesh.cc
```diff
@@ -0,0 +1,1370 @@
...
+  {
+    connect_ofst_.push_back(connect_ids_.size());
+    const auto& vertex_ids = cell_connectivity.at(cell.global_id);
+    for (const auto& id : vertex_ids)
+      connect_ids_.push_back(id);
+  }
+  connect_ofst_.push_back(total_len);
+}
+
+#if 0
```
wdhawkins, line 303: Can this be deleted?

wdhawkins (changes requested):

PR #1134, opened 2026-08-26
Parallelize VolumePostprocessor for GPU sweep

quocdang1998:
This PR parallelizes the computation of `VolumePostprocessor`.

Instead of executing `mpi_comm.all_reduce` for every groups, each rank computes its own contribution and executes the vectorized version of `mpi_comm.all_reduce` once.

The summation of the flux over the cells is also parallelized for each thread.

This allows to significantly reduce the execution time of `VolumePostprocessor` when the mesh per rank is big, especially for GPU sweep case.

For reference, this PR reduces BEAVRS benchmark (1.28M cells)'s computation of power per pins from >3hrs to 10min computation of power per pins.

## PR Checklist

### DIFF
--- a/modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
+++ b/modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
@@ -5,6 +5,7 @@
 #include "framework/object_factory.h"
 #include "framework/math/spatial_discretization/finite_element/finite_element_data.h"
 #include "framework/runtime.h"
+#include "modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/sweep_parallel_for.h"
 #include <cstdint>
 #include <limits>
 #include <memory>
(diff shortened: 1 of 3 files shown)


modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
```diff
@@ -226,23 +227,113 @@ VolumePostprocessor::CreateEnergyRestriction()
...
+                            static_cast<std::vector<double>::difference_type>((i + 1) * n_groups)));
+}
+
+void
+VolumePostprocessor::ExecuteWeightedAverage()
+{
+  const auto n_lvs = cell_local_ids_.size();
+  const auto n_groups = groups_.size();
+  const auto n_threads =
+    std::min<std::size_t>(std::max(1U, opensn_num_threads), std::max<std::size_t>(n_lvs, 1));
```
wdhawkins: Is this correct if `n_lvs == 1`? That would mean that the number of threads is 1 and we lose all of our parallelism.

wdhawkins (changes requested):

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
quocdang1998: Fixed. The code now executes all cells in all levels in parallel.

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
```diff
@@ -265,152 +371,90 @@ VolumePostprocessor::GetCoefficients(const Cell& cell)
...
 
-std::vector<double>
-VolumePostprocessor::ComputeIntegral(const std::vector<uint32_t>& cell_local_ids)
+void
+VolumePostprocessor::ComputeIntegral(std::span<double> row, const Cell& cell)
 {
   const auto& sdm = lbs_problem_->GetSpatialDiscretization();
-  const auto& grid = sdm.GetGrid();
   const auto& uk_man = lbs_problem_->GetUnknownManager();
   const auto phi = lbs_problem_->GetPhiNewLocal();
```
wdhawkins: Technically, not part of the PR, but should this be `const auto& phi`? Otherwise, I think this is going to be a deep copy of phi every time. `coord` has the same problem I think. This comment also applies to `ComputeMax`, `ComputeMin`, etc.

wdhawkins (changes requested):

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
andrsd: For `phi`, yes. Should have been reference. I can send a patch.
For `coord`, no. The `GetSpatialWeightingFunction` returns a value.

andrsd: See #1145 for the patch

andrsd (approved):

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
quocdang1998: All occurrences of `GetPhiNewLocal` have been converted to `const auto&`. @andrsd You can close [1145](https://github.com/Open-Sn/opensn/pull/1145).

andrsd: Ok.

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
```diff
@@ -226,23 +227,128 @@ VolumePostprocessor::CreateEnergyRestriction()
...
+      work.emplace_back(i, cell_id);
+
+  const auto n_threads =
+    std::min<std::size_t>(std::max(1U, opensn_num_threads), std::max<std::size_t>(work.size(), 1));
+
+  std::vector<std::vector<double>> thread_local_all(
+    n_threads, std::vector<double>(n_lvs * n_groups, identity));
+
+  const auto& grid = lbs_problem_->GetSpatialDiscretization().GetGrid();
+  ParallelFor(work.size(),
```
wdhawkins, line 273: Should we short-circuit this in the case of 1 thread?

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
```diff
@@ -226,23 +227,128 @@ VolumePostprocessor::CreateEnergyRestriction()
...
+
+  std::vector<std::pair<std::size_t, std::uint32_t>> work;
+  for (std::size_t i = 0; i < n_lvs; ++i)
+    for (const auto cell_id : cell_local_ids_[i])
+      work.emplace_back(i, cell_id);
+
+  const auto n_threads =
+    std::min<std::size_t>(std::max(1U, opensn_num_threads), std::max<std::size_t>(work.size(), 1));
+
+  std::vector<std::vector<double>> thread_local_all(
```
wdhawkins, line 269: For some operations (like integral and average), this might not be deterministic across thread counts. It might be worth adding a comment about that.

modules/linear_boltzmann_solvers/lbs_problem/postprocessors/volume_postprocessor.cc
wdhawkins, line 361: A bit outside the scope, but this should probably be `const auto xs_map&` to avoid a per-thread copy.

wdhawkins (changes requested):

PR #1135, opened 2026-08-27
Add group and block restriction to CreateFieldFunction (for reaction rate field functions)

kitamuraa24:
For an ordinary XS name, ' CreateFieldFunction ' already produced the reaction-rate density sum_g xs[g] * phi_g at each node, so we just extended it with two optional arguments:

  group:     restrict the field to a single energy group (-1 = all groups)
  block_ids: restrict the field to a set of block ids (empty = all cells)

These changes are reflected in ' UpdateDerivedFieldFunction ' and ' ComputeXSFieldFunctionData ', so there is still exactly one pathway computing the field. Cells outside block_ids are skipped before they write, which leaves their nodes at zero and lets a restricted field be integrated over the whole domain to recover a per-block reaction rate.

group and block_ids are rejected for xs_name == "power", whose normalization is a global reduction and thus has no meaningful per-group or per-block form ( at least to my understanding ).

XS names remain those understood by MultiGroupXS::GetByName; no alias layer is introduced.

Adds a 1D Reed test covering the all-group and block-restricted paths, and a 2D two-group test covering the per-group path plus PVTU export.

## PR Checklist

- [✔] I have updated the user guide and/or Python API documentation if necessary.


andrsd (approved):

doc/source/userguide/post_processors.rst
```diff
@@ -143,12 +143,41 @@ If a power-normalized view is needed:
...
+   :py:class:`pyopensn.post.VolumePostprocessor` when a scalar reaction rate
+   over a volume is wanted rather than a spatially resolved field.
+
 Important points:
 
 * ``name`` is the field-function name assigned to the returned object
 * ``xs_name`` can be a built-in 1D XS name, a custom XS name, or ``"power"``
 * ``power_normalization_target`` is optional and affects only the returned
   field function
+* ``group`` selects a single energy group; ``-1`` (default) sums over all groups
```
ragusa: in volume postprocessor, is `-1` our choice for all groups? We should try to stay consistent.

modules/linear_boltzmann_solvers/lbs_problem/lbs_problem.h
```diff
@@ -296,9 +296,22 @@ class LBSProblem : public Problem, public std::enable_shared_from_this<LBSProble
   std::shared_ptr<FieldFunctionGridBased> CreateScalarFluxFieldFunction(unsigned int g,
                                                                         unsigned int m = 0);
 
-  /// Creates a named field function from a 1D XS or the special case `power`.
-  std::shared_ptr<FieldFunctionGridBased> CreateFieldFunction(
-    const std::string& name, const std::string& xs_name, double power_normalization_target = -1.0);
+  /**
+   * Creates a named field function from a 1D XS or the special case `power`.
```
ragusa, line 301: feels awkward to me that this is the description of a function called `CreateScalarFluxFieldFunction`. Thoughts, @wdhawkins ?

test/assets/xs/rr_2d_2g_absorber_mat.xs
```diff
@@ -0,0 +1,20 @@
...
+
+SIGMA_T_BEGIN
+0 1.5
+1 2.0
+SIGMA_T_END
+
+SIGMA_A_BEGIN
+0 1.5
+1 2.0
+SIGMA_A_END
```
ragusa, line 12: it is often better not to give that to OpenSn. It recomputes it with total and scattering anyway

test/assets/xs/rr_2d_2g_source_mat.xs
```diff
@@ -0,0 +1,21 @@
...
+
+SIGMA_T_BEGIN
+0 1.0
+1 0.7
+SIGMA_T_END
+
+SIGMA_A_BEGIN
+0 0.3
+1 0.4
+SIGMA_A_END
```
ragusa, line 12: same

test/python/modules/linear_boltzmann_solvers/transport_steady/reed_test_rr_field_func.py
```diff
@@ -0,0 +1,126 @@
...
+    ss_solver = SteadyStateSourceSolver(problem=phys)
+    ss_solver.Initialize()
+    ss_solver.Execute()
+
+    logvol_whole_domain = RPPLogicalVolume(infx=True, infy=True, zmin=0.0, zmax=sum(widths))
+
+    total_rr_ff = phys.CreateFieldFunction("rr_total", "sigma_t")
+    abs_rr_ff = phys.CreateFieldFunction("rr_abs", "sigma_a")
+    block3_total_rr_ff = phys.CreateFieldFunction("rr_total_block3", "sigma_t", block_ids=[3])
+    phi_ff = phys.GetScalarFluxFieldFunction(True)[0]
```
ragusa, line 93: how is this functionality different than that of volumepostprocessors?

ragusa (commented):
maybe i didn't fully get it, but is volume postprocess not able to do this?

test/python/modules/linear_boltzmann_solvers/transport_steady/reed_test_rr_field_func.py
kitamuraa24: My understanding is that VolumePostProcessors is a global reduction on a specified volume. This is not; it is creating fields of reaction rate densities across a specified group and or block_id

doc/source/userguide/post_processors.rst
kitamuraa24: No, it uses an optional param. I also realize that I should implement the groupset option too; I didn't think about that at the time. I will go ahead and do that this weekend.

test/assets/xs/rr_2d_2g_absorber_mat.xs
kitamuraa24: these were done because I am reading in the provided cross-section values. I had it this way so that when I start doing tritium production rates, I would just have this array in the .h5 xs file, and the code will just find it and read it.

PR #1141, opened 2026-09-03
Fix non-deterministic GPU sweep results with opposing reflecting boundaries

wdhawkins:
GPU sweeps with opposing reflecting boundaries and 2+ MPI ranks could produce wrong results that vary run-to-run.

`RecursiveAngleSort` used a `std::set<AngleSet*>` ordered by raw pointer value to decide which angle set's indices get
propagated to its reflected partner. With one reflecting boundary this works fine, but with opposing boundaries, some
angle sets are constrained by two reflections simultaneously and which one wins depends on visitation order. This ordering was driven by heap addresses that vary per run and per rank. Since the GPU cross-rank sweep tags messages by angleset ID and assumes every rank agrees on the meaning, two ranks could end up disagreeing and corrupting the exchange from the first sweep onward. The fix is straightforward: order by `AngleSet::GetID()` instead of pointer address.

### DIFF
--- a/modules/linear_boltzmann_solvers/discrete_ordinates_problem/boundary/discrete_ordinates_boundaries.cc
+++ b/modules/linear_boltzmann_solvers/discrete_ordinates_problem/boundary/discrete_ordinates_boundaries.cc
@@ -49,12 +49,22 @@ GetGlobalUniqueBoundaryIDs(const std::shared_ptr<MeshContinuum>& grid, mpi::Comm
   return global_unique_bids_set;
 }
 
+// RecursiveAngleSort's traversal order determines, for each reflected pair of angle sets, which
+// set's angle indices get propagated to the other. With a single reflecting boundary this is
+// order-independent but with opposing reflecting boundaries, the final angle-to-angleset
+// assignment depends on visitation order. unsorted/sorted MUST order by a value stable across
+// MPI ranks and repeated runs. For this purpose, we use AngleSet::GetID().
+struct AngleSetIdLess
+{
+  bool operator()(const AngleSet* a, const AngleSet* b) const { return a->GetID() < b->GetID(); }
+};
+
 void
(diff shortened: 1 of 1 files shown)


modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/angle_set.h
```diff
@@ -138,7 +139,7 @@ class AngleSet
   /// Number of anglesets the current angle set depends on.
   std::size_t num_dependencies_ = 0;
   /// Counter for un-resolved dependencies.
-  std::size_t dependency_counter_ = 0;
+  std::atomic<std::size_t> dependency_counter_ = 0;
```
quocdang1998: I think this deserves its own PR.

I chose not to use `std::atomic` because it degrades the performance of the sweep. The only data race occurs when the main thread queries `AAHD_AngleSet::IsReady` while a worker thread is executing `AAHD_AngleSet::DecrementCounter`. This data race is harmless in practice: if the main thread reads the counter before it's decremented, the angleset is simply deferred to the next `while` iteration.

Using `std::atomic` here would hurt performance, since each atomic operation carries its own lock guard. The decrement operation is already protected by a lock shared across all `dependency_counter_` instances within a given angleset. Please verify this with the scaling test.

quocdang1998 (changes requested):

modules/linear_boltzmann_solvers/discrete_ordinates_problem/sweep/angle_set/angle_set.h
wdhawkins: Fair enough. You are correct that the worst side effect of this is a recoverable stall - although it is technically undefined behavior. I'll move this to a separate PR and we can decide there whether to keep it or not.

quocdang1998 (approved):

